PR #18, opened 2016-12-02
Generate test reports

Psirus:
The reports are in a format JUNIT, which is needed for the Jenkins
Performance plugin. Only boost from version 1.62 onwards can generate
JUNIT reports. To generate the reports, set an environment variable
PERFORMANCE_TEST.


Psirus:
Just a test.

PR #36, opened 2017-06-07
Refactor interpolation

TTitscher:
The aim is to simplify the interpolation type. 

### Major changes:
- Remove `GetNumDofsPerNode()`  from the InterpolationType and move it to the `NodeEnum`. This change is in preparation of a future `class Dof`. 
- Remove the `IntegrationTypeBase` from the InterpolationType. This dependence was introduced for a _caching_ (or _memoization_) of the shape functions and their derivatives at certain integration points. This was done via the _integration point index_. This dependency is removed by introducing the `NuTo::NaturalCoordinateMemoizer` that performs this caching based on the _natural coordinate_.

### Minor changes:
- Renaming: remove prefix and `Get`, add suffix `IGA/FEM` in the `InterpolationBase` interface, IMO the prefix `Get` is somehow redundant... What else could `Interpolation::ShapeFunction(...)` mean?

### Open issues:
- somehow "non polymorphic" coupling of InterpolationFEM and InterpolationIGA
- definition of the "standard integration type" - this might have to be defined by the user

### @ Reviewers
Please help me to improve the quality of the code! Thanks!


applications/benchmark/ShapeFunctionMemoization.cpp
```diff
@@ -4,61 +4,203 @@
 #include "mechanics/interpolationtypes/Interpolation3DTetrahedron.h"
 #include "mechanics/nodes/NodeEnum.h"
 #include "mechanics/interpolationtypes/InterpolationTypeEnum.h"
+#include "mechanics/elements/ElementShapeFunctions.h"
 
-#include "math/NaturalCoordinateMemoizer.h"
+#include <vector>
+#include <unordered_map>
+//#include <boost/container/flat_unordered_map.hpp>
```
Psirus: No need to keep this around.

applications/integrationtests/mechanics/PlateWithHole.cpp
```diff
@@ -101,27 +50,17 @@ bool CheckSolution(NuTo::Structure& s)
...
 }
 
-//! @brief Imports a mesh file and builds the hessian and the internal gradient.
-int main(int argc, char* argv[])
+BOOST_AUTO_TEST_CASE(PlateWithHole)
 {
-    boost::filesystem::path binaryPath = std::string(argv[0]);
+    auto binary = boost::unit_test::framework::master_test_suite().argv[0];
+    boost::filesystem::path binaryPath = std::string(binary);
     binaryPath.remove_filename();
```
Psirus, line 64: What is happening here? What is being deleted, and why?

src/mechanics/elements/ContinuumContactElement.cpp
```diff
@@ -27,7 +27,7 @@ NuTo::ContinuumContactElement<TDim>::ContinuumContactElement(const ContinuumElem
                                                              const Group<ElementBase>* elementGroup,
                                                              const Group<NodeBase>* nodeGroup,
                                                              const IntegrationTypeBase *rIntegrationType)
-    : ContinuumBoundaryElement<TDim>(rSlaveElement, rSurfaceId), mIntegrationType(rIntegrationType)
+    : ContinuumBoundaryElement<TDim>(rSlaveElement, *rIntegrationType, rSurfaceId), mIntegrationType(rIntegrationType)
```
Psirus: `ContinuumElement` and `ContinuumBoundaryElement` etc. now take the integration type as a reference in the constructor. To make this more uniform, the constructor of this class should as well.

src/mechanics/elements/ContinuumElementIGA.cpp
```diff
@@ -55,7 +56,8 @@ const Eigen::VectorXd NuTo::ContinuumElementIGA<TDim>::GetIntegrationPointVolume
     Eigen::VectorXd volume(this->GetNumIntegrationPoints());
     for (int theIP = 0; theIP < this->GetNumIntegrationPoints(); theIP++)
     {
-        Eigen::MatrixXd derivativeShapeFunctionsNatural = this->mInterpolationType->Get(Node::eDof::COORDINATES).CalculateDerivativeShapeFunctionsNatural(theIP, mKnotIDs);
+        const auto ipCoords = this->GetIntegrationType().GetLocalIntegrationPointCoordinates(theIP);
```
Psirus, line 59: What's with all the `this->Whatever()` here?

src/mechanics/elements/ContinuumElementIGA.cpp
```diff
@@ -55,7 +56,8 @@ const Eigen::VectorXd NuTo::ContinuumElementIGA<TDim>::GetIntegrationPointVolume
     Eigen::VectorXd volume(this->GetNumIntegrationPoints());
     for (int theIP = 0; theIP < this->GetNumIntegrationPoints(); theIP++)
```
Psirus, line 57: Would it be nice to have the integrationType be iterable, i.e. `for(auto& IP : integrationType)`?

src/mechanics/elements/ContinuumElementIGA.cpp
```diff
@@ -96,7 +100,7 @@ void NuTo::ContinuumElementIGA<TDim>::CheckElement()
         size += this->GetIntegrationPointWeight(iIP) * detJacobian;
     }
 
-    assert(std::abs(GetIntegrationPointVolume().sum() / size - 1) < 1.e-6); // just to be sure ...
+    //assert(std::abs(GetIntegrationPointVolume().sum() / size - 1) < 1.e-6); // just to be sure ...
```
Psirus: Why is that not true anymore?

src/mechanics/interpolationtypes/Interpolation1DTruss.h
```diff
@@ -75,10 +47,16 @@ class Interpolation1DTruss: public Interpolation1D
     template<class Archive>
     void serialize(Archive & ar, const unsigned int version);
 #endif  // ENABLE_SERIALIZATION
+protected:
+
+    Eigen::VectorXd CalculateNaturalNodeCoordinates(int rNodeIndexDof) const override;
```
Psirus, line 52: Good move. We often declare functions `public` by default, but should tend towards `protected` or `private`.

src/mechanics/interpolationtypes/InterpolationBase.cpp
```diff
@@ -130,7 +126,6 @@ void NuTo::InterpolationBase::serialize(Archive & ar, const unsigned int version
     ar & BOOST_SERIALIZATION_NVP(mNodeIndices);
 
     ar & BOOST_SERIALIZATION_NVP(mSurfaceNodeIndices);
-    ar & BOOST_SERIALIZATION_NVP(mUpdateRequired);
```
Psirus, line 133: This serialization stuff. Changes you had to make, I had to review, no one benefits from...

src/mechanics/interpolationtypes/InterpolationBaseFEM.cpp
```diff
@@ -14,95 +14,79 @@
...
+#include "mechanics/nodes/NodeEnum.h"
 
-NuTo::InterpolationBaseFEM::InterpolationBaseFEM(Node::eDof rDofType, NuTo::Interpolation::eTypeOrder rTypeOrder, int rDimension) :
-    InterpolationBase::InterpolationBase(rDofType, rTypeOrder, rDimension)
+using namespace NuTo;
+
+InterpolationBaseFEM::InterpolationBaseFEM(Node::eDof rDofType, Interpolation::eTypeOrder rTypeOrder, int rDimension)
+    : InterpolationBase::InterpolationBase(rDofType, rTypeOrder, rDimension),
+    mShapeFunctions([=] (const Eigen::VectorXd& v) { return CalculateShapeFunctions(v);}),
+    mMatrixN([=] (const Eigen::VectorXd& v) { return CalculateMatrixN(v);}),
```
Psirus, line 28: Nice use of lambdas. I like lambdas :-)

Psirus (approved):
Nice! Quite a bit of cleanup. Personally, I like the move from `GetShapeFunctions` to just `ShapeFunctions`. Good job on the IGA test refactor. I like the idea of reusing the analytic solution. Maybe we can collect more such cases?

Two more things that I noticed:
- The coverage decreased because the `GetEnumType` is now untested in many integration types. If we don't need it, let's just remove it.
- One minor style thing: in [applications/integrationtests/mechanics/InterpolationTypes.cpp](https://github.com/nutofem/nuto/blob/master/applications/integrationtests/mechanics/InterpolationTypes.cpp) there are sometimes not even one blank line between test cases. I'd prefer two, but one at the least.

applications/integrationtests/mechanics/PlateWithHole.cpp
TTitscher: This tests writes its output in a directory based on the current binary location. This location is passed via `argv[0]` - path and filename and the filename is removed (`remove_filename()`). In `master` this test was a normal main file and `argv[0]` was passed as an argument. Now, it is extracted from the `master_test_suite()`

src/mechanics/elements/ContinuumElementIGA.cpp
TTitscher: See issue #40

src/mechanics/elements/ContinuumElementIGA.cpp
TTitscher: http://en.cppreference.com/w/cpp/language/this

`mInterpolationType` is not a member of `ContinuumElementIGA` but from its base class `ContinuumElement` and has to be referenced with `this`.

applications/benchmark/ShapeFunctionMemoization.cpp
TTitscher: Sure! eae302b3422aa61af4e3449aaf0eca164b01e61a

src/mechanics/elements/ContinuumContactElement.cpp
TTitscher: Sure! eae302b3422aa61af4e3449aaf0eca164b01e61a

src/mechanics/elements/ContinuumElementIGA.cpp
TTitscher: This is a tricky one. For some reason, this assert fails when running the IGA test. Maybe there is a bug in `GetIntegrationPointVolume()`, maybe in the calculation of the determinant. @potto1, you are the IGA expert, could you have a look?

TTitscher: nevermind, fixed c8bf299728085e2fb79a34ffc8958d7667ad19ec

src/mechanics/elements/ContinuumElementIGA.cpp
Psirus: The problem is that `ContinuumElement` is a base class template, where spezializations don't need to have the same interface. Therefore, the compiler doesn't look for the inherited names. Adding `this->Whatever()` promises to the compiler that all spezializations will have `Whatever()`. The same could be achieved with `using ElementBase::Whatever;` in the `ContinuumElementIGA` class.

applications/benchmark/ShapeFunctionMemoization.cpp
```diff
@@ -4,61 +4,202 @@
...
+    }
+};
+
+
+//! @brief provides a memoization of the a std::function<TResult(TNaturalCoords)>, mainly used for
+//!        element shape functions and their derivatives
+//! @param TResult ... return type of the function
+//! @param TNaturalCoords ... vector type of natural coordinates
+//! @param TIdHash ... function object that transforms TNaturalCoords into a vector id
+template <typename TResult, typename TNaturalCoords, typename TIdHash = NaturalCoordianteToId<32>>
```
joergfunger, line 50: How do we choose the maximum number of TRaster? Is there also a  check that compares the natural coordinates of the stored and the newly calculated setup close to machine precision?

src/mechanics/elements/ContinuumBoundaryElementConstrainedControlNode.h
```diff
@@ -11,7 +11,9 @@ class ContinuumBoundaryElementConstrainedControlNode : public ContinuumBoundaryE
 {
 
 public:
-    ContinuumBoundaryElementConstrainedControlNode(const ContinuumElement<TDim>& rBaseElement, int rSurfaceId, NodeBase* rControlNode);
+    ContinuumBoundaryElementConstrainedControlNode(const ContinuumElement<TDim>& rBaseElement,
+                                                   const IntegrationTypeBase& integrationType, int rSurfaceId,
+                                                   NodeBase* rControlNode);
 
     //! @brief Gets the control node of an boundary element, if it has one
```
joergfunger, line 18: What is a control node of a boundary element?

src/mechanics/elements/ContinuumElementIGA.cpp
joergfunger: Why do I need the this-> keyword here?

src/mechanics/elements/ElementBase.h
```diff
@@ -71,8 +71,9 @@ class ElementBase
 
 public:
     //! @brief constructor
-    //! @param rInterpolationType ... interpolation type
-    ElementBase(const InterpolationType& rInterpolationType);
+    //! @param interpolationType ... interpolation type
+    //! @param integrationType ... integration type
+    ElementBase(const InterpolationType& interpolationType, const IntegrationTypeBase& integrationType);
```
joergfunger, line 76: Should the ElementBase class contain the integration type as well as the constitutive models? In the current setup, this is perfectly fine, since for the "next generation" pde class this will change anyway (hopefully).

joergfunger (approved):
Nice job

src/mechanics/elements/ContinuumBoundaryElementConstrainedControlNode.h
TTitscher: I only adapted the interface, the actual implementation is from @vhirtham

applications/benchmark/ShapeFunctionMemoization.cpp
TTitscher: This benchmark contains a more efficient (but unsafer, more error prone) way of the memoization using `std::vector` instead of `std::map`. And indeed, the choice of TRaster is important. It has to be big enough to eliminate __clashes__ in one block - so that the different natural coordinates are not assigned to the same ID. It is easily possible to _detect_ these clashes with some additional lines of code and throw an exception. But how would the user solve the problem? He has to look at the implementation... So I went for the slightly more inefficient implementation as it is now. The code in this benchmark could be the starting point for a more efficient implementation. The goal here should be to actually __solve__ the ID clash instead of just detecting it. I am convinced that this is possible.

joergfunger:
One option might be to use this in a similar way such as the partitioning
of the geometry particles - just use this integer list as a subbox and only
in these subboxes use the map?

2017-06-12 10:40 GMT+02:00 TTitscher <notifications@github.com>:

applications/benchmark/ShapeFunctionMemoization.cpp
TTitscher: [I just add your answer here, something went wrong with your comment...
@joergfunger wrote: One option might be to use this in a similar way such as the partitioning
of the geometry particles - just use this integer list as a subbox and only
in these subboxes use the map?]

Certainly an option! I find it hard to tell (and to benchmark), what actually consumes the majority of the time. Comparing two natural coordinates with respect to `less`? Or with respect to `equal`? Or the actual memory access? And optimizing this is quite time consuming, so I stopped :)
I already tried using ANN. 8 times slower in the end. Why? No idea, the benchmark stops at the library call. But efficient octrees might be an option, maybe with [CGAL](http://doc.cgal.org/latest/Manual/packages.html#PartSearchStructures).

phuschke (approved):

Issue #47, opened 2017-06-14
Unused functions and members


phuschke:
Many classes have lots of unused functions and/or members.

**Detection of unused functions and members**
Most IDEs can detect usages of functions and members. It is usually safe to delete them as they have no purpose. However, there are exceptions. 

1. If a function is only called in a python test, some IDEs will not detect its usage
2. If a member is only used in a constructor (e.g. initialization to its default value), it is used but has no purpose.

**Good practice**: Members of a primitive data type should not be initialized to their default value in the initializer list of a constructor but directly when declared. This makes it easier to check for unused members and speeds up the search for their default value.

**Classes are cluttered with accessors and mutators** 
Many of the functions that are unused are accessors and/or mutators and for a **good reason**! Why do we want to access members of an object? We would much rather have the object _do_ something for us with its members. Of course, there are exceptions where accessors are necessary but those are rare and I do not think we should clutter classes with unused accessors.

> Objects expose behavior and hide data. This makes it easy to add new kinds of objects
> without changing existing behaviors. It also makes it hard to add new behaviors to existing
> objects. Data structures expose data and have no significant behavior. This makes it easy to
> add new behaviors to existing data structures but makes it hard to add new data structures
> to existing functions.
Robert Cecil Martin

**Endeavors**
- Are there any good reasons to keep unused functions you can think of?
- What do you think of initializing a primitive member to its default value directly when declared?

TTitscher:
> This makes it easier to check for unused members and speeds up the search for their default value.

This might dependent on your workflow or your IDE. If I edit a cpp file, I can perform a simple text search (no fancy code model required) in the file and the first use of a member variable often is its initialization in the ctor. I donno, as long as it is consistent, I'm fine with it. 

> Are there any good reasons to keep unused functions you can think of?

- Someone implements a new method he needs for his private main files and forgets to write a test.
- There was no test in the first place and still we trust in these methods in private main files.

[...]

PR #48, opened 2017-06-14
Cleanse time integration

phuschke:
This pull request addresses the following issue:
#47

### DIFF
--- a/src/mechanics/timeIntegration/NewmarkBase.cpp
+++ b/src/mechanics/timeIntegration/NewmarkBase.cpp
@@ -28,35 +28,18 @@
 #include "math/SparseMatrixCSRSymmetric.h"
 #include "mechanics/structures/StructureOutputBlockVector.h"
 
-//! @brief constructor
-//! @param mDimension number of nodes
 NuTo::NewmarkBase::NewmarkBase (StructureBase* rStructure)
     : TimeIntegrationBase (rStructure)
 {
-	mMuDampingMass = 0.;
-	mToleranceForce = 1e-6;
-	mMaxNumIterations = 20;
-	mBeta = 0.25;
-	mGamma = 0.5;
-	mInternalEnergy = 0.;
-	mExternalEnergy = 0.;
-	mKineticEnergy = 0.;
-	mDampedEnergy = 0.;
-
-	mUseLumpedMass = false;
 }
 
-//! @brief merges the dof values depending on the numTimeDerivatives and rMergeAll
-//! @param rDof_dt0 ... 0th time derivative
-//! @param rDof_dt1 ... 1st time derivative
-//! @param rDof_dt2 ... 2nd time derivative
-//! @param rMergeAll ... false: merges dof_dt1 only when mMuMassDamping = 0, ignores dof_dt2
 void NuTo::NewmarkBase::MergeDofValues(const StructureOutputBlockVector& rDof_dt0, const StructureOutputBlockVector& rDof_dt1, const StructureOutputBlockVector& rDof_dt2, bool rMergeAll)
 {
     mStructure->NodeMergeDofValues(0, rDof_dt0.J, rDof_dt0.K);
+
     if (mStructure->GetNumTimeDerivatives() >= 1)
     {
-        if (rMergeAll || mMuDampingMass == 0)
+        if (rMergeAll or mMuDampingMass == 0)
         {
             mStructure->NodeMergeDofValues(1, rDof_dt1.J, rDof_dt1.K);
         }
@@ -73,11 +56,6 @@ void NuTo::NewmarkBase::MergeDofValues(const StructureOutputBlockVector& rDof_dt
 }
 
 
-//! @brief ... Info routine that prints general information about the object (detail according to verbose level)
-void NuTo::NewmarkBase::Info()const
-{
-	TimeIntegrationBase::Info();
-}
 
 #ifdef ENABLE_SERIALIZATION
 // serializes the class
--- a/src/mechanics/timeIntegration/NewmarkBase.h
+++ b/src/mechanics/timeIntegration/NewmarkBase.h
@@ -29,59 +29,14 @@ public:
     	mMuDampingMass = rMuDampingMass;
     }
 
-    double GetDampingCoefficientMass()const
-    {
-    	return mMuDampingMass;
-    }
-
     void SetToleranceForce(double rToleranceForce)
     {
     	mToleranceForce = rToleranceForce;
     }
 
-    double GetToleranceForce()const
-    {
-    	return mToleranceForce;
-    }
-
     void SetMaxNumIterations(int rMaxNumIterations)
     {
     	mMaxNumIterations = rMaxNumIterations;
-    }
-
-    int GetMaxNumIterations()const
-    {
-    	return mMaxNumIterations;
-    }
-
-    void SetNewmarkBeta(double rBeta)
-    {
-    	mBeta = rBeta;
-    }
-
-    double GetNewmarkBeta()const
-    {
-    	return mBeta;
-    }
-
-    void SetNewmarkGamma(double rGamma)
-    {
-    	mGamma = rGamma;
-    }
-
-    double GetNewmarkGamma()const
-    {
-    	return mGamma;
-    }
-
-    bool GetUseLumpedMass()const
-    {
-    	return mUseLumpedMass;
-    }
-
-    void SetUseLumpedMass(bool rUseLumpedMass)
-    {
-    	mUseLumpedMass = rUseLumpedMass;
     }
 
     //! @brief merges the dof values depending on the numTimeDerivatives and rMergeAll
@@ -102,8 +57,6 @@ public:
 #endif// SWIG
 #endif // ENABLE_SERIALIZATION
 
-    //! @brief ... Info routine that prints general information about the object (detail according to verbose level)
-    void Info()const override;
 
 
 protected:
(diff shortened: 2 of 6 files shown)


src/mechanics/timeIntegration/NewmarkBase.h
```diff
@@ -29,61 +29,16 @@ class NewmarkBase : public TimeIntegrationBase
...
-    }
-
-    bool GetUseLumpedMass()const
-    {
-    	return mUseLumpedMass;
-    }
-
-    void SetUseLumpedMass(bool rUseLumpedMass)
-    {
-    	mUseLumpedMass = rUseLumpedMass;
```
TTitscher, line 84: If no one uses the lumped mass, can we get rid of Structure::BuildGlobalHessian2Lumped() as well?

src/mechanics/timeIntegration/NewmarkBase.h
```diff
@@ -114,19 +67,22 @@ class NewmarkBase : public TimeIntegrationBase
...
-	double mInternalEnergy;
-	double mExternalEnergy;
-	double mKineticEnergy;
-	double mDampedEnergy;
-
-	bool mUseLumpedMass;
+	double mMuDampingMass = 0;
+
+        //NewtonRaphson parameters
+    double mToleranceForce = 1.e-6;
```
TTitscher: Tiny mess up in the indentation?

src/mechanics/timeIntegration/TimeIntegrationBase.cpp
```diff
@@ -101,16 +81,11 @@ void NuTo::TimeIntegrationBase::SetTimeDependentLoadCase(int rTimeDependentLoadC
     mTimeDependentLoadCase = rTimeDependentLoadCase;
```
TTitscher, line 81: I thought we removed the load case. Should we clean it up here as well?
E.g. remove the member TimeIntegration::mTimeDependentLoadCase completely?

src/mechanics/timeIntegration/TimeIntegrationBase.h
```diff
@@ -184,47 +169,13 @@ class TimeIntegrationBase
...
-    //! @brief returns if automatic time stepping is turned on
-    bool GetAutomaticTimeStepping()const
-    {
-    	return mAutomaticTimeStepping;
-    }
-
-    //! @brief sets the coefficient matrix check (on or off)
-    void SetCheckCoefficientMatrix(bool rCheckCoefficientMatrix)
-    {
-        mCheckCoefficientMatrix = rCheckCoefficientMatrix;
```
TTitscher, line 208: If you remove both getter and setter for a member, you can remove the member as well, right?
mCheckCoefficientMatrix and mExportDataFileNodes

src/mechanics/timeIntegration/TimeIntegrationBase.h
```diff
@@ -238,10 +189,6 @@ class TimeIntegrationBase
         mCallback = rCallback;
     }
 
-    int GetIterationCount() const
-    {
-        return mIterationCount;
```
TTitscher, line 243: I think I use this method in my main files from time to time. If so, I will commit a test soon.

TTitscher (commented):
I am not sure about the initialization of the members. If the others are fine with that, I am too. 
As for the unused code: That is a valid start for refactoring. In the future, I would like to see (and participate in) a split in more meaningful classes #49.

src/mechanics/timeIntegration/NewmarkBase.h
joergfunger: This lumped feature is used in the Runge-Kutta explicit schemes

    StructureOutputBlockMatrix hessian2 = mStructure->BuildGlobalHessian2Lumped();

src/mechanics/timeIntegration/NewmarkBase.cpp
```diff
@@ -28,35 +28,18 @@
...
-	mGamma = 0.5;
-	mInternalEnergy = 0.;
-	mExternalEnergy = 0.;
-	mKineticEnergy = 0.;
-	mDampedEnergy = 0.;
-
-	mUseLumpedMass = false;
 }
 
-//! @brief merges the dof values depending on the numTimeDerivatives and rMergeAll
```
joergfunger, line 49: Did we agree on putting all doxygen code only in the headers? If so, we should communicate this.

src/mechanics/timeIntegration/NewmarkBase.h
TTitscher: Right! But maybe, we can get rid of the option. Runge-Kutta uses it by default and Newmark is currently not using it at all, at least not in the covered tests.
https://codecov.io/gh/nutofem/nuto/src/master/src/mechanics/timeIntegration/NewmarkDirect.cpp#L161

But this is the problem with the coverage. Maybe someone uses it in his/her examples...

src/mechanics/timeIntegration/NewmarkBase.cpp
```diff
@@ -73,11 +56,6 @@ void NuTo::NewmarkBase::MergeDofValues(const StructureOutputBlockVector& rDof_dt
 }
```
joergfunger, line 58: Apparently nobody is using this, but there could be some information regarding beta/gamma ... that might be of interest.

src/mechanics/timeIntegration/NewmarkBase.h
```diff
@@ -29,61 +29,16 @@ class NewmarkBase : public TimeIntegrationBase
...
     {
     	mMaxNumIterations = rMaxNumIterations;
     }
 
-    int GetMaxNumIterations()const
-    {
-    	return mMaxNumIterations;
-    }
-
-    void SetNewmarkBeta(double rBeta)
```
joergfunger, line 57: Why is setting/getting the parameters for beta/gamma removed? There are other schemes options possible
[see Wikipedia](https://en.wikipedia.org/wiki/Newmark-beta_method)

src/mechanics/timeIntegration/NewmarkDirect.h
```diff
@@ -30,26 +30,11 @@ class NewmarkDirect : public NewmarkBase
...
-    {
-    	return mMinLineSearchStep;
-    }
-
     void SetPerformLineSearch(bool rPerformLineSearch)
     {
         mPerformLineSearch = rPerformLineSearch;
     }
 
-    bool GetPerformLineSearch() const
```
joergfunger, line 48: Why are Get Methods removed? We are currently not using the serialize feature (but at least plan to). In this case, after deserialization (or when a restore with a restart file is used), it would make sense to determine the actual state of these member variables.

joergfunger (approved):
We should discuss if want to remove all getters for parameters that are usually set by the user.

src/mechanics/timeIntegration/NewmarkBase.cpp
phuschke: Not all doxygen code. `\todo` certainly makes sense in source files while \brief, \return, \param, etc. do not imo.

src/mechanics/timeIntegration/NewmarkBase.cpp
phuschke: I agree but as of right now this function does nothing

src/mechanics/timeIntegration/NewmarkBase.h
phuschke: 1. Nobody uses other values 
2. Do we even know if our implementation works for other values?

src/mechanics/timeIntegration/NewmarkDirect.h
phuschke: Well, in that case we should probably implement a method that _streams_ all necessary information and not just a function that returns a copy or reference to that variable.

src/mechanics/timeIntegration/NewmarkDirect.h
```diff
@@ -30,26 +30,11 @@ class NewmarkDirect : public NewmarkBase
     //! @brief constructor
     NewmarkDirect(StructureBase* rStructure);
 
-    void SetMinLineSearchStep(double rMinLineSearchStep)
```
Psirus, line 33: I didn't know such a method existed! It may not be used atm, but it's not useless though. After this change, I would have to recompile Newmark if I wanted to change that.

src/mechanics/timeIntegration/NewmarkBase.cpp
Psirus: I think we did agree, and we have added this some time ago to the "developer documentation": http://nuto.readthedocs.io/en/master/CodingStyle.html

src/mechanics/timeIntegration/TimeIntegrationBase.h
```diff
@@ -309,21 +256,21 @@ class TimeIntegrationBase
...
+	double mTime = 0.;
     //adapt the time step based on the number of iterations required (or decrease, if no convergence can be achieved)
-	bool mAutomaticTimeStepping;
+	bool mAutomaticTimeStepping = false;
 	//initial time step (or at the end this is the last time step used)
-	double mTimeStep;
-    //maximum time step (for adaptive simulations)
-	double mMaxTimeStep;
-    //minimum time step (for adaptive simulations)
-	double mMinTimeStep;
```
Psirus, line 320: Instead of using normal `//` comments on the line above, these should be inline doxygen comments `double mTime = 0.; //!< accumulated time`. I might add a commit to that effect later.

Psirus (approved):
The initialization of members at their declaration is fine by me.

PR #54, opened 2017-06-16
Fix dirichlet precond on travis

phuschke:


src/mechanics/feti/NewmarkFeti.h
```diff
@@ -805,55 +805,51 @@ class NewmarkFeti : public NewmarkDirect
         }
         case eFetiPreconditioner::Dirichlet:
```
TTitscher, line 806: Have you considered a `class Preconditioner` instead of enums?

Psirus:
Have you considered using a virtual machine running Ubuntu 14.04 to debug this? Seems like this could greatly improve the speed of your debug cycle.

PR #69, opened 2017-06-28
Changed Lobatto Integration to variable order version

pmueller2:
IntegrationType Lobatto takes number of integration points as parameter in the constructor.
Lobatto weights and nodes are calculated instead of looking them up.

The enums are still there, which maybe should be changed also, if this is a useful addition.

Similar changes could be made to Gauss integration.

Is this useful? What do you think?


src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
+#include <stdlib.h>
+#include <iostream>
+#include <vector>
+#include <math.h>
+#include "boost/math/special_functions/legendre.hpp"
+#include <algorithm>
+#include "math/Polynomial.h"
+
+double NuTo::Math::Polynomial::legendre_p_deriv(int n, double x, int k) {
```
TTitscher: Please stick to our code style - method names in CamelCase.

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
+#include <stdlib.h>
+#include <iostream>
+#include <vector>
+#include <math.h>
```
TTitscher: #include <cmath> instead. <math.h> is the c version
https://stackoverflow.com/questions/15656434/include-cmath-vs-include-math-h-in-a-c-program

math.h defines methods globally, and cmath in the namespace `std`. So you'll have to modify the math methods

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
+#include <stdlib.h>
```
TTitscher: c header, remove that

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <iostream>
+#include <vector>
+#include <math.h>
+#include "boost/math/special_functions/legendre.hpp"
+#include <algorithm>
+#include "math/Polynomial.h"
+
+double NuTo::Math::Polynomial::legendre_p_deriv(int n, double x, int k) {
+    if (n < k) {
```
TTitscher: Again code style, opening and closing brackets in a new line. Run clang-format on your code, there is a tutorial in our documentation https://github.com/nutofem/nuto/blob/master/doc/dev/CodingStyle.md

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
...
+        return(0.);
+    } else if (n==k) {
+        double fac_k = 1.;
+        double fac_2k = 1.;
+        double pow2_k = 1.;
+        for (int i=1; i<=k;i++) {
+            fac_k *= i;
+            pow2_k *= 2;
+        }
+        for (int i=1; i<= (2*k);i++) {
```
TTitscher: Please implement a generic factorial function (including a simple test) to make the code more expressive.
Same goes for the power function. You could use std::pow (only defined for floating point types) or move this code into a helper function

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
...
+double NuTo::Math::Polynomial::legendre_p_deriv(int n, double x, int k) {
+    if (n < k) {
+        return(0.);
+    } else if (n==k) {
+        double fac_k = 1.;
+        double fac_2k = 1.;
+        double pow2_k = 1.;
+        for (int i=1; i<=k;i++) {
+            fac_k *= i;
+            pow2_k *= 2;
```
TTitscher: why is fac_k, fac_2k and pow2_k not an integer type?

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
...
+    for(int i=1; i<n; i++) {
+        x.push_back(0.5 * (y[i-1] + y[i]) );
+    }
+    // newton iterations
+    double dx;
+    double tol = 1e-8;
+    for(int i=0; i<(n-1); i++)
+    {
+        double xnew;
+        do
```
TTitscher: You implement the newton algorithm twice.

src/math/Polynomial.h
```diff
@@ -0,0 +1,29 @@
...
+namespace NuTo {
+namespace Math
+{
+
+class Polynomial
+{
+public:
+    //! value of the k-th derivative of legendre polynomial
+    //! of the first kind of order n
+    static double legendre_p_deriv(int n, double x , int k);
```
TTitscher: Please add proper doxygen documentation. What is x?

src/mechanics/integrationtypes/IntegrationType1D2NLobatto.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+
+//! @brief returns the total number of integration points for this integration type
+//! @return number of integration points
+int NuTo::IntegrationType1D2NLobatto::GetNumIntegrationPoints()const
+{
+    return nIps;
+}
+
+//! @brief returns the weight of an integration point
```
TTitscher: Remove doxygen comments in the cpp files.

src/mechanics/integrationtypes/IntegrationType1D2NLobatto.h
```diff
@@ -1,4 +1,6 @@
-#pragma once
+// $Id$
+#ifndef IntegrationType1D2NLobatto_H
```
TTitscher: #pragma once is our style

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <vector>
+#include <math.h>
+#include "boost/math/special_functions/legendre.hpp"
+#include <algorithm>
+#include "math/Polynomial.h"
+
+double NuTo::Math::Polynomial::legendre_p_deriv(int n, double x, int k) {
+    if (n < k) {
+        return(0.);
+    } else if (n==k) {
```
TTitscher: I personally like to keep the code as flat as possible. If your `if` block returns something, you can continue without an `else`. In general, this would have been much shorter and more readable:
~~~cpp
if (n < k)
   return 0;
if (n == k)
   return Factorial(2*k) / (Power(2, k) * Factorial(k));
return ( ( (2.*n-1.)*x* legendre_p_deriv(n-1,x, k) - (n-1.+k) * legendre_p_deriv(n-2,x, k) )/(n-k)
~~~

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
...
+        }
+        return( fac_2k/(pow2_k * fac_k) );
+    } else {
+        return( ( (2.*n-1.)*x* legendre_p_deriv(n-1,x, k) - (n-1.+k) * legendre_p_deriv(n-2,x, k) )/(n-k) );
+    }
+}
+
+std::vector<double> NuTo::Math::Polynomial::legendre_roots(int n) {
+    // first guess
+    std::vector<double> x;
```
TTitscher: it is generally good practice to allocate memory in the `std::vector` at once, whenever it is possible. Now, the push_back resizes the vector multiple times. (In this case, performance does not matter, but still, good practice...). You can either call a x.reserve(n) or the ctor vector<double> x(n).

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
...
+    for(int i=1; i<=n; i++) {
+        x.push_back(cos(M_PI*(4.*i-1)/(4.*n + 2)));
+    }
+    // newton iterations
+    double dx;
+    double tol = 1e-16;
+    for(int i=0; i<n; i++)
```
TTitscher: I suggest renaming the vector x to xs or some other kind of plural term. Then use a range based for loop
~~~cpp
for (double x : xs)
~~~
instead of x[i] all over the place.

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
...
+    for(int i=1; i<=n; i++) {
+        y.push_back(cos(M_PI*(4.*i-1)/(4.*n + 2)));
+    }
+    std::vector<double> x;
+    for(int i=1; i<n; i++) {
+        x.push_back(0.5 * (y[i-1] + y[i]) );
+    }
+    // newton iterations
+    double dx;
+    double tol = 1e-8;
```
TTitscher: Why 1e-8? Is that enough?

src/mechanics/integrationtypes/IntegrationType1D2NLobatto.cpp
```diff
@@ -0,0 +1,102 @@
...
+    nIps = numIps;
+    std::vector<double> innerNodes = NuTo::Math::Polynomial::legendre_deriv_roots(numIps-1);
+    iPts.push_back(-1.);
+    for (double value : innerNodes) {
+        iPts.push_back(value);
+    }
+    iPts.push_back(1.);
+
+    if (numIps > 1) {
+        weights.push_back(2./(nIps*(nIps-1)) );
```
TTitscher: You can resize/reserve weights before pushing into it.

src/mechanics/integrationtypes/IntegrationType1D2NLobatto.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+//! @brief returns the local coordinates of an integration point
+//! @param rIpNum integration point (counting from zero)
+//! @param rCoordinates (result)
+Eigen::VectorXd NuTo::IntegrationType1D2NLobatto::GetLocalIntegrationPointCoordinates(int rIpNum) const
+{
+    if(rIpNum >= 0 && rIpNum < nIps)
+        return Eigen::Matrix<double, 1, 1>::Constant(iPts[rIpNum]);
+    else
```
TTitscher: Personally, i would delete the else since the corresponding if returns.

src/mechanics/integrationtypes/IntegrationType1D2NLobatto.h
```diff
@@ -68,14 +71,16 @@ class IntegrationType1D2NLobatto5Ip : public IntegrationType1D
         std::vector<unsigned int>& VisualizationCellsIP) const override;
 #endif // ENABLE_VISUALIZE
 private:
+    int nIps;
     //! @brief ... integration points coordinates
-    double iPts[5];
+    std::vector<double> iPts;
     //! @brief ... weights for the integration
-    double weights[5];
+    std::vector<double> weights;
```
TTitscher: Code style. All members should start with `m`.

src/mechanics/integrationtypes/IntegrationType1D2NLobatto.h
```diff
@@ -68,14 +71,16 @@ class IntegrationType1D2NLobatto5Ip : public IntegrationType1D
         std::vector<unsigned int>& VisualizationCellsIP) const override;
 #endif // ENABLE_VISUALIZE
 private:
+    int nIps;
```
TTitscher: Maybe we could remove this member and return the size of on of the other members instead.

src/math/Polynomial.cpp
```diff
@@ -0,0 +1,79 @@
+#include <stdlib.h>
+#include <iostream>
+#include <vector>
+#include <math.h>
+#include "boost/math/special_functions/legendre.hpp"
```
TTitscher: You do not need this include, right?

src/math/Polynomial.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <vector>
+
+namespace NuTo {
+namespace Math
+{
+
+class Polynomial
```
TTitscher: IMO the name `Polynomial` is misleading, it's more like `Legendre`, right?

And please provide a unit test, e.g. a comparison to wikipedia values of the gauss-lobatto points.

src/mechanics/integrationtypes/IntegrationType1D2NLobatto.cpp
```diff
@@ -0,0 +1,102 @@
...
+    for (double value : innerNodes) {
+        iPts.push_back(value);
+    }
+    iPts.push_back(1.);
+
+    if (numIps > 1) {
+        weights.push_back(2./(nIps*(nIps-1)) );
+        for (int i=1; i<nIps-1;i++) {
+            double lp = boost::math::legendre_p(nIps-1,iPts[i]);
+            weights.push_back(2./(nIps*(nIps-1))/( lp * lp) );
```
TTitscher: You can think about moving this code to `Polynomial` as well.

src/mechanics/integrationtypes/IntegrationType1D2NLobatto.cpp
```diff
@@ -0,0 +1,102 @@
...
+// constructor
+NuTo::IntegrationType1D2NLobatto::IntegrationType1D2NLobatto(int numIps)
+{
+    nIps = numIps;
+    std::vector<double> innerNodes = NuTo::Math::Polynomial::legendre_deriv_roots(numIps-1);
+    iPts.push_back(-1.);
+    for (double value : innerNodes) {
+        iPts.push_back(value);
+    }
+    iPts.push_back(1.);
```
TTitscher: You can simplify this code a lot by inserting the `(-1)` at the front of the vector.
~~~cpp
nIps = NuTo::Math::Polynomial::legendre_deriv_roots(numIps-1);
nIps.insert(nIps.begin(),-1.);
nIps.push_back(1.);
~~~

TTitscher (changes requested):
I like this change very much. Increase functionality while reducing the code massively. Please work on the code style changes I suggested. 

One more idea: Can we make it more static? A template parameter <TDegree>? With the recent changes in the language and c++14 constexpr, can we even make the calculations constexpr?

src/math/Polynomial.h
Psirus: This class has no members and only static functions. [It should be a namespace with free functions instead](https://stackoverflow.com/questions/1434937/namespace-functions-versus-static-methods-on-a-class).

src/mechanics/integrationtypes/IntegrationType1D2NLobatto.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "visualize/VisualizeEnum.h"
+#endif // ENABLE_VISUALIZE
+
+#include "mechanics/integrationtypes/IntegrationType1D2NLobatto.h"
+#include "mechanics/elements/ElementShapeFunctions.h"
+#include "boost/math/special_functions/legendre.hpp"
+#include "math/Polynomial.h"
+
+// constructor
+NuTo::IntegrationType1D2NLobatto::IntegrationType1D2NLobatto(int numIps)
```
Psirus: Use direct initialization for `nIps`, i.e `IntegrationType1D2NLobatto(int numIps) : nIps(numIps) { ... }` and then just use `nIps` throughout the constructor. You're using `numIps` and `nIps` seemingly at random.

Psirus (changes requested):
I think that's a good idea. Just to add to the things that @TTitscher has already mentioned: please add unit tests for the Polynomial and the IntegrationType.

test/math/CMakeLists.txt
```diff
@@ -17,6 +17,10 @@ add_unit_test(SparseMatrixCSRVector2General
         math/SparseMatrixCSR.cpp
     )
 
+add_unit_test(Polynomial
+        math/Polynomial.cpp
```
Psirus: This should be redundant. The `add_unit_test` macro should automatically add the corresponding cpp file.

test/math/Polynomial.cpp
```diff
@@ -0,0 +1,164 @@
...
+            0.83603110732663521,
+            0.96816023950762675
+        }
+    };
+
+  int maxOrder = 9;
+
+  for (int order = 1; order <= maxOrder; order++) {
+    std::vector<double> roots = NuTo::Math::Polynomial::LegendreRoots(order);
+    for (int i=0; i<order; i++) {
```
Psirus: You can use `BoostUnitTest::CheckVector` instead of this for loop.

Psirus (approved):
Nicely done!

test/math/Polynomial.cpp
TTitscher: This is a problem with values very close to zero, since BOOST_CHECK_CLOSE calculates the relative error, possibly dividing by zero

TTitscher (approved):

potto1:
In the IGA branch the Gauss integration class with variable number of IPs is implemented as template (https://github.com/nutofem/nuto/blob/iga/src/nuto/mechanics/integrationtypes/IntegrationType1D2NGaussNIp.h). I think we have to agree in the question, whether use templates or argument in the constructor.

TTitscher:
One more remark: Can we think about renaming `Polynomial.h/cpp` to `Legendre.h/cpp`? Its functions can stay in the `namespace Polynomial`. Imagine adding other types of polynomials, whatever that might be. But they can go in a separate `h/cpp` file, also in the `namespace Polynomial`.

Psirus:
1. Unless we manage something `constexpr` where the computation is done at compile time, I'd say we use a "normal" parameter. Nothing gained by using templates except rigidity / no proper python support.
2. Yes, let's rename it to Legendre.

TTitscher:
Nice work with the Gauss integration as well! Is there a way to generate these points for higher dimensions as well? I can imagine that 2DQuad and 3DBrick are tensor products of the underlying 1D type. How does it work for Triangle/Tetrahedron?

PR #70, opened 2017-06-29
Newton raphson

TTitscher:
Here is an idea for a generic Newton-Raphson algorithm as a free function. The implementation relies on _a lot_ of template arguments for defining
- residual function
- derivative function
- norm function
- tolerance value
- optional info method
all collected in NuTo::NewtonRaphson::Problem

Please have a look at https://github.com/nutofem/nuto/blob/newton_raphson/test/math/NewtonRaphson.cpp to see its application to both a scalar valued and a vector valued problem.

Is this tooo cryptic all in all?

### DIFF
--- a/applications/benchmark/CMakeLists.txt
+++ b/applications/benchmark/CMakeLists.txt
@@ -40,3 +40,6 @@ if(ENABLE_GEOMETRYCONCRETE)
     target_link_libraries(DenseSpherePacking NuToMath NuToBase NuToGeometryConcrete ${Boost_LIBRARIES})
     add_test(benchmark::DenseSpherePacking ${CMAKE_CURRENT_BUILD_DIR}/DenseSpherePacking)
 endif()
+
+add_executable(NewtonRaphsonBenchmark NewtonRaphsonBenchmark.cpp)
+target_link_libraries(NewtonRaphsonBenchmark -lprofiler) 
--- a/applications/benchmark/CMakeLists.txt
+++ b/applications/benchmark/CMakeLists.txt
@@ -13,9 +13,9 @@ foreach(test
     )
 
     add_executable(${test} ${test}.cpp)
-    target_link_libraries(${test} NuToMechanics NuToMath NuToBase -lprofiler)
+    target_link_libraries(${test} NuToMechanics NuToMath NuToBase)
     if(MUMPS_FOUND)
-        target_link_libraries(${test} -lprofiler ${MUMPS_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
+        target_link_libraries(${test} ${MUMPS_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
     endif()
     if(ENABLE_VISUALIZE)
         target_link_libraries(${test} NuToVisualize)
@@ -42,4 +42,4 @@ if(ENABLE_GEOMETRYCONCRETE)
 endif()
 
 add_executable(NewtonRaphsonBenchmark NewtonRaphsonBenchmark.cpp)
-target_link_libraries(NewtonRaphsonBenchmark -lprofiler) 
+target_link_libraries(NewtonRaphsonBenchmark) 
--- a/src/math/Average.h
+++ b/src/math/Average.h
@@ -8,7 +8,7 @@ struct Average
     long double mCurrentAverage = 0;
     int mNum = 0;
 
-    void Add(int x)
+    void operator()(int x)
     {
         mCurrentAverage = (x + mNum * mCurrentAverage)/(mNum + 1);
         mNum++;
--- a/src/math/Average.h
+++ b/src/math/Average.h
@@ -1,0 +1,17 @@
+#pragma once
+namespace NuTo
+{
+
+//! @brief calculates the cumulative moving average
+struct Average
+{
+    long double mCurrentAverage = 0;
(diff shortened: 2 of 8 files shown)


applications/benchmark/CMakeLists.txt
```diff
@@ -40,3 +40,6 @@ if(ENABLE_GEOMETRYCONCRETE)
     target_link_libraries(DenseSpherePacking NuToMath NuToBase NuToGeometryConcrete ${Boost_LIBRARIES})
     add_test(benchmark::DenseSpherePacking ${CMAKE_CURRENT_BUILD_DIR}/DenseSpherePacking)
 endif()
+
+add_executable(NewtonRaphsonBenchmark NewtonRaphsonBenchmark.cpp)
+target_link_libraries(NewtonRaphsonBenchmark -lprofiler)
```
Psirus: I don't think we've agreed on making the google profile library a dependency. And if we had, this would not be the "proper way" to link to it.

src/math/Average.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+namespace NuTo
+{
+
+//! @brief calculates the cumulative moving average
+struct Average
```
Psirus, line 6: What does this have to do with the Newton-Raphson? Where do we use it?

src/math/LineSearch.h
```diff
@@ -0,0 +1,92 @@
...
+    //! @brief actual line search implementation
+    //! @remark several return arguments are used to help inlining everything. A lot of time went into actual
+    //! benchmarking. Dunno why, but the a previous version that avoids the return arguments and returns various values
+    //! in a std::tuple was significantly slower (10%) in a benchmark for a scalar function
+    //! @param problem ... class that implements NormFunction, ResidualFunction and mTolerance
+    //! @param r ... residual, return argument, see remark
+    //! @param x ... value of the argument x, return argument, see remark
+    //! @param dx ... dx from the solver
+    template <typename TProblem, typename TX>
+    bool operator()(TProblem&& problem, TX* r, TX* x, TX dx) const
```
Psirus, line 39: I'm guessing there is a reason why these are pointers, but I can't guess what it is. Can you help me?

src/math/NewtonRaphson.h
```diff
@@ -0,0 +1,84 @@
...
+    TDR DerivativeFunction;
+    TNorm NormFunction;
+    TTol mTolerance;
+    TInfo InfoFunction;
+};
+
+//! @brief defines the problem, basically just to enable automatic template deduction. If you
+//! create a Problem directly, you'll have to specify each template parameter. This methods avoids it.
+template <typename TR, typename TDR, typename TNorm, typename TTol, typename TInfo = VoidInfo>
+auto DefineProblem(TR residual, TDR derivative, TNorm norm, TTol tolerance, TInfo info = VoidInfo())
```
Psirus, line 49: Cool pattern, haven't seen that before.

src/math/NewtonRaphson.h
```diff
@@ -0,0 +1,84 @@
...
+//! @brief solves the Problem using the newton raphson iteration with linesearch
+//! @param problem type of the nonlinear problem
+//! @param x0 of the initial value for the iteration
+//! @param solver solver that provides a TX = solver.Solve(TNonlinearProblem::DR, TNonlinearProblem::R)
+//! @param maxIterations default = 20
+//! @param lineSearch line search algorithm, default = NoLineSearch, alternatively use NuTo::LineSearch()
+//! @param numIterations optionally returns the number of iterations required 
+template <typename TNonlinearProblem, typename TX, typename TSolver, typename TLineSearchAlgorithm = NoLineSearch>
+auto Solve(TNonlinearProblem&& problem, TX&& x0, TSolver&& solver, int maxIterations = 20,
+            TLineSearchAlgorithm&& lineSearch = NoLineSearch(), int* numIterations = nullptr)
```
Psirus, line 63: I don't think `numIterations` are actually set anywhere.

Psirus (approved):
Good idea, well implemented, thoroughly tested. What's not to like? How likely do you think it is that we'll replace the alternate implementations strewn about?

src/math/NewtonRaphson.h
TTitscher: Oopsy daisy, i'll fix that.

src/math/NewtonRaphson.h
TTitscher: This _work around_ will be gone in c++17 with [class template deduction](http://en.cppreference.com/w/cpp/language/class_template_deduction).

src/math/LineSearch.h
TTitscher: I like the syntax that is required by the caller. You'll have to call this method with
~~~cpp
linesearch(problem, &r, &x, dx);
~~~
indicating that `r` and `x` are return arguments. This would be hidden otherwise.

pmueller2 (approved):
Nothing to add, except a little question: If I have to use this to find the complex roots of a function (using std::complex) there are only minor additions to make right?

src/math/Average.h
TTitscher: Dunno. I wrote it to calculate the average number of newton iterations. Say, if you run newmark and want to know the average number of newton raphson iterations per time step. IMO, it doesnt hurt.

PR #71, opened 2017-06-30
Remove superflous exception types

Psirus:
1. `MetalmodelException`, `MathException`, `MechanicsException`,
`OptimizeException` and `VisualizeException` did nothing different than the
vanilla `Exception`, and no one was catching them polymorphically. So they've been removed.
2. The `Exception` has been streamlined, i.e. no more `mFatalFlag` and `AddMessage`. In the process, a lot of the "stracktrace" catch-try blocks were removed.

Any objections?


src/geometryConcrete/GeometryConcrete.cpp
```diff
@@ -49,21 +49,7 @@ void NuTo::GeometryConcrete::MaximizeParticleDistance(double rParticleDistance)
...
-        if (mContinueOnException)
-        {
-            std::cout << e.ErrorMessage() << "\n but I'll continue.";
-        }
-        else
-        {
-            throw;
-        }
-    }
+    handler.Simulate(mNumEventsMax, timeEnd, mSecondsWallTimeMax, mSecondsPrint, mInitialTimeBarrier);
```
TTitscher: !!! You changed the logic flow. `if (mContinueOnException)`, the code does __not rethrow__.

TTitscher (approved):
I (accidentally [really big change]) found a crucial change for the geometry concrete. Not your fault, no test was provided. I'll add the test and fix the code. But can you remember other similar branches in the code that you modified? 

Apart from that, I like the changes.

src/geometryConcrete/GeometryConcrete.cpp
Psirus: You're right, I'm sorry. I didn't even notice it did something beyond the usual `AddMessage`. I think it's only in this file, but in two locations.

phuschke (approved):

PR #72, opened 2017-06-30
Replace jsoncpp with nlohmann/json

Psirus:
We're using [JsonCpp](https://github.com/open-source-parsers/jsoncpp) for parsing Json files. This PR replaces it with this [JSON for Modern C++](https://github.com/nlohmann/json) library. The reasons for doing so are as follows: 

1. Speed/Memory use/Code size: According to [this benchmark](https://github.com/miloyip/nativejson-benchmark), the "Nlohmann" (the new one) is faster, uses less memory and results in smaller executables than "JsonCpp".

2. Conformance to the standard is higher. This is also the reason I first tried it, because a JSON file I've created (and that Python and "JSON for Modern C++" handle just fine) could for some reason not be parsed by JsonCpp.

3. Nicer syntax. C++ is able to deduce the return type in most cases automatically, so in most cases lines that previously read `double x = jsonTree["x"].asDouble();` will now simply be `double x = jsonTree["x"];`. In the few cases where the compiler is unable to deduce the type, you can explicitly request any type by e.g. `jsonTree["y"].get<int>()`.

4. Header-only: no need to build and link to a seperate library, just `#include "json.hpp"`. I'm trying to create a [EXPORT Cmake file](https://cmake.org/cmake/help/v3.0/command/install.html) where the `jsoncpp` target is a hindrance. This is not actually a fault of the library, but if I can circumvent that and get 1.-3. for free, I'll take it.

What do you think?


TTitscher (approved):
I like that it is header-only. I guess you just copied the headers into our project. What about somehow _linking_ the corresponding github repository?

Psirus:
Off the top of my head, I'd say there are two ways of going about this: using [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules), which is probably what you are referring to, or start using [Conan](https://www.conan.io/), which is like `pip`, `npm` or `gem`, just for C++. I have not properly looked at either option.

phuschke (approved):
Great. Thank you. 

Git submodules would be nice because there are still recent contributions that improve the code (last commit was 13 days ago).

Issue #21, opened 2016-12-23
Activate more compiler warnings


Psirus:
There are additional warnings that can be enabled in gcc and clang. In particular, I'm thinking of
`-Wextra` in both clang and gcc, which warns if
> - a base class is not initialized in the copy constructor of a derived class,
> - an empty body occurs in an if, else or do while statement,
> - the return type of a function has a type qualifier such as `const`
> - an automatic variable is used without first being initialized
> - a function parameter is unused aside from its declaration

`-Weffc++` in gcc
> Warn about violations of the following style guidelines from Scott Meyers' Effective C++ series of books:
> 
> - Define a copy constructor and an assignment operator for classes with dynamically-allocated memory.
> - Prefer initialization to assignment in constructors.
> - Have operator= return a reference to *this.
> - Don't try to return a reference when you must return an object.
> - Distinguish between prefix and postfix forms of increment and decrement operators.
> - Never overload &&, ||, or ,.

and `-Wdocumentation` in clang:
> Clang parses Doxygen and non-Doxygen style documentation comments and attaches them to the appropriate declaration nodes. By default, it only parses Doxygen-style comments and ignores ordinary comments starting with // and /*.
> `-Wdocumentation`:
> Emit warnings about use of documentation comments. This warning group is off by default.
> This includes checking that \param commands name parameters that actually present in the function signature, checking that \returns is used only on functions that actually return a value etc.

Both of them would be helpful, I think, but activating them right now, we would be drowning in warning messages. Maybe we can have a small coding sprint for each of them.

- [ ] `-Wdocumentation`
- [x] `-Weffc++`
- [x] `-Wextra`

Psirus:
I took a preliminary look at the output of both settings. Many of the warnings were caused by the `FullMatrix` and the `Structure`, so maybe it's easier just to wait until we've removed them.

phuschke:
**Progress**
- [x] `-Wextra`
Enabled with commit 51dc3eb

- [x] `-Weffc++`

> -Weffc++ is not very useful since it is mostly based on the first edition of the book Effective C++, many of whose recommendations were improved in the second edition
[Best GCC warning flags for compiling C++](http://www.artificialworlds.net/blog/2014/07/18/best-gcc-warning-flags-for-compiling-c/)

**Pending:**
- [ ] `-Wdocumentation`

phuschke:
In case you are wondering why there are still so many warnings. Here is an example:
~~~    template <int TRows>
[...]

PR #74, opened 2017-07-06
Remove -Wextra warnings

szorn85:
See issue #21


TTitscher (approved):
Thumbs up! Thank you for cleaning this stuff up!

Psirus (changes requested):
Nice. The only thing that is missing is actually activating the warning in CMake.

Also, looking through the changes I noticed two things:
1. This pattern of `BaseClass::DerivedClassMethod{ throw "Not Implemented"; }` is really annoying.
2. Why does `GetConstitutiveInputs` need an `InterpolationType` anyway?

Psirus (approved):

PR #79, opened 2017-07-14
Installation of NuTo as an independent library

Psirus:
I'm trying to make NuTo more useable "outside" of NuTo, i.e. building your own mainfiles not in `applications/custom` but rather as separate projects.  It's far from done, and I'll add documentation on how to use the `NuToTargets.cmake` file later. For now, the important change here is that the targets in CMake have been renamed. So when linking your own executables, you now need to say
`target_link_libraries(MyExecutable Mechanics)` rather than `target_link_libraries(MyExecutable NuToMechanics)`. Nothing else has changed for "traditional" NuTo use, so this is just to make you aware of the new target names.


phuschke (approved):
Much clearer and easier to understand. Thank you. Awesome idea with the NuTo cmake namespace.

joergfunger (commented):
Why exactly was it necessary to rename the targets?

Psirus:
It was not strictly necessary, but when the targets are exported, they will be put into a NuTo namespace. So an external project would link to `Mechanics` with `target_link_libraries(MyExecutable NuTo::Mechanics)`. Having `NuTo::NuToMechanics` seemed redundant and inelegant.

TTitscher (approved):
I trust you on this one =)

PR #80, opened 2017-07-14
Time integration refactor

Psirus:
We've tried to refactor some of the stuff that has accumulated in `TimeIntegrationBase`. Specifically:

- controlling the time stepping (next step, min/max time step, automatic <-> fixed, etc.) is now a separate class
- all the post-processing has been moved to a separate class
- refactor Newmark; most importantly splitting the gigantic `Solve` routine into several functions

There is still a lot to be done, but too little time, so we would like to merge this, even though there is still a lot of "unattractive" areas in time integration. However, we're open to any comments and suggestions you may have on improving our attempt.

When reviewing, don't be intimidated by the size of the PR. A lot of this is just API changes. Take a look at `TimeIntegrationBase`, `NewmarkDirect`, `PostProcessor` and `TimeControl`.


src/mechanics/timeIntegration/RungeKuttaBase.h
```diff
@@ -38,7 +38,22 @@ class RungeKuttaBase : public TimeIntegrationBase
...
     virtual bool HasTimeChanged(int rStage) const = 0;
 
+
+    // temporary fix
+    //! @brief sets the  time step for the time integration procedure (initial value)
+    virtual void SetTimeStep(double rTimeStep) override
+    {
+        mTimeStep = rTimeStep;
+    }
+
```
phuschke: What is the _temporary fix_ for? Why did mTime and mTimeStep moved from TimeIntegrationBase?

src/mechanics/timeIntegration/TimeControl.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+
+
+void NuTo::TimeControl::ScaleTimeStep(double scaleFactor)
+{
+    if (scaleFactor<= 0.0)
+        throw MechanicsException(__PRETTY_FUNCTION__,"Scaling factor must be a positive number");
+
+    mTimeStepScaleFactor *= scaleFactor;
+    UpdateTimeStep(-1.,-1.,true);
```
phuschke: Why are iterations and maxIterations doubles? What does maxIteration=-1. mean?
`UpdateTimesStep(double iterations, double maxIterations, bool)`

src/mechanics/timeIntegration/TimeIntegrationBase.h
```diff
@@ -195,6 +122,14 @@ class TimeIntegrationBase
...
     {
         mAutomaticTimeStepping = rAutomaticTimeStepping;
+        if(mAutomaticTimeStepping)
+        {
+            mTimeControl.UseDefaultAutomaticTimestepping();
+        }
+        else
+        {
+
+        }
```
phuschke: Is something missing here?

src/mechanics/timeIntegration/RungeKuttaBase.h
```diff
@@ -38,7 +38,22 @@ class RungeKuttaBase : public TimeIntegrationBase
     //! @brief ... return, if time (e.g. for the calculation of external loads) has changed
     virtual bool HasTimeChanged(int rStage) const = 0;
 
+
+    // temporary fix
+    //! @brief sets the  time step for the time integration procedure (initial value)
+    virtual void SetTimeStep(double rTimeStep) override
```
phuschke, line 44: What is the temporary fix for? Why did mTime and mTimeStep moved from TimeIntegrationBase?

phuschke (commented):
Good work. Thank you for improving and cleaning the time integration,

src/mechanics/timeIntegration/NewmarkDirect.cpp
```diff
@@ -1,499 +1,182 @@
...
-void NuTo::NewmarkDirect::Info() const
+void NewmarkDirect::Solve(double rTimeDelta)
 {
-    NewmarkBase::Info();
-}
-
-void NuTo::NewmarkDirect::Solve(double rTimeDelta)
-{
-    NuTo::Timer timerFull(__FUNCTION__, mStructure->GetShowTime(), mStructure->GetLogger());
+    mTimeControl.SetTimeFinal(rTimeDelta);
```
joergfunger: Assume there was a previous solve up to say t=10, and then there is a new solve for another 10 time units. What is rTimeDelta and Final time (10 or 20?).

src/mechanics/timeIntegration/NewmarkDirect.cpp
```diff
@@ -1,499 +1,182 @@
...
-    evaluate_Hessian0_Hessian1[eStructureOutput::HESSIAN0] = &hessian0;
-
-    if (mStructure->GetNumTimeDerivatives() >= 1 && mMuDampingMass == 0.)
+    while (not mTimeControl.Finished())
     {
-        hessian1.Resize(dofStatus.GetNumActiveDofsMap(), dofStatus.GetNumDependentDofsMap());
-        evaluate_InternalGradient_Hessian0Hessian1[eStructureOutput::HESSIAN1] = &hessian1;
-        evaluate_Hessian0_Hessian1[eStructureOutput::HESSIAN1] = &hessian1;
-    }
+        mTimeControl.Proceed();
```
joergfunger, line 44: What exactly does proceed? I think the name is not very intuitive.

src/mechanics/timeIntegration/NewmarkDirect.cpp
```diff
@@ -1,499 +1,182 @@
...
-                                            << "] Trial residual: " << trialNormResidual << "\n";
+            const auto intForce = EvaluateInternalGradient();
 
-                    alpha *= 0.5;
+            structureResidual = CalculateResidual(intForce, extForce, hessians[2], trial_dof_dt1, trial_dof_dt2);
+            residual = Assembler::ApplyCMatrix(structureResidual, constraintMatrix);
 
-                } while (mPerformLineSearch && alpha > mMinLineSearchStep &&
-                         trialNormResidual > (1. - alpha) * normResidual);
+            trialNormResidual = residual.CalculateInfNorm();
```
joergfunger: Is the InfNorm the standard to check convergence? IMO this would mean that by reducing the mesh size the acceptable error in the solution increases.

src/mechanics/timeIntegration/NewmarkDirect.cpp
```diff
@@ -1,499 +1,182 @@
...
-        //! to work fine, but valgrind tells you otherwise. Problem is, that the DofType is changed during the function
-        //! call inside
-        //! the loop which leads to reads in already freed blocks of memory
-        //! !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-        std::set<Node::eDof> currentDofSet = dofStatus.GetDofTypes();
+        // LEAVE IT OR FIX IT IF YOU CAN:
+        // If you dont make a copy of the dof set with dofStatus.GetDofTypes() and use it directly in the for loop
+        // everything seems to work fine, but valgrind tells you otherwise. Problem is, that the DofType is changed
+        // during the function call inside the loop which leads to reads in already freed blocks of memory
+        std::set<Node::eDof> currentDofSet = mStructure->GetDofStatus().GetDofTypes();
```
joergfunger, line 50: Why do we change the doftype during the function call?

src/mechanics/timeIntegration/TimeControl.cpp
```diff
@@ -0,0 +1,105 @@
+#include "TimeControl.h"
+
+
+
+void NuTo::TimeControl::Proceed()
```
joergfunger: The name does not stand for itself. Maybe something like ProceedToNextTimeStep

src/mechanics/timeIntegration/TimeControl.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+void NuTo::TimeControl::UseEquidistantTimestepping()
+{
+    SetTimeStepFunction([](TimeControl& rTimeControl, int iterations, int maxIterations, bool converged)->double
+                        {
+                            return rTimeControl.GetTimeStep();
+                        });
+}
+
+void NuTo::TimeControl::SetMaxTimeStep(double rMaxTimeStep)
```
joergfunger: Shouldn't the max and min time step not be part of this TimeStepFunction? Since they only make sense for adaptive time stepping, whereas for constant steps this is not necessary?

joergfunger (approved):
Thanks for restructuring the time integration. The tmp changes to the explicit time integration schemes should have some priority for the next NuTo session.

src/mechanics/timeIntegration/NewmarkDirect.cpp
```diff
@@ -1,499 +1,182 @@
...
-            PrintInfoStagger();
 
-            // ******************************************************
-            mStructure->Evaluate(inputMap, evaluate_Hessian0_Hessian1);
-            // ******************************************************
+std::pair<int, BlockScalar> NewmarkDirect::FindEquilibrium(StructureOutputBlockVector& structureResidual,
+                                                           const StructureOutputBlockVector& extForce,
+                                                           StructureOutputBlockVector& delta_dof_dt0,
+                                                           std::array<StructureOutputBlockVector, 3>& dof_dt,
+                                                           const BlockSparseMatrix& constraintMatrix)
```
TTitscher: I would remove constraintMatrix from the argument list, zero-cost-getter mStructure->GetConstraintMatrix() (or something similar)

src/mechanics/timeIntegration/NewmarkDirect.cpp
```diff
@@ -584,84 +248,320 @@ void NuTo::NewmarkDirect::PrintInfoStagger() const
...
+    auto inputMap = CreateInputMap();
+    mStructure->Evaluate(inputMap, evalHessians);
+    if (mCheckCoefficientMatrix)
+        mStructure->ElementCheckHessian0(1.e-6, 1.e-8);
+
+    return hessians;
+}
+
+
+std::pair<StructureOutputBlockVector, std::array<StructureOutputBlockMatrix, 3>>
```
TTitscher, line 387: How about `using Hessians = std::array<StructureOutputBlockMatrix, 3>`?

src/mechanics/timeIntegration/NewmarkDirect.cpp
```diff
@@ -584,84 +248,320 @@ void NuTo::NewmarkDirect::PrintInfoStagger() const
...
+NuTo::NewmarkDirect::EvaluateGradientAndHessians()
+{
+    const auto& dofStatus = mStructure->GetDofStatus();
+    StructureOutputBlockVector intForce(dofStatus, true);
+    std::array<StructureOutputBlockMatrix, 3> hessians = {StructureOutputBlockMatrix(dofStatus, true),
```
TTitscher, line 392: We could call `EvaluateHessians()` for the hessians part and `EvaluateGradient()` for the gradient. This means that they are not evaluated simultaneously, causing some overhead. IMO this overhead is small compared to `Solve()`.

src/mechanics/timeIntegration/NewmarkDirect.cpp
```diff
@@ -584,84 +248,320 @@ void NuTo::NewmarkDirect::PrintInfoStagger() const
...
+                                                        StructureOutputBlockVector(dofStatus, true),
+                                                        StructureOutputBlockVector(dofStatus, true)};
+    StructureOutputBlockVector delta_dof_dt0(dofStatus, true);
+    const auto& constraintMatrix = mStructure->GetAssembler().GetConstraintMatrix();
+
+    // Needed for mTimeControl.Proceed() function
+    int timeStepMaxIterations = 0;
+    bool converged = true;
+
+    for (const auto& activeDofs : mStepActiveDofs)
```
TTitscher, line 433: This is quite a big method and may need further splits. Maybe into something like
~~~cpp
auto something = TrialState();
auto somethingElse = EquilibriumIterations();
if (converged)
   HandleConvergedSolution();
else
  ...
~~~

src/mechanics/timeIntegration/TimeControl.cpp
```diff
@@ -0,0 +1,105 @@
...
+                                                      "the RestorePreviosTime() function of the time control before reducing "
+                                                      "the timestep.");
+}
+
+
+
+void NuTo::TimeControl::SetTimeStep(double timeStep)
+{
+    if (timeStep<=0)
+        throw MechanicsException(__PRETTY_FUNCTION__,"Timestep must be larger tham 0!");
```
TTitscher: than :)

src/mechanics/timeIntegration/TimeControl.h
```diff
@@ -0,0 +1,137 @@
+#pragma once
```
TTitscher, line 1: This file (among others) needs a clang format run.

src/mechanics/timeIntegration/postProcessing/PostProcessor.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include "mechanics/timeIntegration/postProcessing/ResultNodeAcceleration.h"
+#include "mechanics/timeIntegration/postProcessing/ResultGroupNodeForce.h"
+#include "mechanics/timeIntegration/postProcessing/ResultElementIpData.h"
+
+using namespace NuTo;
+
+
+void PostProcessor::PostProcess(const StructureOutputBlockVector& outOfBalance)
+{
+    if (mResultDir.length() == 0)
```
TTitscher: if (mResultDir.empty()) is better IMO.

src/mechanics/timeIntegration/postProcessing/PostProcessor.h
```diff
@@ -0,0 +1,93 @@
+#pragma once
+
+#include <string>
+#include <boost/ptr_container/ptr_vector.hpp>
+
+#include "mechanics/structures/StructureBase.h"
+#include "mechanics/timeIntegration/TimeControl.h"
+#include "mechanics/timeIntegration/postProcessing/ResultBase.h"
```
TTitscher: All those includes can be forwarded, right?
(StructureBase, TimeControl, ResultBase)

src/mechanics/timeIntegration/postProcessing/ResultElementIpData.cpp
```diff
@@ -0,0 +1,84 @@
...
+        break;
+    case IpData::eIpStaticDataType::DAMAGE:
+        numComponents = 1;
+        break;
+    case IpData::eIpStaticDataType::SLIP:
+    case IpData::eIpStaticDataType::BOND_STRESS:
+        numComponents = 3;
+        break;
+    default:
+        throw MechanicsException(std::string(__PRETTY_FUNCTION__) + "\t: Ip data type not supported yet.");
```
TTitscher: Since we are touching it right now, this exception is equivalent to
`throw MechanicsException(__PRETTY_FUNCTION__,  "Ip data type not supported yet.");`

test/mechanics/timeIntegration/TimeControl.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+    std::string expectedMsg = "";
+    std::function<bool(MechanicsException const&)> CheckException
+            = [&expectedMsg](MechanicsException const& ex)->bool
+                {
+                    return CheckExceptionMessageCorrect(ex,expectedMsg);
+                };
+
+    expectedMsg = "Current timestep is 0 or negative!";
+//    BOOST_CHECK_EXCEPTION(timeControl.SetTimeStepFunction([](double,double)->double{return 0.0;}),
```
TTitscher: should be uncommented, right?

test/mechanics/timeIntegration/TimeControl.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include "mechanics/timeIntegration/TimeControl.h"
+
+#include <iostream>
+
+using namespace NuTo;
+
+// %%% Helper functions %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+bool CheckExceptionMessageCorrect(MechanicsException const& ex,std::string expectedMsg)
```
TTitscher: I think, you should use the BOOST_CHECK_EQUAL macro instead of this method,
something like `BOOST_CHECK_EQUAL(GetRefinedErrorMsg(ex), expectedMsg);`
Or provide further comments on why you did it with a custom implementation.

TTitscher (changes requested):
Thank you for tackling this topic! 
As you said, this is still work in progress, so you do not have to work on all my remarks. But some of them are easy to fix. 
The main reason why I do not approve the merge is clang format. Please run it on all those files.

src/mechanics/timeIntegration/NewmarkDirect.cpp
Psirus: I'm trying to use type aliases sparingly. If I see `std::array<StructureOutputBlockMatrix, 3>` I know what's going on, assuming I know what a `StructureOutputBlockMatrix` is. With `Hessians` I'll have to go look what this new type I've never heard of represents. But I'll bow to the consensus if we want to go that way.

test/mechanics/timeIntegration/TimeControl.cpp
```diff
@@ -38,55 +38,119 @@ BOOST_AUTO_TEST_CASE(Debug_Fuctionality)
...
 
-        timeControl.SetTimeStep(5.5);
+        timeControl.SetTimeStep(5.0);
+        timeControl.UseEquidistantTimestepping();
 
         timeControl.Proceed();
         timeControl.Proceed();
-        timeControl.Proceed();
-        BOOST_CHECK_EQUAL(timeControl.GetCurrentTime(), 19.5);
+        BOOST_CHECK_EQUAL(timeControl.GetCurrentTime(), 20.0);
```
TTitscher: Please use `BOOST_CHECK_CLOSE(a, b, tolerance)` for floating point comparisons.

src/mechanics/timeIntegration/RungeKuttaBase.h
vhirtham: Because all time related stuff is now done by the TimeControl class. But it is not that simple to replace them in all derived time integration schemes. Plan is to that step by step.

src/mechanics/timeIntegration/NewmarkDirect.cpp
vhirtham: Well rTimeDelta is somehow misleading. It should be renamed to rTimeFinal or something like this. We just didn't change it. So you provide the time where you like to stop when you call solve.

test/mechanics/timeIntegration/TimeControl.cpp
vhirtham: This is just a workaround. Just wait what we decide concerning the exception issue

test/mechanics/timeIntegration/TimeControl.cpp
vhirtham: Dont know about all the boost test features yet. Will change it if  I don't forget to. :p

test/mechanics/timeIntegration/TimeControl.cpp
```diff
@@ -38,55 +38,119 @@ BOOST_AUTO_TEST_CASE(Debug_Fuctionality)
                 };
 
     expectedMsg = "Current timestep is 0 or negative!";
-//    BOOST_CHECK_EXCEPTION(timeControl.SetTimeStepFunction([](double,double)->double{return 0.0;}),
-//                          MechanicsException,CheckException);
-
-
-
+    timeControl.SetTimeStepFunction([](TimeControl&,double,double,bool)->double{return 0.0;});
```
TTitscher: The method is defined to take
`std::function<double(TimeControl&, int, int, bool)> timestepFunction`
Since the test is some kind of reference for the usage of the class, this should be correct here (even though it has no influence in this case...).

TTitscher:
Very nice that you checked all the edge cases! But IMO this is a bit over the top and a lot of duplicate code (the actual exception messages). Just checking that an exception is thrown should be enough. Your debugger will tell you what and why.
What do the others think?

test/mechanics/timeIntegration/TimeControl.cpp
vhirtham: Yeah you are right. Problem is, that was the old function interface and I changed it due to a hint of pHuschke. I missed that one. Strange though, that the compiler accepts it.

src/mechanics/timeIntegration/TimeControl.cpp
```diff
@@ -0,0 +1,109 @@
...
+    mTimeStep = timeStep;
+}
+
+void NuTo::TimeControl::SetTimeFinal(double timeFinal)
+{
+    if (timeFinal <= mCurrentTime)
+    {
+        throw MechanicsException(__PRETTY_FUNCTION__,"Final time must be larger than current time!");
+    }
+    mTimeFinal          = timeFinal;
```
TTitscher: Way too many spaces here... Please setup clang-format and run it! Merci much! (Then, IMO, this can be merged)

vhirtham:
Well concerning the exception message checking. That's related to the exception issue. It will be changed when we use specific exception types to distinguish between them.  I want to test specific exceptions because I want to make sure, that the "right" exception is triggered in the "right" circumstances. If you change something and accidentally mess up the exception hierarchy the "right" one might never be trigger because another one is thrown before. 
This one exception might give you a totally wrong hint whats wrong. As a stupid example: You get a "No Convergence" exception because you divided somewhere on the way by 0 and the check for that ends up to be somewhere after the convergence check. As I said, stupid example, but I already got those cases, where the meaningful exception was shadowed by some other, which took me a lot of time to figure out.  
So in the end: I don't think that everybody should test every edge case, but we should probably not complain about to much being tested. Who cares about some code repetition in a rather small test file, that nobody touches anymore, except when it fails?

src/mechanics/timeIntegration/TimeIntegrationBase.h
vhirtham: Yep. Fixed it

src/mechanics/timeIntegration/NewmarkDirect.cpp
vhirtham: changed parameter name to timeFinal and cahnged the doxy-doc in the TimeIntegrationBase and NewmarkDirect headers

TTitscher (approved):

PR #96, opened 2017-08-31
PDE: Linear elastic law

TTitscher:
To start somewhere, I implemented a simple linear elastic law. Since I wanted to use proper EngineeringStrain and EngineeringStress classes, I added the suffix _PDE_. Since the Voigt strain contains a division by 2 in the lower diagonal, it is important to differentiate between those two. Thus, a simple
~~~c++
using Stress = Eigen::Matrix<...>;
using Strain = Eigen::Matrix<...>;
~~~
is not enough and I chose define two separate classes that inherit from `Eigen`. 

Comments/wishes/changes?

Please feel free to provide commits on your own :)

### DIFF
--- a/src/mechanics/constitutive/EngineeringStrainInvariants.h
+++ b/src/mechanics/constitutive/EngineeringStrainInvariants.h
@@ -29,14 +29,14 @@ Tensor ToTensor(EngineeringStrainPDE<3>& v)
 
 //! @brief returns I1 - the first strain invariant of the characteristic equation
 //! \f[ \lambda^3 - I_1 \lambda^2 + I_2 \lambda - I_3  \f]
-double InvariantI1(const EngineeringStrainPDE<3>& v)
+double I1(const EngineeringStrainPDE<3>& v)
 {
     return v.segment<3>(0).sum();
 }
 
 //! @brief returns I2 - the first strain invariant of the characteristic equation
 //! \f[ \lambda^3 - I_1 \lambda^2 + I_2 \lambda - I_3  \f]
-double InvariantI2(const EngineeringStrainPDE<3>& v)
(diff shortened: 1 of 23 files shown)


src/mechanics/constitutive/EngineeringStrainInvariants.h
```diff
@@ -0,0 +1,85 @@
...
+EngineeringStrainPDE<3> Deviatoric(EngineeringStrainPDE<3> v)
+{
+    double I1_3 = InvariantI1(v) / 3.;
+    v[0] -= I1_3;
+    v[1] -= I1_3;
+    v[2] -= I1_3;
+    return v;
+}
+
+} /* TensorCalc */
```
Psirus, line 84: Do you mean `/* EngineeringStrainInvariants */`? And if the namespace is `EngineeringStrainInvariants`, wouldn't be `EngineeringStrainInvariants::I3` sufficient, instead of `EngineeringStrainInvariants::InvariantI3`?

src/mechanics/constitutive/EngineeringStrainPDE.h
```diff
@@ -0,0 +1,69 @@
...
+ *
+ */
+template <int TDim>
+class EngineeringStrainPDE : public Eigen::Matrix<double, Voigt::Dim(TDim), 1>
+{
+    using Parent = Eigen::Matrix<double, Voigt::Dim(TDim), 1>;
+public:
+    EngineeringStrainPDE() = default;
+
+    // This constructor allows you to construct EngineeringStrainPDE from Eigen expressions
```
Psirus: Why are these not Doxygen comments?

Psirus (approved):

Issue #43, opened 2017-06-09
Evaluate in the PDE concept - data transfer from structure level to integrand and vice versa


joergfunger:
In the PDE concept, the general layout is as follows. The structure has a list of (quadrature) cells. The only function of these cells is the integration, thus attached to each cell is an integration type. Furthermore, in each cell, an interpolation for a specific dof_type is defined, including the coordinate interpolation. For each integration point, a separate integrand function is defined that specifies what is to be integrated (e.g. for a standard mechanics problem this is B^T \sigma  or K = B^T C B). This could however also include the time integration, so instead of assembling only K, directly assemble K+a/tC+b/(t^2)M with a and b coefficients from the time integration scheme such as Newmark).
This requires to evaluate a variety of different objects (mass matrix, damping matrix, stiffness, residual, full hessian) with a variety of options and global inputs (residual w/o update of history data, implicit/explicit integration of history variables (IMPLEX, cycle jump with number of cycles), time step t. In order to avoid writing a separate global function (on the structure level) for each specific output (requiring different inputs)

`Structure.EvaluateHessian0(), Structure.EvaluateHessian0(double t), ..`

the following options are possible:

### Solution 1: Option class
 I would propose to write a single evaluate function that has an optionBase class as an argument. Note that this is very different from the current input/output, since it only includes global options (so no reallocation on structural, element and constitutive level is required, it's just passed by reference). This class also defines what should be stored in the corresponding single output objects.

```
Hessian0 = Structure.EvaluateSparseMatrix(myOptionsHessian0);
Gradient = Structure.EvaluateFullMatrix(myOptionsGradient);
Hessian = Structure.EvaluateSparseMatrix(myOptionsFullHessianIncludingTimeIntegration);
```
Here, we should decide, if we want to pass the outputObjects by reference (which has the advantage of being able to predefine the sparsity pattern of a sparse matrix), or by defining the result directly as an output. Both approaches would be possible. I would only write two global functions, one for full matrices (including vectors) and another one for sparse matrices. This could also be templated.

Output type would be an empty base class. For specific integrands, we would define certain options. Note that this option class can be redefined by any user in its main file and is then directly passed to the integrand. So the "user" can do very weird thinks without having to change any of the interfaces in NuTo, just change the optionClass and the integrand in his main file.

```
class OptionsTimeIntegration : OptionsBase 
{
private: 
  ResultEnum mResult; //e.g. Hessian0, Hessian1, Residual, ...
  double mDeltaT;
[...]

PR #97, opened 2017-08-31
Pde hessian0

TTitscher:
This pull request basically adds a Hessian0() to cell and integrand. As lively debated in #43, this method can be later replaced by Integrate(Operator) or something. It returns a `NuTo::DofMatrix` and we can start to write an assembler and eventually solve the first global system.

The other commits are pure renaming, replacing the `rStuff` pattern in the arguments with `stuff`.

### DIFF
--- a/src/mechanics/cell/Cell.h
+++ b/src/mechanics/cell/Cell.h
@@ -41,6 +41,24 @@ public:
         return gradient;
     }
 
+    //! @brief builds the hessian0 matrix
+    DofMatrix<double> Hessian0() override
+    {
+        DofMatrix<double> hessian0;
+        CellData cellData(mElements);
+        for (int iIP = 0; iIP < mIntegrationType.GetNumIntegrationPoints(); ++iIP)
+        {
+            auto ipCoords = mIntegrationType.GetLocalIntegrationPointCoordinates(iIP);
+            auto ipWeight = mIntegrationType.GetIntegrationPointWeight(iIP);
+            Jacobian<TDim> jacobian(mCoordinateElement.ExtractNodeValues(),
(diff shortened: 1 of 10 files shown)


src/mechanics/cell/Cell.h
```diff
@@ -41,6 +41,24 @@ class Cell : public CellInterface
         return gradient;
     }
 
+    //! @brief builds the hessian0 matrix
+    DofMatrix<double> Hessian0() override
+    {
+        DofMatrix<double> hessian0;
```
Psirus, line 47: Will this not simply call the default constructor of Eigen, possibly resulting in a nonzero initial value?

Psirus (approved):

src/mechanics/cell/Cell.h
TTitscher: This is kind of a special behavior of the `DofContainer` things. They allow - as I called it - [uninitialized addition](https://github.com/nutofem/nuto/blob/master/test/mechanics/nodes/DofMatrix.cpp#L38). So this line will create a DofMatrix<double> containing zero sized vectors that is resized when needed.

Psirus: Okay. Not the most intuitive behaviour though.

TTitscher: If you can come up with a better solution, please let me know. The problem is that the cell does not know the size of this container, or how many dof types it contains. We could do sth like
~~~c++
DofMatrix<double> hessian0 = mIntegrands[0].Gradient(...);
for (i = 1, ...) 
   hessian0 += mIntegrands[i].Gradient(....);
~~~
this seemed kind of silly though.

Issue #63, opened 2017-06-22
Group refactor


Psirus:
The groups are another one of these areas that could simply be nicer.

- they don't need to be stored at the structure
- they don't need to have member IDs
- there is no need for a base class
- the operations `Difference`, `Intersection` and so on should be free functions

Oh, and things like `GetType` or `GetTypeString` should be a punishable offense anyway.

In addition, all the group related `Structure` methods should either go away to where they came from, or be converted to free functions too.

What has been stopping me (apart from lack of time) is that I would like to iterate over the group, i.e.
~~~cpp
for(auto& member : group)
    member.DoMemberishThings();
~~~

What I've been able to come up with so far can be seen in commit f433b77. There, the group saves raw pointers to its constituents and the `begin()` and `end()` iterators have been implemented as `boost::indirect_iterator`s to get the value semantics.
Pro:
- `auto`
- value semantics

Con:
- raw ptr
- `begin` and `end` behave differently from normal

Another option would be to use `std::reference_wrapper` instead of raw pointers. This would get you the value semantics and normally behaving `begin` and `end`, but no `auto`.

Maybe you have an idea that circumvents all the cons? Shark below :shark:

TTitscher:
IMO a raw ptr is nothing bad. In contrast to a raw _owing_ ptr that requires `new` and `delete`. So if the ptr is hidden as an implementation detail and accessed via reference, I see no con there.

Psirus:
Well, sort of. What happens if the object pointed to goes out of scope earlier than the group? Unlikely in our use cases, but still.

joergfunger:
Is it that unlikely that the object goes out of scope? IMO, we do sometimes a reallocation of certain components (i.e. currently nodes when the dofs are changing, static data when the integration type (and thus the number of static data changes), .... If the group is just stored outside (of the structure) either with a raw ptr or a shared ptr, this might cause issues. (either the object is out of scope and the ptr is not valid, or the object is not deleted but now the group object and the structure object are different). This would require some exchange_ptr methods.
Similar, the restore procedure for the structure (before starting e.g. the visualization/restart), this changes the pointers as well, so all groups would have to be recreated.

TTitscher:
[...]

PR #99, opened 2017-09-01
Pde group

TTitscher:
Add group as proposed in #63. And this group is handy for the next steps in the PDE implementation. So I just added it. 

Anyways, in #63 we discussed the behavior concerning duplicates. 

6748a06 adds the group as it is in the gist, based on std::vector, allowing duplicates.
Can you think of cases where multiple same entries in a group make sense? IMO it is reasonable to ignore duplicates. 

Thus, the current version uses a std::set. This automatically ignores duplicates but loses the ordering. As @joergfunger mentioned, this could cause problems in tests that rely on an order. One example, he thought of (and I hope I understood correctly), are constraint equations that are defined as follows. First, select nodes and put them in our group. Second, pick _some_ node (say `Group::begin()`) from that group and define it as the active dof. Third, define constraint equations, connecting all remaining group entries with the active dof. Last, test some global result vector. So depending on `Group::begin()` - that may change from one run to another (Group [= std::set] is now sorted by pointer addresses) - the active/dependent dofs and thus the result vector changes. This test could easily be rewritten to avoid this problem by manually picking the active dof by e.g. its coordinate. By maybe I misunderstood the issue or you can think of other cases where order matters. 

So what do you think?

### DIFF
--- a/src/base/Group.h
+++ b/src/base/Group.h
@@ -28,6 +28,19 @@ public:
     typedef std::vector<T*> parent;
     typedef boost::indirect_iterator<typename parent::iterator> GroupIterator;
 
+    Group() = default;
+
+    Group(T& element)
+    {
+        Add(element);
+    }
+    
+    Group(std::vector<std::reference_wrapper<T>> elements)
+    {
+        for (auto element : elements)
+            Add(element);
+    }
+
     void Add(T& element)
     {
         auto it = std::lower_bound(pbegin(), pend(), &element, TCompare());
@@ -41,6 +54,11 @@ public:
         if (result == pcend())
             return false;
         return *result == &element;
+    }
+
+    bool Empty() const
+    {
+        return size() == 0;
     }
 
     GroupIterator begin()
--- a/src/base/Group.h
+++ b/src/base/Group.h
@@ -20,6 +20,18 @@ struct IdSmaller
 
 class Utils;
 
+//! @brief Ordered container class for elements, nodes and the like. No duplicates.
+//!
+//! The group saves pointers to its members but can be iterated through
+//! by value. By default the members are sorted by id (this must be unique)
+//! that is accessed by calling Id(), e.g as provided by NuTo::UniqueId. Other
+//! methods operating on the group also use this comparator.
+//! This behaviour can be changed by providing a different comparator but the basis
+//! of the comparison has to be a unique attribute of the group member.
+//! See test file (CutomCompare).
+//!
+//! @tparam T type of group members
+//! @tparam TCompare functor representing the comparison operation used for sorting
 template <typename T, typename TCompare = IdSmaller<T>>
 class Group : private std::vector<T*>
 {
@@ -29,19 +41,23 @@ public:
     typedef std::vector<T*> parent;
     typedef boost::indirect_iterator<typename parent::iterator> GroupIterator;
 
+    //! @brief Empty group
     Group() = default;
 
+    //! @brief Group containing one element
     Group(T& element)
     {
         Add(element);
     }
 
+    //! @brief Group containing multiple elements
     Group(std::vector<std::reference_wrapper<T>> elements)
     {
         for (auto element : elements)
             Add(element);
     }
 
+    //! @brief Add element to group
     void Add(T& element)
     {
         auto it = std::lower_bound(pcbegin(), pcend(), &element, TCompare());
@@ -49,6 +65,11 @@ public:
             parent::insert(it, &element);
     }
 
+    //! @brief True if element is contained in group
+    //!
+    //! Comment:
+    //! Search for element is done using Comparator,
+    //! Equality is checked by pointer equivalence.
     bool Contains(const T& element) const
     {
         auto result = std::lower_bound(pcbegin(), pcend(), &element, TCompare());
@@ -57,21 +78,25 @@ public:
         return *result == &element;
     }
 
+    //! @brief True if group is empty
     bool Empty() const
     {
         return Size() == 0;
     }
 
+    //! @brief Number of constituents
     auto Size() const
     {
         return size();
     }
 
+    //! @brief Iterate over group elements (by value not pointer)
     GroupIterator begin()
     {
         return parent::begin();
     }
 
+    //! @brief Iterate over group elements (by value not pointer)
     GroupIterator end()
     {
         return parent::end();
@@ -96,6 +121,7 @@ private:
     using parent::size;
 };
 
+//! @brief Unite two groups
 template <typename T, typename TCompare>
 static Group<T, TCompare> Unite(const Group<T, TCompare>& one, const Group<T, TCompare>& two)
 {
@@ -105,6 +131,7 @@ static Group<T, TCompare> Unite(const Group<T, TCompare>& one, const Group<T, TC
     return newGroup;
 }
 
+//! @brief Returns group with elements of group one that are not in group two
 template <typename T, typename TCompare>
 static Group<T, TCompare> Difference(const Group<T, TCompare>& one, const Group<T, TCompare>& two)
 {
@@ -114,6 +141,7 @@ static Group<T, TCompare> Difference(const Group<T, TCompare>& one, const Group<
     return newGroup;
 }
 
+//! @brief Returns group with elements that are in both groups
 template <typename T, typename TCompare>
 static Group<T, TCompare> Intersection(const Group<T, TCompare>& one, const Group<T, TCompare>& two)
 {
@@ -123,6 +151,7 @@ static Group<T, TCompare> Intersection(const Group<T, TCompare>& one, const Grou
     return newGroup;
 }
 
+//! @brief Returns group with elements that are only in one group not in both
 template <typename T, typename TCompare>
 static Group<T, TCompare> SymmetricDifference(const Group<T, TCompare>& one, const Group<T, TCompare>& two)
 {
--- a/src/base/Group.h
+++ b/src/base/Group.h
@@ -12,7 +12,7 @@ namespace Groups
 template <typename T>
 struct IdCompare
 {
-    bool operator()(T* a, T* b) const
+    bool operator()(const T* a, const T* b) const
     {
         return a->Id() < b->Id();
     }
--- a/src/base/Group.h
+++ b/src/base/Group.h
@@ -62,13 +62,12 @@ public:
         return parent::cend();
     }
 
-    bool Contains(const T& element)
+    bool Contains(const T& element) const
     {
-        auto result = std::find(parent::begin(), parent::end(), &element);
-        if (result != parent::end())
-            return true;
-        else
+        auto result = std::lower_bound(pcbegin(), pcend(), &element, TCompare());
+        if (result == pcend())
             return false;
+        return *result == &element;
     }
 
     using parent::size;
--- a/src/base/Group.h
+++ b/src/base/Group.h
@@ -76,42 +76,43 @@ public:
 };
 
 
-template <typename T>
-Group<T> Unite(const Group<T>& one, const Group<T>& two)
+template <typename T, typename TCompare>
+Group<T, TCompare> Unite(const Group<T, TCompare>& one, const Group<T, TCompare>& two)
 {
-    Group<T> newGroup;
-    std::insert_iterator<Group<T>> newGroupInsertIterator(newGroup, newGroup.pbegin());
-    std::set_union(one.pcbegin(), one.pcend(), two.pcbegin(), two.pcend(), newGroupInsertIterator);
+    Group<T, TCompare> newGroup;
+    std::insert_iterator<Group<T, TCompare>> newGroupInsertIterator(newGroup, newGroup.pbegin());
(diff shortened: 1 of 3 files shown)


Psirus (approved):

joergfunger:
The std::set is certainly the appropriate container for a group. However, my concern here is that the order of iterating through a container depends on the runtime (since its ordered according to the ptr). This might not only be an issue for constraints as discussed above (where the dependent/independent dof is different for different orderings), but also in case of implementation errors (that give different memory violations depending on the ordering), and thus might lead to unreproducible errors (sometimes it works, sometimes it does not). Furthermore, we should at some point still have some id for elements/cells or nodes (in particular,  when we will have nodes that do no longer have a coordinate), since identifying (while debugging either the source code or my model) the node from a (varying) ptr is not really practical.

joergfunger (approved):

TTitscher:
@joergfunger I tried to address the Id issue with #100. Cells, Integrands, Nodes, ..., can be derived from `NuTo::UniqueId` to get a unique Id that does not change when rerun.

src/base/Group.h
```diff
@@ -0,0 +1,110 @@
...
+#include <boost/iterator/indirect_iterator.hpp>
+
+namespace NuTo
+{
+namespace Groups
+{
+
+template <typename T>
+class Group : private std::set<T*>
+{
```
phuschke: #### General
Why don't derive from `std::set<std::weak_pt<T>>`?

From [stackoverflow](https://stackoverflow.com/questions/12030650/when-is-stdweak-ptr-useful):
_This is exactly what a weak pointer does -- it allows you to locate an object if it's still around, but doesn't keep it around if nothing else needs it._

It is implemented in base, i.e. users will use it for types other than Nodes and Elements. This could happen:
~~~cpp

Group<int> g;
{
int a = 1;
g.AddMember(a);
} // a leaves scope
// undefined behavior ...
~~~
 
Users don't know if it allows duplicates or if it is sorted or not. The sorting uses the memory address which is completely counterintuitive and should definitely be mentioned in the **documentation**.

src/base/Group.h
```diff
@@ -0,0 +1,110 @@
...
+        return parent::begin();
+    }
+
+    GroupIterator end()
+    {
+        return parent::end();
+    }
+
+    typename parent::iterator pbegin()
+    {
```
phuschke, line 115: 1. Very fancy use of the `indirect_iterator`. 
**But** we use 6 functions just to hide the underlying implementation which can also be misleading?

2. Apparently `cbegin` and `cend` are not necessary to use
~~~cpp
for (const auto& ele : group)
...
~~~
but I don't understand why...

3. I think renaming `pbegin` to `parentBegin` or sth more expressive would be better. I first thought `pbegin` was part of STL :smile:

src/base/Group.h
```diff
@@ -0,0 +1,110 @@
...
+        if (result != parent::end())
+            return true;
+        else
+            return false;
+    }
+
+    using parent::size;
+    using typename parent::iterator;
+    using typename parent::value_type;
+    using parent::insert;
```
phuschke: Why do we want to allow users to use `insert` and `AddMember`?

phuschke (commented):
I like the idea of hiding the underlying data structure from the user but it feels like we are pretending to store values/references/owning ptrs instead of raw pointers which can be a pitfall.

src/base/Group.h
TTitscher: > [weak_ptr](http://en.cppreference.com/w/cpp/memory/weak_ptr) is a smart pointer that holds a non-owning ("weak") reference to an object that is managed by std::shared_ptr. It must be converted to std::shared_ptr in order to access the referenced object.

So `weak_ptr` would require that each object is allocated as a shared_ptr. This is not the way to go.

What we are actually looking for is 
`std::InsertContainerHere<`[`std::reference_wrapper`](http://en.cppreference.com/w/cpp/utility/functional/reference_wrapper)`<T>>`. This would work without the indirect iterator, but the syntax would include `get()`:
~~~c++
for (auto entity : group)
    entity.get()->MemberFunction();
~~~

---
The whole out-of-scope topic did not occur because our _god class_ `structure` owned every object. IMO we want to eliminate that and have to take care of the scope. Technically, you can mess up whenever you have a reference or pointer member.
~~~c++
NuTo::NewmarkDirect CreateNewmark(...)
{
    NuTo::Structure s;
    s.Import(...)
    // ...
    return NuTo::NewmarkDirect(&s); 
} // Boom. s out of scope.
~~~

---
> Users don't know if it allows duplicates or if it is sorted or not.

I got your hint, I will add documentation. But what should the requirements be like? IMO: 
- no duplicates
- not sorted. Meaning that you should not rely on any sorting. Internally, some sorting might occur for performance reasons.

src/base/Group.h
TTitscher: You are right, we do not want users to use `insert`. However, it is required by the [`insert_iterator`](http://en.cppreference.com/w/cpp/iterator/insert_iterator). 
> The container's insert() member function is called whenever the iterator (whether dereferenced or not) is assigned to.

If we want to keep `Unite(), Difference(),...` as free functions, we have to keep that. As well as all this `pbegin(), pcend()` stuff. And the more I think about it, the more it annoys me as well. Any suggestions? :)

pmueller2:
To get some reproducible ordering one could save an id alongside with the pointer that is just incremented from zero each time a member is added. And add a sort routine that is based on these "ids" and can be called by functions that rely on such ordering, e.g. visualization

joergfunger:
I also think that the only think we actually require is a standard set that is ordered (due to debugging purposes to get the some results for multiple runs). This ordering could, e.g. be a global id counter. The only (minor) problem with this approach is the fact that we also wanted to store the node id (that is e.g. assigned from the mesh file to find a specific node), which would mean having two ids at the node - this might be counter-intuitive.

vhirtham:
Some stuff concerning the current design process:

What we want:
Class that stores its members with the following attributes
- unique elements
- ordered (reproduceable -> Pointer problem)

Design proposals:

---------------------------------------------------------------------------------------------------------------------------------

**std::vector**

- ordering by insertion order (could be sorted afterwards any way you like)
- not unique itself

How to solve unique Problem?

1. AddMember function checks existence with std::find before insertion
2. Make unique when the first Getter/Iterator/Whatever is called (like lazy evaluation)

Comment: I think I would prefer the first solution, because it less prone to errors. You have just one function where you need to make your storage unique. The linear search is quite fast and you fill the group only once.
 
---------------------------------------------------------------------------------------------------------------------------------

**std::set**

- ordering by default < operator. Alternative sort function can be provided as template parameter
-  unique by design

How to solve Pointer Problem?

1. Use IDs to keep a persisting order
2. Got nothing...

---------------------------------------------------------------------------------------------------------------------------------

Comment:
We (Philip M. and me) prefer the std::vector. Whats your  opinion? Other alternatives?

TTitscher:
Can you provide some argumentation about the complexity? E.g. [`std::find`](http://en.cppreference.com/w/cpp/algorithm/find#Complexity) is __O(n)__ for __n__ elements. So adding n elements to the proposed vector container and calling find every time is __O(n*n)__ and _very_ bad. As a crazy alternative, we could have both a set and a vector. Instead of std::find on a vector, we insert the element to the set. This is __O(log n)__. If insertion fails (already exists), we do not push it into the vector. 

There are other problems with vector. Currently, we use std algorithms like [`set_intersection`](http://en.cppreference.com/w/cpp/algorithm/set_intersection). They require a sorted (via operator<) container. Do we write those manually?

pmueller2:
I think a complexity of **O(n*n)** is only bad if it is slow. And it is only slow (whatever this means exactly) if **n** is really large (whatever this means exactly - we tested it with filling a group of integers of size 1e9 and it took a second or so on my machine).
It is imho easier to solve the problems a vector has with set operations and uniqueness, than this "sorting by pointer addresses that change on each run" problem a set has.

TTitscher:
So you claim that 1e9 calls to std::find on a vector is only one second? Can you provide your code for that? I quickly came up with this and it takes ages.
~~~cpp
#include <vector>
#include <algorithm>

struct UniqueVector
{
    void Add(int d)
    {
        if (std::find(mData.begin(), mData.end(), d) != mData.end())
            return; // already in there.
        mData.push_back(d);
    }
    std::vector<int> mData;
};

int main()
{
    UniqueVector v;
    for (int i = 0; i < 1e6; ++i)
        v.Add(i);
}
~~~

vhirtham:
Okay, we messed up the comparison in our example, so the vector was never filled. You are right, now it takes forever, which basically leaves us with the std::set or equivalent containers. Now there are two options to solve the pointer problem:

- sort by IDs (container internal ID or ID of the thing you put inside)
- overload the comparison operator for elements and nodes,... well... I think that will be a big pain in multiple ways


So what shall it be? Node and element IDs or internal container IDs?

pmueller2:
We could provide a Comparator as second template parameter to `std::set` that compares Ids of the member objects. This would mean we have Ids again. The group just uses them for sorting and other uses and misuses can be discussed elsewhere.

TTitscher:
Sorting via the id of the entity seems reasonable (no additional internal id!). This could at least be the default compare operator. If you like, you can provide a custom one as template parameter. 
~~~cpp
template <typename T, typename TCompare = IdCompare>
class Group {...};
~~~

As for the internals, I don't care. std::set or std::vector with sorted insert ([taken from here](http://lafstern.org/matt/col1.pdf)) like
~~~cpp
SortedVector<T>::insert(T&& t)
{
  auto it = std::lower_bound(mData.begin(), mData.end(), t);
  if (it == mData.end() || t < *it)
    mData.insert(it, t);
}
~~~
vector performance should be better (see pdf linked above), but I guess it won't matter much.

vhirtham:
Another shot:

Since we are not the only ones with this kind of problem I asked aunt Google. She gave me this thread:

https://stackoverflow.com/questions/28767234/what-container-to-store-unique-values

Interesting is the first answer. Since we want uniqueness and ordered data, we can take advantage of that to use a vector as base class. Inserting new data into an ordered vector is not too expensive. Worse than a set, but it does not take ages. If you know that your data is ordered, you can check uniqueness now with a binary search instead of a linear search, which has a log(O) complexity (the same as std::set).

Because this is super smart, boost already did it with their flat_set and flat_map containers:
http://www.boost.org/doc/libs/1_50_0/doc/html/boost/container/flat_set.html

So we get the cache and speed advantages of a vector while using the group for the cost of a little bit higher "construction time". I ll build a test to check the speed. 

However, the ID problem is still the same as for the std::set  

**EDIT:** 
Dang it! Thomas was faster xD

TTitscher:
I cannot approve/decline this request, since I opened it, but I think it looks good now. Upon multiple requests, we introduced a deterministic ordering based on a Id() method (or any other user provided sorting comparison function). 

Oh, and I changed std::set to a sorted std::vector. Just because of personal antipathy against the set (Actually, because of the performance/memory overhead, but that would count as pre-optimization...).  We can revert e071035 for an undo.

Another open point, mentioned by @phuschke, are the unnecessary public methods. I moved them to a `class Utils`. This allows the four parent iterator methods to be private. `parent::insert` still has to be public. Any further thoughts on that?

src/base/Group.h
```diff
@@ -10,10 +10,19 @@ namespace Groups
 {
 
 template <typename T>
-class Group : private std::set<T*>
+struct IdCompare
+{
```
phuschke: Can we think of a more expressive name? 

I would think that if `IdCompare(a,b)` yields `true`, it implies that `a==b` which is not correct.

src/base/Group.h
```diff
@@ -31,6 +71,17 @@ class Group : private std::set<T*>
         return parent::end();
     }
 
+    using parent::size;
+    using parent::insert;
```
phuschke: Is `insert` still necessary? It would mess up all the sorting if someone were to use `insert` instead of `Add`, right?

TTitscher:
@phuschke: I fixed the insert() thing, you come up with a new name for IdCompare. Deal? :hammer:

phuschke:
IdSmaller :heavy_heart_exclamation:

phuschke (approved):

TTitscher:
@pmueller2 @vhirtham are you pleased with the current version as well?

TTitscher:
Oh. And the documentation is missing. Who is assigned to the issue again? :ghost:

pmueller2:
Looks good. Only thing left now is to use it. Any thoughts on exposing this to python (Yes/No/How)?

pmueller2 (approved):

Psirus:
Of course this should also be exposed to Python, including the utility functions. As an entrypoint, add the headers to `ModuleNuToBase.i` and see what happens.

TTitscher:
There is no python interface for the other classes. So I'd like to postpone that. However, I'd like to see some documentation :)

Psirus:
I didn't say this needs to happen in this PR. But sooner or later you'll need this if you want the Python examples to work with it.

PR #100, opened 2017-09-01
PDE_ids

TTitscher:
Really quick implementation of a trivial static counted id class. 

Obtain a unique id for your class `Foo` by deriving from `NuTo::UniqueId<Foo>´. The ids start at 0 and are incremented at each object instantiation. The template argument provides a new numbering (starting at 0) for each template argument.

Without template:
~~~c++
struct Foo : UniqueId {};
struct Bar : UniqueId {};

Foo f0; // 0;
Foo f1; // 1;
Bar b0; // 2;
Foo f2; // 3;
bar b1; // 4;
~~~

With template:
~~~c++
struct Foo : UniqueId<Foo> {};
struct Bar : UniqueId<Bar> {};

Foo f0; // 0;
Foo f1; // 1;
Bar b0; // 0;
Foo f2; // 2;
bar b1; // 1;
~~~

### DIFF
--- a/src/base/UniqueId.h
+++ b/src/base/UniqueId.h
@@ -1,0 +1,31 @@
+#pragma once
+
+namespace NuTo
+{
+//! @brief This class provides a unique id (beginning at 0 and incrementing for each object).
+//! @tparam T Without the template, the id would be unique for all classes. Say, 1000 nodes are created first and get
+//! the ids 0...999. Element ids would then start at 1000. With this template, T=Node or T=Element, each T starts at 0
+//! again.
+template <typename T>
+class UniqueId
+{
+public:
+    UniqueId()
+        : mId(mIdCounter++)
+    {
+    }
+
+    int Id() const
+    {
+        return mId;
+    }
+
+private:
+    int mId;
+    static int mIdCounter;
+};
+
+template <typename T>
+int UniqueId<T>::mIdCounter = 0;
+
+} /* NuTo */
--- a/test/base/CMakeLists.txt
+++ b/test/base/CMakeLists.txt
@@ -1,1 +1,3 @@
+add_unit_test(UniqueId)
 add_subdirectory(serializeStream)
+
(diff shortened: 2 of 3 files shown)


src/base/UniqueId.h
```diff
@@ -0,0 +1,31 @@
...
+class UniqueId
+{
+public:
+    UniqueId()
+        : mId(mIdCounter++)
+    {
+    }
+
+    int Id() const
+    {
```
phuschke, line 19: Not sure if this is what you intended.

It will return the `Id` of the **next** object that will eventually be created. Not the last one.

Mb this is what you wanted?
~~~cpp
private:
static int mIdCounter = -1;
public:
UniqueId() : mId(++mIdCounter){}
~~~

TTitscher: Have a look at the test. First call to Id() returns 0.

phuschke: You are right. For some reason I thought it would return `mIdCounter`

joergfunger (commented):
In principal I agree. The only remark from my side is related to the storage of the id's only at the entity itself (node number, cell number). I some cases, it might be necessary to access entities directly (e.g. when communication between different processors are required or constraints are applied to specific elements, ..). In this case, it would be advantageous to keep the map with the ids to access the elements directly. This would then mean to store the id twice (which is not a problem).

TTitscher:
> I some cases, it might be necessary to access entities directly. 

Maybe I misunderstood, but I think this is exactly our goal: accessing entities directly (by reference/pointer) without an id. For the constraints, this works well already. And at the moment, I can hardly anticipate how this whole MPI thing will work. Maybe we should have a look at some other FEM frameworks how they deal with it. I could think of an _MPI domain boundary substructure_ that could use and id-entity-map for more efficient communication. But IMO this is another issue.

joergfunger:
Maybe the term "it might be necessary to access entities directly" was not used correctly. I meant that we might have to define relationships between domains (say e.g. a master node in one domain is a slave node with the same dofs in the neighboring domain). These two domains are created independently from each other on two different MPI processes. At the point, where we want to exchange information, we have no idea what the ptr in the other domain is, only some node id from the mesh file. There is certainly an option to communicate the ptr for all entities as soon as they have been created, though this is probably more difficult to debug.

Psirus (approved):

PR #103, opened 2017-09-08
Pde iga

potto1:
**IGA support and interpolation in the PDE concept**
- Added class implementing NURBS
- Added classes CellInterpolationFEM and CellInterpolationIGA
- Removed ElementSimple.h


src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,478 @@
...
+class NURBS
+{
+public:
+
+    enum mParametrization {chord, centripetal};
+
+    /** Constructors **/
+
+    //! @brief ... default constructor
+    NURBS(){}
```
Psirus: What do you need this constructor for?

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,478 @@
...
+
+    //! @brief ... constructor
+    //! @param rDegree ... degree of the polynomial
+    //! @param rKnots ... knot vector
+    //! @param rControlPoints ... control points
+    NURBS(const std::array<std::vector<double>, TDimParameter> &rKnots,
```
Psirus: There has been an agreement that `const` references should no longer feature the prefix `r`.

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,478 @@
...
+    {
+        assert(dir <= TDimParameter && dir >= 0);
+        return mDegree[dir]+1;
+    }
+
+    int GetNumControlPointsElement() const
+    {
+        int numCPs = 1;
+        for(int i = 0; i < TDimParameter; i++)
+            numCPs *= mDegree[i]+1;
```
Psirus: Consider a range-based for loop.
~~~cpp
for(int degree : mDegree)
    numCPs *= degree + 1;
~~~

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,478 @@
...
+            knotCoordinates(1) = mKnots[i][knotIDs[i]+1];
+            knots[i] = knotCoordinates;
+        }
+        return knots;
+    }
+
+
+    Eigen::VectorXd GetControlPointsElement(std::array<int, TDimParameter> knotID) const
+    {
+        throw new std::string("Not implemented for an arbitrary dimension!");
```
Psirus: Throw an exception rather than a string. Even better: consider using `std::enable_if` to define this function only for specific cases.

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,478 @@
...
+    {
+        const std::array<int, TDimParameter> spanIdx = FindSpan(parameter);
+        Eigen::MatrixXd shapeFunctions = BasisFunctionsAndDerivativesRational(0, parameter);
+
+        Eigen::VectorXd coordinates(GetDimension());
+        coordinates.setZero(GetDimension());
+
+        Eigen::VectorXd cpCoords = GetControlPointsElement(spanIdx);
+
+        std::cout << "shapeFunctions: " << shapeFunctions << std::endl;
```
Psirus: I don't think you want to output "shapeFunctions" every time you call `Evaluate`.

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,478 @@
...
+          const std::vector<std::vector<double>>               &rWeights,
+          const std::array<int,TDimParameter>                  &rDegree)
+        : mKnots(rKnots), mControlPoints(rControlPoints), mWeights(rWeights), mDegree(rDegree)
+    {}
+
+    /** Getter **/
+
+    //! @brief ... get the dimension of the curve
+    //! @return ... dimension of the curve
+    int GetDimension() const
```
Psirus: Maybe you could add a comment about what dimension `TDimParameter` is, and the dimension returned here.

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,478 @@
...
+            iterations++;
+            if(iterations > size)
+                throw new std::string("The maximum number of iterations for finding the span exceeded.");
+        }
+        return mid;
+    }
+
+    static Eigen::MatrixXd BasisFunctionsAndDerivatives(int der, double parameter, int spanIdx, int rDegree, const std::vector<double> &rKnots)
+    {
+        Eigen::MatrixXd ndu(rDegree+1, rDegree+1);
```
Psirus: What is `ndu`? Consider adding a comment or giving a more descriptive name.

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,478 @@
...
+            {
+                ndu(j,r) = right[r+1] + left[j-r]; // lower triange (knot differences)
+                double temp = ndu(r, j-1)/ndu(j,r);
+                ndu(r,j) = saved + right[r+1]*temp; // upper triangle
+                saved = left[j-r]*temp;
+            }
+            ndu(j,j) = saved;
+        }
+
+        Eigen::MatrixXd ders(der+1, rDegree+1);
```
Psirus: I assume `der` is short for *derivative*, and `ders` for *derivatives*. But `der` is an `int`, and `ders` is a `double` matrix, so `ders` are clearly not multiple `der`, which is what the naming suggests. Consider renaming to something like `derivativeOrder` and `basisFctsDerivatives`.

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,478 @@
...
+
+        for(int j = 0; j <= rDegree; j++) ders(0,j) = ndu(j,rDegree);
+
+        Eigen::MatrixXd a(2, rDegree+1);
+        for(int r = 0; r<=rDegree; r++)
+        {
+            int s1 = 0;
```
Psirus: Is there a book/paper that you could refer to, where one would be able to see the corresponding theory and immediately know what `s1, s2, a, j, k, rk, pk, d, j1, j2` are? If so, please mention it in a comment.

src/mechanics/cell/CellIPData.h
```diff
@@ -13,17 +13,25 @@ template <int TDim>
...
-        return mElements[dofType]->GetInterpolation().GetN(mIPCoords);
+        int dim = mCellInterpolation[dofType]->GetDofDimension();
+        Eigen::MatrixXd N(dim, dim * mCellInterpolation[dofType]->GetNumNodes());
+
+        auto shapeFunctions = mCellInterpolation[dofType]->GetShapeFunctions(mIPCoords);
+
+        for (int i = 0; i < mCellInterpolation[dofType]->GetNumNodes(); ++i)
+            N.block(0, i * dim, dim, dim) = Eigen::MatrixXd::Identity(dim, dim) * shapeFunctions[i];
+        return N;
+//        return mCellInterpolation[dofType]->GetInterpolation().GetN(mIPCoords);
```
Psirus: Remove this line instead of commenting it out.

src/mechanics/interpolation/CellInterpolationBase.h
```diff
@@ -0,0 +1,42 @@
...
+class CellInterpolationBase
+{
+public:
+    CellInterpolationBase(){}
+
+    //! @brief extracts all node values of this element
+    //! @remark virtual to make it testable
+    virtual NodeValues ExtractNodeValues() const = 0;
+
+//    Eigen::VectorXd Interpolate(const NaturalCoords& rNaturalCoords) const
```
Psirus: Again, remove if no longer needed.

src/mechanics/interpolation/CellInterpolationBase.h
```diff
@@ -0,0 +1,42 @@
...
+namespace NuTo
+{
+
+class CellInterpolationBase
+{
+public:
+    CellInterpolationBase(){}
+
+    //! @brief extracts all node values of this element
+    //! @remark virtual to make it testable
```
Psirus: What is meant by "virtual to make it testable"? Isn't it "virtual to make it polymorphic"? (which would be a superfluous remark).

src/mechanics/interpolation/CellInterpolationFEM.h
```diff
@@ -0,0 +1,59 @@
...
+
+class CellInterpolationFEM : public CellInterpolationBase
+{
+public:
+    CellInterpolationFEM(std::vector<NuTo::NodeSimple*> rNodes, const InterpolationSimple& rInterpolation)
+        : mNodes(rNodes)
+        , mInterpolation(rInterpolation)
+    {}
+
+    //! @brief extracts all node values of this element
```
Psirus: Inherited member functions inherit their documentation, no need for this comment.

Psirus (changes requested):
There are a couple of issues if noticed. Also, please use clang-format.

potto1:
Thanks @Psirus for reviewing

joergfunger (approved):

src/mechanics/cell/CellIPData.h
```diff
@@ -13,17 +13,24 @@ template <int TDim>
...
+    CellIPData(const DofContainer<CellInterpolationBase*> cellInterpolation, const NuTo::Jacobian<TDim>& jacobian,
                const NaturalCoords& ipCoords)
-        : mElements(elements)
+        : mCellInterpolation(cellInterpolation)
         , mJacobian(jacobian)
         , mIPCoords(ipCoords)
     {
     }
 
     NMatrix GetNMatrix(const DofType& dofType) const
```
TTitscher, line 24: IMO it is beneficial to cache the whole N matrix for FEM. I saw the performance boost in the visualization, when I introduced it.  
If the calculation of N is here in the CellIpData, the cache can only live for one evaluation of a cell and is useless. Can we somehow move that back to CellInterpolationFEM? Is an N matrix like that also required in IGA? 

My proposal: We keep it as it is for now and I'll handle that in a separate pull request after this one here is merged. But please try to answer the question. 
There is also this [todo test](https://github.com/nutofem/nuto/blob/master/test/mechanics/todo/todoCache.cpp) where I outlined a possible test for the feature. I can try to get that running (in this separate pull request).

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,540 @@
...
+Eigen::MatrixXd NURBS<1>::BasisFunctionsAndDerivativesRational(int der,
+                                                               const Eigen::Matrix<double, 1, 1>& parameter) const
+{
+    assert(der >= 0 || der <= 2);
+
+    int spanIdx = FindSpan(parameter)[0];
+    Eigen::MatrixXd ders = BasisFunctionsAndDerivatives(der, parameter[0], spanIdx, mDegree[0], mKnots[0]);
+
+    // NURBS specific ...
+    Eigen::VectorXd sum(der + 1);
```
TTitscher: Following [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) you could write
~~~c++
Eigen::VectorXd sum = Eigen::VectorXd::Zero(der+1);
~~~

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,540 @@
...
+class NURBS
+{
+public:
+    enum mParametrization
+    {
+        chord,
+        centripetal
+    };
+
+    /** Constructors **/
```
TTitscher: IMO we do not need this comment. The `/** xyz **/` below as well. If you want your template class to look cleaner, you could only put the declarations here and move the implementation below the class. As you did with the template specializations.

Psirus:
I hate to be this person, but I've only noticed this now: the part that seems the most complex in the `NURBS` class is actually not tested, [see here](https://codecov.io/gh/nutofem/nuto/compare/8a15ba2b59e0be64b4dd0778d198f82d68389565...4380e59550b4934d5260a5245e361e4c4c92c080/diff). I think at least this for-loop in `BasisFunctionsAndDerivatives` should really be tested. Also, the comment about "don't throw strings, throw exceptions" applies to a couple more locations in the code.

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,540 @@
+#pragma once
+
+#include "mechanics/nodes/NodeSimple.h"
+#include "base/Exception.h"
+#include <array>
+#include <eigen3/Eigen/Dense>
```
TTitscher: could Eigen/Core be enough?

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,540 @@
...
+    };
+
+    /** Constructors **/
+
+    //! @brief ... constructor
+    //! @param rDegree ... degree of the polynomial
+    //! @param rKnots ... knot vector
+    //! @param rControlPoints ... control points
+    NURBS(const std::array<std::vector<double>, TDimParameter>& knots,
+          const std::vector<std::vector<NodeSimple>>& controlPoints, const std::vector<std::vector<double>>& weights,
```
TTitscher: This class could provide a very general NURBS definition. So I would ask you to save the control points in some other format (Eigen::VectorXd or Eigen::Matrix<double, TDimParameter, 1>?) to decouple it from `NodeSimple`.

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,540 @@
...
+    //! @param spanIdx ... spanIdx to the parameter
+    //! @param degree ... degree of the given NURBS curve
+    //! @param knots ... knot vector
+    //! @param weights ... weights vector
+    //! @return Eigen::VectorXd ... vector containing the shape functions (no derivatives)
+    static Eigen::VectorXd BasisFunctionsAndDerivativesRational(int derivativeOrder, double parameter, int spanIdx,
+                                                                int degree, const std::vector<double>& knots,
+                                                                const std::vector<double>& weights)
+    {
+        assert(derivativeOrder >= 0 || derivativeOrder <= 2);
```
TTitscher: derivativeOrder values of -12 or 6174 will not trigger this assertion. You should use &&.

src/mechanics/IGA/NURBS.h
```diff
@@ -0,0 +1,540 @@
...
+        {
+            parameterIDs[i] = FindSpan(parameter[i], mDegree[i], mKnots[i]);
+        }
+        return parameterIDs;
+    }
+
+    Eigen::MatrixXd BasisFunctionsAndDerivativesRational(int der,
+                                                         const Eigen::Matrix<double, TDimParameter, 1>& parameter) const
+    {
+        throw new std::string("Not implemented for an arbitrary dimension!");
```
TTitscher: Like @Psirus said, please throw a NuTo::Exception instead. And you could be more specific:
~~~cpp
throw NuTo::Exception(__PRETTY_FUNCTION, "Only implemented for TDim = 1,2. "); // or something like that
~~~

 But I like to point out another bug/error/... . A raw `new` should never appear in our code.

TTitscher (changes requested):
I first thought of implementing IGA as CellIga and CellFem. Your approach, however, allows mixing both. I could have IGA coordinates, FEM displacements and IGA temperatures. I am not sure, if that makes sense - especially if you think about automatically creating the corresponding meshes, but in general it is possible. 

I have some specific requests before I can accept this request, mainly in the NURBS class:
- decoupling from any mechanics class (`NodeSimple` in this case), then move to `/math/`?
- complete documentation
- cover all the complicated calculations with tests

There is another issue regarding the naming. We agreed on avoiding these UPPERCASE names (even for acronyms) and replace it with CamelCase. This would affect
- `class Nurbs`
- `class CellInterpolationIga`
- `class CellInterpolationFem`

__Edit:__ Please tell me, if you need assistance. I like to help!

src/mechanics/IGA/NURBS.h
Psirus: If we do that, I think we should move it to `math/`.  Seems like the more appropriate location.

src/mechanics/IGA/NURBS.h
Psirus: Or even `and`.

src/mechanics/cell/CellIPData.h
joergfunger: IMO, there the N matrix is not possible to cache for IGA, since the shape functions are not only a function of the natural coordinate, but also from the knot vector. Certainly, this could be circumvented by making an IGA-cell and a FE cell. However, for a better generality of the code (and maybe sacrificing some performance) I would opt for just defining an interpolation that only provides shape functions and its derivatives. Everything else such as the calculation of B should be done somewhere else. I would even ask the question, if we really need to store these operators in a class such as cellIpData, or if this could not just be global operator function that gets the shape function derivatives and returns the B matrix (or any other operator). From my point of view, these operators are strongly related to the PDE, and not the cell interpolation, and should therefore not be stored at the interpolation routines.

joergfunger:
The usage of NodeSimple in the IGA implementation is done in order to be able to interpret the free parameters (dofs) of the interpolation (nodes in the case of FEM, and control points in the case of IGA) in the same way. The control points in IGA have the same dimension, dof numbers related to the solution vector and so on. I suggest to keep the NodeSimple in the IGA interpolation, since it is essentially identical to a node in the FEM model. Then, all further operations (such as constraints, application of the solution, numbering of dofs) can then be done on the same data structure.

src/mechanics/cell/CellIPData.h
TTitscher: Sure, storing N for IGA is not feasible, not the question. But does a interpolation in IGA work with the N matrix? Say, for some boundary forces, is `Integrate(GetN() * pressure)` valid for IGA as well? Then it could be easily moved to `CellInterpolationBase`. `CellInterpolationIga` calculates N, `CellInterpolationFem` provides a precalculated N. Should be no problem, right?

And for me, this caching is crucial. And apart from that, I want to be able to define a `CellInterpolationFemRegularGrid` that is equal for all elements where everything is precalculated. Or maybe that is a special CellIpData class...

TTitscher:
Then, the Nurbs class might be too complex, it provides the nurbs interpolation _and_ handles the degrees of freedom. The member mControlPoints is only used in the two methods `GetDimension()` and `GetControlPointsElement()`. The latter one looks like a "ExtractNodeValues". Most of the remaining functions are static. 

For me, it is easier to understand if you move the `mControlPoints` and `GetControlPointsElement()` to `CellInterpolationIga` (and rename it to the interface method `ExtractNodalValues`). Then `Nurbs` is responsible for only the math (provides basis functions and derivatives) - like InterpolationSimple. And `CellInterpolation` combines this math with some degrees of freedom. 

I think that it would be useful to have a general nurbs implementation in NuTo, similar to [Legendre](https://github.com/nutofem/nuto/blob/master/src/math/Legendre.h), [LinearInterpolation](https://github.com/nutofem/nuto/blob/master/src/math/LinearInterpolation.h) or [CubicSplineInterpolation](https://github.com/nutofem/nuto/blob/master/src/math/CubicSplineInterpolation.h). And CellInterpolationIga uses this nurbs implementation for IGA elements.

src/mechanics/cell/CellIPData.h
joergfunger: Yes, the N matrix works in the same way as with standard FEM for the interpolation, thus calculating the equivalent force vector for a pressure function is also Integrate N * pressure. But isn't this already an specific application (integration of some mechanics force) that is more related to the PDE and not to the interpolation? But anyway, if there is  a significant performance boost, we can move this to interpolation base (and B as well). 
The definition of your own CellInterpolationFemRegularGrid with a fixed number of nodes, dimensions, integration and even a defined linear elastic law or even a hard coded stiffness should not be a problem at all. It would be interesting to see the performance differences.
That would by the way even work for the template concept, where the cell is derived from the integrand interface class, and then the operator includes an additional function that takes as an argument not the integrand but a cell. The only additional feature here is to distinguish between a cell interface and an ip interface.

potto1:
But `GetControlPointsElement()` is used in `Evaluate()` ...

What Thomas suggests, is implemented in the current NuTo version. IMO moving the control points to each element, means we don't have a "general nurbs implementation in NuTo". We need to pick out an element, and evaluate the NURBS there. So then we only have structural analysis with NURBS discretization and no NURBS geometry. The implementation here suggests to handle a NURBS as one element on the one side, but assembling of the stiffness is still done for each span of knots, because we have implemented `CellInterpolationIGA`. So element vectors and matrices are still dense.

When one just wants to construct NURBS geometry without `NodeSimple`, we could provide a constructor parsing it.

TTitscher:
OK, I see. Another constructor won't help, I thought of a complete decoupling. But I see that this not as straight forward. Maybe something like `/math/Nurbs.h/` and `/mechanics/Iga/InterpolationIga.h` (with controlpoints, Extract() and Evaluate()). This is optional, I guess :ghost:

I just saw a potential bug. [IMO, this should be a const reference.](https://github.com/nutofem/nuto/pull/103/files#diff-673b6f1e7644067321e172ff8040c84aR337) So if a `NodeMerge` is performed, these should update, right?

potto1:
Yeah, `const std::array<int, TDim>& knotID` is better.
`NodeMerge` for IGA is, from the theoretical point of view, not as "simple" as for FEM. So there will be a routine providing at least knot insertion. Knot insertion then computes the right control point such that the geometry is not changed.

TTitscher:
I meant NodeMerge that is performed after each solution of the global system. And maybe I linked the wrong line. After each solution, the NodeSimple should have new values. So they have to be references.

potto1:
Yes, thats right.

TTitscher:
It seems like there is still some missing documentation. On the other hand, there is missing documentation everywhere... And it seems like the Nurbs class is not covered for second order derivatives. If you can come up with a quick test on that, that would be nice.

TTitscher (approved):

Psirus (approved):

TTitscher:
@joergfunger just mentioned that `CellInterpolationBase::GetNMatrix()` seems to be a bit out of place. Since NMatrix only depends on the natural element coordinates, it can be cached has to be stored somewhere. And IMO it fits there quite well, since we define our interpolation as
~~~cpp
CellInterpolationBase::GetNMatrix() * CellInterpolationBase::ExtractNodeValues();
~~~

Btw: The code for the NMatrix is now [here](https://github.com/nutofem/nuto/blob/PDE_IGA/src/mechanics/cell/Matrix.h#L8) so we could easily move the GetNMatrix() around.  We could add BGradient(), BStrain(),... there as well...

TTitscher:
There is another crucial difference in IGA and FEM. `CellInterpolationFem::GetDerivativeShapeFunctions()` should return a __const reference__ to precalculated derivative shape functions. Same goes for `GetNMatrix()` and `GetShapeFunctions()`. IGA has to return __by value__. Damn that I did not see that earlier.
I removed this whole caching topic for now and suggest we postpone this issue until we can run suitable benchmarks.

PR #104, opened 2017-09-08
split cell into element and cell

joergfunger:
The current branch demonstrates the concept of splitting the cell into an interpolation part (now called PDE_element or simply element) and the integration part with the quadrature scheme and the integrands). See also issue #43 . The cell is now also proposed to be a template, since the structure does not have to be a template any more. This pull request is not to be merged, but to show the differences between the master and the proposed approach and have a design discussion on each proposed class/method definition with a side-by-side overview about the changes.


TTitscher (changes requested):
As far as I understand the change, this pull request should have been about moving the two members `mCoordinateElement` and `mElements` to a new class `PDE_Element`. This seems like a good idea.

But on top of that, this pull request introduces numerous changes. It reimplements a lot of things, we already have (MechanicsLaw, Voigt stuff, IntegrandMomentumBalance). And it removes the `template <int TDim>` from jacobian (the call to `.inverse()` on a dynamically sized matrix is noticeable expensive). And it introduces `template <typename TIntegrand>` to the cell (we discussed that in our group and decided against it).

joergfunger:
I have committed changes to re-add the template of the Jacobian. 

Regarding the criticism from @TTitscher  a few remarks from my side.  I m very open to any other suggestions, in particular, since I'm certainly not one that is working with NuTo on a daily basis. I also do not want to make everything as general as possible without knowing what the future needs will be. However, the current approach should at least cover the standard backward Euler time integration schemes with Gradient, Hessian0..2 as well as the CycleJump approach the requires (elastic stiffness and gradient without changes of the history variables, an integration of a single cycle with substeps and the calculation of the delta history variables and extrapolation routine and some transfer routine that allows to couple standard Newmark to
 perform the first few cycles in a direct Newmark solution, and then do the cycle jump afterwards and move back at the end to Newmark)

- "But on top of that, this pull request introduces numerous changes. It reimplements a lot of things, we already have (MechanicsLaw, Voigt stuff, IntegrandMomentumBalance)."

What exactly do you mean by this? We currently do not have neither a standard Mechanics law with a pure stress strain functionality, nor an IntegrandMomentumBalance (or am I wrong?). In the current (PDE-reviewed) branch, there is an explicit linear elastic momentum balance integrand. Even though I think having the general momentum balance (with some base class that provides the stress strain relation) is more general (since e.g. a Mises Law can be directly implemented on the material level and not on the integrand level), I don't see a reason why those two integrands cannot be kept in parallel (in particular, if the linear elastic integrand really provides performance advantages).

- And it removes the template <int TDim> from jacobian (the call to .inverse() on a dynamically sized matrix is noticeable expensive).
I've reintroduced the template. However, the only reason for the cell being a template (of the dimension) is the Jacobian. Everything else is handled (at least currently) with dynamic matrices, e.g. (ipCoords or  the TransformDerivativeShapeFunctions in the Jacobian with DerivativeShapeFunctionsNatural also having a dynamic size). If it the problem is only the inverse function, this could be handled differently (e.g. by having a Base class that provides a general - not templated - interface, then the cell does not need to be a template). We should also keep in mind that at some point we will add lower dimensional elements in higher dimensions, such as a 1D truss in 3D. The cell/element integration dimension is certainly 1D, but there has to be a conversion from the 3D global coordinates to 1D (certainly in the PDE_element or similar class when the NodeValues are extracted) - just have to make sure not to mix up these dimensions.

-And it introduces template <typename TIntegrand> to the cell (we discussed that in our group and decided against it).

In the group, we discussed the changes with respect to the structure being a template (which we disagreed on, but we also disagreed on passing all parameters to the integrand via shared references). So in my current proposal, the structure is not a template any more. My suggestion is to divide the current structure into 4 components

-- a cell container that only stores the cell (and a reference to the element that includes the interpolation). This is essentially only a list of the history variables with a reference to the global position/interpolation at this point. It also makes it easier to perform operations on the history data)
-- an element container that stores the elements (essentially the interpolation function)
-- a solution vector 
-- an assembler that includes the constraints and provides the interface to map the active dof solution to the complete one - including dependent and independent dofs

A time integration scheme can then evaluate a function such as the gradient with a given set of history data and a given solution vector. This would avoid the imo difficult to understand routines such as node merge that automatically does some magic to include the dependent variables. 

Regarding the flexibility of NuTo as a lib, I'm still convinced that the operator concept has some advantages. Instead of 
`
   auto myVector = CellContainer.Gradient(t,delta_t);
`
the interface would look like
`
   auto myVector = CellContainer.EvaluateVector(GradientOperator(t,delta_t));
`
which I do not consider as more difficult to understand.

Psirus (commented):
**This is not a comment about whether we should use this operator-template-approach or not.**

Introducing this template here is a major design decision that multiple people feel uneasy about. Maybe it's the way to go, maybe it isn't. But decisions such as this should be discussed elsewhere beforehand, and not be tried to "slip by" unnoticed in a PR about something different (splitting interpolation from integration).

joergfunger:
Maybe the initial commit was not correct, but the last comment clearly discusses this. I've updated the initial comment.

Psirus:
I still think that PRs should be discussions about code, not design. Seems counterproductive for someone to *implement* a design, other people evaluating a design based on possibly hundreds of lines of code, and in the end maybe rejecting it. All the implementation work was then done in vain.

TTitscher:
> but we also disagreed on passing all parameters to the integrand via shared references

I was not present, when this decision was made... Anyways.

With the reimplementation, I specifically meant [LinearElastic.h](https://github.com/nutofem/nuto/blob/PDE_reviewed/src/mechanics/constitutive/laws/LinearElastic.h) and [IntegrandLinearElastic.h](https://github.com/nutofem/nuto/blob/PDE_reviewed/src/mechanics/cell/IntegrandLinearElastic.h). I agree that a IntegrandMomentumBalance is more versatile. Following the YAGNI principle (to keep the pull requests as small as possible), there is no reason to change that now, since we only have LinearElastic. If we (later, in another pull request) introduce more laws, we should certainly think about clever abstractions.

joergfunger:
As added in the initial comment, this pull request is not supposed to be directly merged into the master, but to allow for a discussion on the design concepts (now labeled design). Even though this might cause some additional (and maybe wasted) amount of code changes and time, it provides a simple to read overview about the design differences and source code that already compiles with the option to directly discuss these changes.

TTitscher:
From a technical view, the new PDE_Element contains two references. I see that both somehow belong together, but I would rather implement it as public members like
~~~cpp
class PDE_Element
{
public:
    const ElementInterface& mCoordinateElement;
    DofContainer<const ElementInterface*> mDofElements;
};
~~~
to avoid forwarding all methods. What do you think?

TTitscher:
I implemented the following nice concepts introduced in this pull request in #112. 
(_I thought it was easier to start from the more up-to-date PDE_review branch than updating this branch._)
- PDE_Element (simply called Element)
- operations concept
- MechanicsInterface for laws
- IntegrandTimeDepentent
- IntegrandMomentumBalance

IMO, as soon as #112 is accepted, we can close this pull request. Please tell me if I forgot some features. I'll add them asap.

PR #107, opened 2017-09-22
Pde renaming to Element

TTitscher:
Just a renaming issue. `CellInterpolation...` contains an interpolation and dof values. For me, and some ppl I talked to, this should be called `Element`.

Plus, some minor changes regarding some documentation and renaming (remove r... in arguments). No big deal IMO.

### DIFF
--- a/src/mechanics/cell/CellIpData.h
+++ b/src/mechanics/cell/CellIpData.h
@@ -13,9 +13,9 @@ template <int TDim>
 class CellIpData
 {
 public:
-    CellIpData(const DofContainer<ElementInterface*> cellInterpolation, const NuTo::Jacobian<TDim>& jacobian,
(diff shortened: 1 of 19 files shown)


src/mechanics/cell/CellIpData.h
```diff
@@ -13,7 +13,7 @@ template <int TDim>
 class CellIpData
 {
 public:
-    CellIpData(const DofContainer<CellInterpolationBase*> cellInterpolation, const NuTo::Jacobian<TDim>& jacobian,
+    CellIpData(const DofContainer<ElementInterface*> cellInterpolation, const NuTo::Jacobian<TDim>& jacobian,
```
Psirus: Should this not also be `elements` then instead of `cellInterpolation`?

Psirus (approved):

PR #108, opened 2017-09-22
Pde simple assembler

TTitscher:
Here is a proposal for a simple assembler. I call it simple, because it requires minimal information: 
- number of active dofs (per dof type)
- number of dependent dofs (per dof type)
- cell container
- dof types to assemble

and its only task is the assembly. The numbering, constraints, whatever has to be done by another class.

I chose to pass the first two arguments via the ctor and the last two arguments to the actual assembly methods to keep the argument lists short. We could easily build static assembly functions by passing all four as arguments. Do you think this is beneficial?

Currently, this pull request only contains a unit test where the `CellInterface` is mocked. This means that the following features are not implemented yet:
- constraints
- global dof numbering
- local dof numbering (integrand level)

I plan on adding them soon (help appreciated :ghost:), but I'd like to have some early feedback.


src/mechanics/cell/SimpleAssember.h
```diff
@@ -0,0 +1,144 @@
...
+                        gradient.J[*dof][globalDofNumber] += globalDofValue;
+                    else
+                        gradient.K[*dof][globalDofNumber - numActiveDofs] += globalDofValue;
+                }
+            }
+        }
+        return gradient;
+    }
+
+    GlobalDofMatrixSparse BuildHessian0(const std::vector<NuTo::CellInterface*>& cells,
```
Psirus: `BuildHessian1` and `BuildHessian2` are going to be copy&paste programming, or something else?

src/mechanics/cell/SimpleAssember.h
```diff
@@ -0,0 +1,144 @@
...
+                                            .coeffRef(globalDofNumberI - numActiveDofsI, globalDofNumberJ) +=
+                                            globalDofValue;
+                                }
+                                else
+                                {
+                                    hessian.KK(*dofI, *dofJ)
+                                            .coeffRef(globalDofNumberI - numActiveDofsI,
+                                                      globalDofNumberJ - numActiveDofsJ) += globalDofValue;
+                                }
+                            } // argh. any better ideas?
```
Psirus: - Is there not a function that takes lists of "i", "j" and "value"?
- Would it be possible/advantageous to assemble everything into KK first, and then "reorder" in a second step?

test/mechanics/cell/SimpleAssembler.cpp
```diff
@@ -0,0 +1,96 @@
...
+     *   21    22    23 ||             active J
+     *                  ||
+     *   31    32    44 || 12    13
+     *   ===========================
+     *               21 || 22    23
+     *                  ||             dependent K
+     *               31 || 32    33
+     *
+     *     active J         dependent K
+     */
```
Psirus, line 92: Nice diagrams!

src/mechanics/cell/SimpleAssember.h
TTitscher: If we chose to keep this Hessian012 syntax, code duplication can be avoided with something like

~~~cpp
template <typename TFunction>
GlobalDofMatrixSparse BuildMatrix(cells, dofs, TFunction f)
{ ...
    hessian = f(cell);
... }

GlobalDofMatrixSparse BuildHessian0(cells, dofs)
{
    auto f = (CellInterface& cell) {return cell.Hessian0();};
   return BuildMatrix(cells, dofs, f);
}
~~~

(Same pattern can be used to avoid possible code duplication inside the cell.)

joergfunger: Extending this approach with the function template is not necessary, if we only assemble a scalar, a vector or a sparse matrix. Imo, the operator concept should here really be used, the operator is essentially deciding what function should be called
```
GlobalDofMatrixSparse BuildSparseMatrix(const std::vector<NuTo::CellInterface*>& cells,
const std::vector<NuTo::DofType*>& dofTypes, const MatrixOperator& operator) const
{
   for (NuTo::CellInterface* cell : cells)
   {    

      const DofMatrix<double> cellMatrix = cell.Integrate(operator);
   }
}
```
and on the cell level, this evaluate loops over the ips and assembles the contribution from each ip
```
DofMatrix Cell::Integrate(Operator& operator) const
{
   cellData = ...
   for (Integrands: integrand)
   {    
      ipData = ...
      const DofMatrix<double> cellMatrix += operator.Evaluate(integrand, ipData, cellData);
   }
}
```
In the Evaluate method of the operator for the hessian0, ...(or any other function of the integrand) is called.
```
DofMatrix Hessian0::Evaluate(integrand, ipData, cellData) const
{
  return integrand.Hessian0(ipData, cellData);
}
```

src/mechanics/cell/SimpleAssember.h
```diff
@@ -0,0 +1,144 @@
...
+        for (NuTo::CellInterface* cell : cells)
+        {
+            const DofVector<int> numbering = cell->DofNumbering();
+            const DofVector<double> cellGradient = cell->Gradient();
+
+            for (const DofType* dof : dofTypes)
+            {
+                const Eigen::VectorXi& numberingDof = numbering[*dof];
+                const Eigen::VectorXd& cellGradientDof = cellGradient[*dof];
+                const int numActiveDofs = mNumActiveDofs[*dof];
```
joergfunger: I would prefer an interface, where element vectors/matrices are always returned together with the corresponding row dof/col dof vector. For standard FEM stuff, this is not much of an issue, there is no difference between calling the function sequentially (first vector, then dofs) or simultaneously. However, for contact problems, we will have for each slave element (over which we integrate) a projection of the master side, thus the possible number of interacting dofs is a function of the current displacement state. As a consequence, we could either return a sparse stiffness matrix (mmh, not my favorite) or return a full matrix with only the relevant dofs on the master, but this requires again the calculation of the projection, if it is done sequentially in two separate function calls.

src/mechanics/cell/SimpleAssember.h
```diff
@@ -0,0 +1,144 @@
...
+                for (int i = 0; i < numberingDof.rows(); ++i)
+                {
+                    int globalDofNumber = numberingDof[i];
+                    double globalDofValue = cellGradientDof[i];
+                    if (globalDofNumber < numActiveDofs)
```
joergfunger: I would suggest to distinguish between active/inactive dof types and dependent/independent dofs (for the naming). So in this case it would be numIndependentDofs. Furthermore, I would suggest to base the decision on whether a dof is dependent or independent not on the dof number that is stored at the node, but rather on a list in the assembler. That means that each dof is numbered at the very beginning (without distinguishing between dependent and independent dofs). When adding constraints, only the assembler knows which global entry (including dependent or independent) dofs belongs to a specific dof. In particular, a renumbering (with setting this specifically at the node) is avoided when adding constraints - this can all be encapsulated in the assembler.

src/mechanics/cell/SimpleAssember.h
```diff
@@ -0,0 +1,144 @@
...
+            {
+                m.JJ(*dofI, *dofJ).resize(mNumActiveDofs[*dofI], mNumActiveDofs[*dofJ]);
+                m.JK(*dofI, *dofJ).resize(mNumActiveDofs[*dofI], mNumDependentDofs[*dofJ]);
+                m.KJ(*dofI, *dofJ).resize(mNumDependentDofs[*dofI], mNumActiveDofs[*dofJ]);
+                m.KK(*dofI, *dofJ).resize(mNumDependentDofs[*dofI], mNumDependentDofs[*dofJ]);
+            }
+        return m;
+    }
+
+    NuTo::DofContainer<int> mNumActiveDofs;
```
joergfunger: Similar with the naming here (active/inactive, dependent/independent)

joergfunger (changes requested):
Even if we decided on starting an implementation without any global parameters to be passed to the integrands, thus avoiding the discussion on the template concept, I would still prefer the operator concept in the assembler. IMO, the assembler does not have to know, what kind of matrix is assembled.

test/mechanics/cell/SimpleAssembler.cpp
joergfunger: How do we want to use the GlobalDofMatrix in the time integration scheme? If we want to avoid copying/moving the ii part of the matrix, we will have to expose this matrix somehow. This would then invalidate this GlobalDofMatrix (since we only change the ii component and use this afterwards.  Wouldn't it be possible, not to return a GlobalDofMatrix that itself has the 4 submatrices as members, but directly a tuple of these 4 matrices, thus avoiding a GlobalDofMatrix type at all. In addition, what would be the disadvantage if assembling everything into the complete matrix and then use the block operations of eigen to access the submatrices?

TTitscher: This `GlobalDofMatrixSparse` is quite equivalent to the currently used `StructureOutputBlockMatrix`, except that it uses the `DofType` instead of the `enum class eDof`. 
And its usage in the time integration scheme should also be equivalent. 

I introduced a type (instead tuple/vector/array) mainly because the access via JJ, JK, KJ, KK is better than via index. Is [2] == JK or KJ? On top, you can later define handy methods like a custom outstream or arithmetic operations.

src/mechanics/cell/SimpleAssember.h
TTitscher: This SimpleAssembler aims to be the another fast step towards a first solution of a global system in the PDE concept. 

In our current NuTo, there is no feature similar to what you described here. And if I recall correctly, @potto1 implemented a [contact element](https://github.com/nutofem/nuto/blob/master/src/mechanics/elements/ContinuumContactElement.h) without a special dof numbering. So if someone decides to implement it in the future in the way you described, we need to touch the assembler again. But how should I anticipate the details/syntax of such an implementation? How should I test that?

src/mechanics/cell/SimpleAssember.h
TTitscher: I totally agree on the naming. I just thought: is a constrained value technically still a degree of _freedom_? We could introduce just `dofs` for all active dofs and something else for the constrained values. What do you think?

For the decision whether a dof is dependent or independent, I again like to refer to our current NuTo version. I works like a charm. This way, the assembler is not involved in the dof numbering or the constraints. This a big advantage, since it nicely separates all those tasks. 
[I am aware that this simple assembler _might_ (we never implemented it...) not be able to work in an MPI context. This might require a `class MpiAssembler` and imo is whole different problem.]

src/mechanics/cell/SimpleAssember.h
TTitscher: The more we implement in the "_not_ operator/template concept", the more work is required to change it later. I am aware of that and I want to ensure that I happily adapt all the code to any concept we decide on.
But AFAIK, for now, the plan is to build the most simple running version that is able to solve something like a linear elastic patch test. This introduces various challenges (dof numbering, node selection, mesh class, ConvertToCells(), ... ). So for now, there is only Gradient() and Hessian() and imo no need to make it more complicated that that.

My code above is just _one_ way to assemble an arbitrary cell method.

src/mechanics/cell/SimpleAssember.h
TTitscher: Yeah, you mean an `Eigen::Triplet<>`. You see the usage [here](https://eigen.tuxfamily.org/dox/group__SparseQuickRefPage.html). This sould be more efficient than the `coeffref` thing. IMO we have to perform benchmarks to see, what is best.

TTitscher:
@joergfunger This branch now runs with the operations concept, which, I think, was your requested change. Maybe you could reevaluate your review. :smile:

joergfunger (approved):
I've only had a short overview about the changes, everthing seems to be fine and can be merged. Still want to discuss @Psirus proposal regarding the operator/integrand concept.

Issue #98, opened 2017-08-31
Simplify CMake


Psirus:
Having recently touched almost all CMake files, I've noticed that there are still a lot of unnecessary complications in there. Maybe we can simplify our build system a bit.

# External libraries and static linking

## BUILD_SHARED_LIBS

Does anyone need NuTo as a static lib? For that, you would need to have all the external libs as static libs as well. The size of your executables becomes huge. I don't think this is necessary anymore.

## ENABLE_STATIC_EXTERNALS

This is connected to the previous point: We get our libraries primarily from the Debian repos, compiling them yourself just isn't fun. Also, I think we're at a point where introducing a dependency that is not in the Debian repos or a single header we can put in `external/`, this will meet a lot of resistance. I'd say this should just go as well.

## NUTO_EXTERNAL_LIBRARIES_ROOT

Again, same difference. If you don't expect to find your libs in unusual places, you don't need this variable. Let's get rid of it.

# Preprocessor definitions

**Don't confuse with CMake options of the same name.**

Preprocessor definitions, aka.
~~~cpp
void MyFunction()
{
#ifdef ENABLE_OPTIMIZE
    doThis();
#else
    doThat();
#endif
}
~~~
are very bad for readability and reasoning about the code. They should be kept to a minimum. `ENABLE_VISUALIZE` will probably stay for a while, but `ENABLE_MECHANICS`, `ENABLE_OPTIMIZE` should go. What about `HAVE_MUMPS`, `HAVE_PARDISO`, `HAVE_ARPACK`?

---

- [x] Upgrade to CMake 3.5
- [x] Use Boost::filesystem syntax
- [x] ENABLE_STATIC_EXTERNALS
- [x] NUTO_EXTERNAL_LIBRARIES_ROOT
- [x] ENABLE_MECHANICS
- [x] ENABLE_OPTIMIZE
- [x] ENABLE_VISUALIZE
- [x] ENABLE_METAMODEL
- [x] ENABLE_GEOMETRYCONCRETE
- [x] ENABLE_ARPACK
    - [ ] HAVE_ARPACK?
- [x] `ENABLE_TESTS` + `make test`
- [x] `ENABLE_COVERAGE`
- [x] `ENABLE_EXAMPLES` + `make examples`
- [x] `ENABLE_BENCHMARKS` + `make benchmarks`
- [x] Remove FindOPENBLAS
- [x] Remove FIND_STATIC_LIBRARY
- [x] Move dependency finding to seperate file
[...]

PR #110, opened 2017-09-26
Update cmake version to 3.5 on Travis

Psirus:
**Don't merge this!**

I wanted to update CMake to 3.5 on Travis to start working on #98. However, this doesn't work as you can see. I have no idea what causes this, and I don't know how to fix it (setting `include_directories` is not applied to the swig target module for some reason). If you want to have a crack at this

~~~python
import numpy.distutils as nd
print(nd.misc_util.get_numpy_include_dirs()[0])
~~~

will get you the correct include directory, also on Travis. You just need to tell CMake to use it. Most likely to be done with [target_include_directories](https://cmake.org/cmake/help/v3.0/command/target_include_directories.html). Good luck.

### DIFF
--- a/scripts/install_cmake.sh
+++ b/scripts/install_cmake.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+set -ev
+# 3.5 is in Debian Stable and in 16.04, and has IMPORTED targets for boost
+wget https://cmake.org/files/v3.5/cmake-3.5.2-Linux-x86_64.sh
+sudo sh cmake-3.5.2-Linux-x86_64.sh -- --skip-license --prefix=/usr
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ before_install:
     - export PATH=/home/travis/miniconda/bin:$PATH
     - ./scripts/install_numpy.sh
     - ./scripts/install_gmsh.sh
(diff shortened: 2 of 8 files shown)


Psirus:
**Do merge this!** 🥁 
I've set the path explicitly, and now it seems to work. What made this necessary is beyond me; don't care. Also, I've added a few paths and libnames to `FindMUMPS` and `FindOPENBLAS`. This allows compiling NuTo on Arch Linux/Manjaro.

scripts/install_cmake.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
TTitscher, line 1: I don't know if it is important, but better use
~~~
#!/usr/bin/env bash
~~~
?

TTitscher (approved):
Apparently, it works. I'm fine with it.

scripts/install_cmake.sh
Psirus: Doesn't matter really. Afaik, all Linuxes have their bash at `/bin/bash`. This might not me the case with OSX or some BSD. But a) this script is currently only for Travis, running Ubuntu, and b) if you want to compile NuTo on other OSs, I guess this would be your smallest problem. Feel free to change it though.

PR #111, opened 2017-09-27
Remove template <TDim>. Only unit tests on travis

TTitscher:
This allows also removing it everywhere else (Cell, Integrand, CellIpData)

The jacobian calculation is actually performed on fixed size matrices to keep it efficient. Eigen apparently swaps the algorithm for `.inverse()` and `.determinant()` for small, fixed size matrices. Only the results are saved in a dynamic matrix (max size 3x3). On top, I avoid the strange `nodeBlockValues` loop by using the (far more efficient) `Eigen::Map`. 

To reduce the load on travis, only unit tests are executed.

### DIFF
--- a/src/mechanics/cell/Jacobian.h
+++ b/src/mechanics/cell/Jacobian.h
@@ -6,24 +6,26 @@
 namespace NuTo
 {
 
-template <int TDim>
 class Jacobian
 {
 public:
+    using Dynamic3by3 = Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::ColMajor, 3, 3>;
+
     Jacobian(const NodeValues& nodeValues, const DerivativeShapeFunctionsNatural& derivativeShapeFunctions)
     {
-        const int numRows = derivativeShapeFunctions.rows();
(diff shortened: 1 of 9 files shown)


Psirus (approved):

src/mechanics/cell/Jacobian.h
```diff
@@ -6,24 +6,26 @@
...
-        // x0  x1  x2  x3 ...
-        // y0  y1  y2  y3 ...
-        // z0  z1  z2  z3 ...
-        for (int i = 0; i < numRows; ++i)
-            nodeBlockCoordinates.col(i) = nodeValues.block<TDim, 1>(TDim * i, 0);
-
-        mJacobian = nodeBlockCoordinates.lazyProduct(derivativeShapeFunctions);
-        mInvJacobian = mJacobian.inverse();
-        mDetJacobian = mJacobian.determinant();
+        const int dim = nodeValues.rows() / derivativeShapeFunctions.rows();
```
joergfunger, line 16: Why are you not using dim = derivativeShapeFunctions.cols(), which is also directly the dimension, but without using a division operator?

PR #112, opened 2017-09-27
Pde operations

TTitscher:
This implements the operations concept including casts as discussed in #43. Tomorrow, I plan on adding

- PDE_Element (somehow renamed)
- move some stuff from header to cpp files

But I am open to early criticism.


src/mechanics/integrands/TimeDependent.h
```diff
@@ -0,0 +1,56 @@
...
+namespace NuTo
+{
+namespace Integrand
+{
+namespace TimeDependent
+{
+struct Interface : Base
+{
+    virtual ~Interface() = default;
+    virtual DofVector<double> Gradient(const CellData&, const CellIpData&, double t, double deltaT) = 0;
```
joergfunger, line 17: What about the suggestion here to not directly use the options t and delta_t, and instead pass a GradientOptionsObject that can be easily extended to accomodate other time dependent integrands.

src/mechanics/integrands/MomentumBalance.h
```diff
@@ -0,0 +1,63 @@
...
+namespace Integrand
+{
+namespace TimeDependent
+{
+
+template <int TDim>
+class MomentumBalance : public Interface
+{
+public:
+    MomentumBalance(const NuTo::DofType& dofType, const Laws::MechanicsInterface<TDim>& law)
```
joergfunger, line 21: const of the law would IMO not allow to perform an update of the history data

src/mechanics/integrands/Base.h
```diff
@@ -0,0 +1,29 @@
...
+class Base
+{
+public:
+    virtual std::unique_ptr<Base> Clone() const = 0;
+    virtual ~Base() = default;
+
+    template <typename TIntegrand>
+    TIntegrand& As()
+    {
+        return dynamic_cast<TIntegrand&>(*this);
```
joergfunger, line 18: Why is that now a dynamic_cast, and not done via the boost::any approach?

joergfunger (commented):
The operator() version of the cell is nice, I little more difficult to read, but removes could duplication.
What about the boost::any idea, seems as if this is not used here?

src/mechanics/integrands/Base.h
TTitscher: I somehow thought that I need a polymorphic clone to fill the vector of integrands at each integration point in `Cell`. I have to do some more experimenting, but I might have been wrong. I could maybe just use a `std::vector<boost::any>` (value type( = nice!)) and be done. That would indeed be much better.

src/mechanics/integrands/MomentumBalance.h
TTitscher: That somehow depends on where the history data is stored. I have not thought about that when I implemented the code above. 
- store at law: We somehow have to separate material parameters from history data to avoid duplicate storage of material parameters. 
- store at integrand: We somehow lose the type of the history data. 

So I would prefer your suggestion to store it at the law and make that nonconst. Actually, the integrand then _owns_ the law and we need a `unique_ptr` instead of a reference.

src/mechanics/integrands/Base.h
TTitscher: I may be nice to have a common base class for other common methods like visualization. And if we have a common base class, there is no need to use boost::any.

TTitscher:
I would like to keep the request as it is with the following comments/explanations:

- Moving stuff to cpp files may be too early. I still have to touch the interfaces quite often. Having cpp files would double that work. 
- `boost::any` was a nice idea but we need `boost::any_cast<TimeDependent::Interface&>(...)`. `boost::any_cast` only allows casting to a non-abstract type. I first thought that we need a common base class anyways, e.g. for visualization. But who knows, maybe we solve that with operations as well. I would be happy if anyone could actually make boost::any work in our case.
- I'd like to postpone the problem concerning the history data at the laws. Until we actually implement a law with history data. 

Looking forward for your reviews.

src/mechanics/integrands/Base.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <memory>
+
+namespace NuTo
+{
+namespace Integrand
```
Psirus: Do we need this namespace? I'd rather have `NuTo::Integrand` as the base class. Also, the namespace is called `Integrand`, and the directory `integrands`.

src/mechanics/cell/Cell.h
```diff
@@ -1,89 +1,85 @@
...
-    //! @brief Extracts a vector (each IP) of vectors (several IPValues for the same integrion point) of IPValues
-    std::vector<std::vector<IPValue>> IPValues() override
+private:
+    //! @brief integrates various operations with various return types
+    //! @param op operation to perform
+    //! @param result result value. It is not clear how to properly initialize an arbitrary TResult to zero. Thus, the
+    //! user has to provide it with this argument.
+    template <typename TOperation, typename TReturn>
+    TReturn Integrate(TOperation&& op, TReturn result)
     {
```
phuschke, line 66: You pass a copy of `TReturn return` and copy it again when returning? You have probably explained it before but why don't you want to use a reference here?

There should also be an overloaded (or default argument) `Integrate(TOperation&& op)` function for the cases above imo.

src/mechanics/cell/Cell.h
```diff
@@ -1,89 +1,85 @@
...
+        , mIntegrands()
     {
         for (int i = 0; i < integrationType.GetNumIntegrationPoints(); i++)
-            mIntegrand.push_back(integrand.Clone());
+            mIntegrands.push_back(integrand.Clone().release());
     }
 
-    //! @brief builds the internal gradien
-    DofVector<double> Gradient() override
+    DofVector<double> operator()(const VectorOperation& op) override
```
Psirus: I think these should be called `Integrate`, because that's what they do. Calling `myCell(someOperation)` does not simply "do the operation", but integrates it.

src/mechanics/cell/Cell.h
TTitscher: Are you concerned with performance? You assume that all those values are copied. This is not correct. The compiler is forced to perform a `move` if you return by value. And probably, he will optimize the pass-by-value as well somehow. Even if he does not, the value is supposed to be zero/empty, thus small.

src/mechanics/cell/CellData.h
```diff
@@ -27,13 +32,13 @@ class CellData
 #ifdef _OPENMP
 #pragma omp critical
 #endif
-            nodeValues = mElements[dofType]->ExtractNodeValues();
+            nodeValues = mElement.DofElement(dofType).ExtractNodeValues();
```
phuschke: What is the `#pragma omp critical` here for?

src/mechanics/elements/Element.h
```diff
@@ -0,0 +1,39 @@
...
+    {
+    }
+    Element(const ElementInterface& coordinateElement, DofContainer<const ElementInterface*> dofElement)
+        : mCoordinateElement(coordinateElement)
+        , mDofElements(dofElement)
+    {
+    }
+
+    void AddElement(const DofType& dofType, const ElementInterface& dofElement)
+    {
```
phuschke: Naming implies that an Element object can add another Element object to itself... That does not make sense.

src/mechanics/elements/Element.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+#include "mechanics/elements/ElementInterface.h"
+#include "mechanics/nodes/DofContainer.h"
+
+namespace NuTo
+{
+class Element
```
Psirus: Naming again: an `Element` does not derive from `ElementInterface`, and contains several "elements"? Could be confusing. Maybe `ElementStack`, `ElementDeck`, `ElementPack`, `ElementCollection` or `ElementAggregate`?

src/mechanics/integrands/Base.h
```diff
@@ -0,0 +1,29 @@
...
+{
+class Base
+{
+public:
+    virtual std::unique_ptr<Base> Clone() const = 0;
+    virtual ~Base() = default;
+
+    template <typename TIntegrand>
+    TIntegrand& As()
+    {
```
phuschke, line 17: Out of curiosity: Does template argument deduction work here?

src/mechanics/elements/Element.h
TTitscher: I agree. How about renaming `ElementInterface` instead? `SingleDofElementInterface` or something. I think its nice to have one `Element` at the `Cell`.

phuschke (commented):
What about the input and output maps? Is there no chance to avoid all the cases?

src/mechanics/cell/Cell.h
TTitscher: Ok, then we would have `Integrate` for `Scalar/Vector/MatrixOperation` and `Apply` for `VoidOperation`? Or better keep `operator()` for the `VoidOperation`?

src/mechanics/elements/Element.h
Psirus: But you have several. The `ElementInterface` elements are named correctly imho.

src/mechanics/cell/Cell.h
phuschke: It is not he performance I am worried about but the way this functions is being used.

This
```
TReturn result(bla);
Integrate(op, result);
```

is much easier to understand than passing a temporary object. I agree that you save 1 line of code though...

```
TReturn result = Integrate(op, TReturn(bla));
```

TTitscher: Save two lines :)
~~~
return Integrate(op, ReturnObj(...));
// vs
ReturnObj bla;
Integrate(op, bla);
return bla;
~~~

src/mechanics/integrands/Base.h
TTitscher: Nope
~~~cpp
struct Base
{
    virtual ~Base() = default;
    template <typename T>
    T& As()
    {
        return dynamic_cast<T&>(*this);
    }
};
struct Foo : Base {};
int main()
{
    Foo foo;
    Base& base = foo;
    Foo& fooCasted = base.As();
}
~~~

~~~
template_deduction_return.cpp:20:27: error: no matching member function for call to 'As'                                                                                                                          
    Foo& fooCasted = base.As();                                                                                                                                                                                   
                     ~~~~~^~                                                                                                                                                                                      
template_deduction_return.cpp:7:8: note: candidate template ignored: couldn't infer template argument 'T'                                                                                                         
    T& As() 
~~~

src/mechanics/integrands/Base.h
TTitscher: Hmm. In the directory, you have many integrands. I just saw that the laws in /constitutive/laws are also in namespace `Laws` instead of `Law`. I think it's reasonable to have a singular word in the namespace. If I'm the only one, I'll change it!

Anyways, I vote for keeping a namespace. Maybe, that looks a bit strange for the base class. The alternative is to have the name `Integrand` in each class name. /mechanics/integrands/IntegrandTimeDependent.*/ (Or do you have another suggestion?).  I'd rather use `Integrand::TimeDependent`.

TTitscher:
@phuschke I do not get your last comment
> What about the input and output maps? Is there no chance to avoid all the cases?

Can you explain that a bit more? Or give a reference to some code?

phuschke:
@TTitscher 
Element2dInterface and ContinuumBoundaryElement. But those files are just remains of the old Version and you have not touched them. Why are they still in there?

TTitscher:
Oh, I explained that here 73e0d94a70770934898f296651717d60c1159d88. As @Psirus suggests, we should come up with a new name to `NuTo::Element`. Then, I can revert this commit.

src/mechanics/cell/Cell.h
TTitscher: Addressing your comment regarding a default argument. If you do not provide e.g. `DofVector()` as an argument, the template method is not able to deduce the type of `TReturn`. But I think that this still nicely eliminates code duplication.

src/mechanics/integrands/MomentumBalance.h
TTitscher: Btw: You can have a look/edit at the [Project PDE](https://github.com/nutofem/nuto/projects/1). I added your remark there.

src/mechanics/integrands/Base.h
TTitscher: The `boost::any` topic is also content of the _neat ideas_ in [Project PDE](https://github.com/nutofem/nuto/projects/1).

Psirus (approved):
Do we have a use-case for `Apply` in mind? I'd say let's get rid of it for now, it doesn't have to go through all the iterations if we don't use it yet. Otherwise we should add a test case.

TTitscher:
UpdateHistoryData would be a use case. But you are right, following the yagni approach, I should not have implemented it yet. Feel free to remove it ;)

src/mechanics/integrands/MomentumBalance.h
joergfunger: I thought that the history data is part of the integrand, there is (at least in the standard case) no law. The only thing I would use is a reference to a list of material parameters (or similar a class that stores these values such as a class MaterialParameters::LinearElastic. Why do we loose the type here, since the Stress routine of the law is virtual and in the derived class (that stores the history data) this type is known.

src/mechanics/cell/Cell.h
joergfunger: From my personal point of view, there should be no passing of a reference at all
```
ReturnObj bla(IntegrateReturnObj(op));
```
Even though this requires to change the name of the function to define the type of the return object, this seems to me a much clearer interface, since the parameter TReturn is neither an input nor an output.

src/mechanics/cell/CellData.h
```diff
@@ -14,26 +14,36 @@ namespace NuTo
...
+            /* The mutable member mNodeValues may be accessed by multiple threads
+             * at once - in an OpenMP context.
+             * Concurrent read access is no problem. This if branch, however,
+             * performs write access. No concurrency here.
+             */
 #ifdef _OPENMP
 #pragma omp critical
 #endif
-            nodeValues = mElements[dofType]->ExtractNodeValues();
+            nodeValues = mElements.DofElement(dofType).ExtractNodeValues();
```
joergfunger: Should we really add an omp critical here. Each cell and thus each cell_data should be accessed only by a single thread, thus this should not be an issue at all. This critical then unnecessarily slows down the parallel performance, since different threads accessing different cells have to wait for each other.

src/mechanics/cell/Cell.h
TTitscher: I am happy to see a working implementation without a second argument. I didnt manage to do that. Maybe some `decltype` magic. Currently the argument has two purposes:
- enable template deduction
- provide a _zero_ value. E.g. double is not initialized to zero by default.

src/mechanics/integrands/MomentumBalance.h
TTitscher: Those were two options. We lose the type information, if we store the history data of the law at the integrand. If we store it at the law, there is no problem. 
One alternative to a MaterialParameters class is to use our current LawWithHistoryData pattern. So the laws would derive from `MechanicsInterfaceWithData`. 
~~~cpp
template <typename TLaw>
struct MechanicsInterfaceWithData : MechanicsInterface
{
    TLaw::HistoryDataType mHistoryData;

    EngStress Stress(Strain, deltaT) override
    {
        return Stress(Strain, mHistoryData, deltaA);
    }

protected:
    virtual EngStress Stress(Strain, TLaw::HistoryDataType&, deltaT) = 0;
    // overridden by actual laws.
}
~~~
Your reference approach seems a lot easier. The usage will not be as nice.
~~~cpp
LawMaterialParameters params(E, nu); // must not go out of scope
Law law(params);
// vs
Law law(E, nu);
~~~
Everything is solveable. I think we get a feel for a nice solution if we start implementing some laws with data.

PR #113, opened 2017-09-29
Rework new Group to chronological ordering

TTitscher:
As I mentioned in #63, I had some problems with the Group based on `UniqueId` and want to remove this dependency. Here it is.

- Unique elements - based on pointer compare
- deterministic ordering - based on order of insertion
- no more public parent iterators (`pcbegin(), pend()`)
     - ... because of the _very_ straight forward and short custom implementation of Unite/Difference/Intersection/SymmetricDifference (that keep the ordering)
- overhead: a second container (std::vector) that contains the entities sorted by pointer value. That one is only used to check for duplicates, not for the actual ordering of the group.


TTitscher:
@Psirus Maybe you can explain why you chose to inherit privately from `std::vector` instead of having one as a member. I quickly implemented the latter version because of simplicity. Composition over inheritance and so on. If there is a good reason for the inheritance, we should obviously revert this commit.

Psirus:
No biggie, just being able to `using vectorFct = parent::vectorFct` (size, insert, etc.) instead of implementing them. Laziness basically.

Psirus (approved):
If it works...personally, I don't find reasoning about when to use `mUniqueData` and when `mData` particularly easy, but any other implementation of the same logic would probably even harder. Also, do we need to write the `Unite`, `Difference`, etc. ourselves?

TTitscher:
The builtin set_algorithms only work on ordered containers. We could order the group via ptr address and use those algorithms. But this would destroy the chronological (deterministic) order. 
Plus, I like the simplicity and shortness of the implementation.

pmueller2 (approved):

PR #114, opened 2017-09-30
Simplify cmake

Psirus:
Start on #98. The parts that are done here have been checked off on the issue page. Would be good to know if this is alright, before tackling the remaining parts.


test/mechanics/dofSubMatrixSolvers/CMakeLists.txt
```diff
@@ -52,7 +52,7 @@ endif()
 #add_unit_test(SpatialContainer)
```
TTitscher: Whats up with these commented pieces of code? Some copy-paste remains?

TTitscher (approved):
Roughly 200 lines of cmake code gone and it still works? I like. Thanks!

test/mechanics/dofSubMatrixSolvers/CMakeLists.txt
Psirus: Was introduced by fe11df171e, I didn't touch it.

PR #116, opened 2017-10-09
Simplify cmake continued

Psirus:
Further work on simplifying and cleaning up our build system. The corresponding issue is #98. The most prominent changes are:
- All "NuTo modules" (Mechanics, Optimize, Metalmodel, etc.) are now always enabled (except for the Python wrappers).
- Unit tests, integration tests and examples don't build "by default" any more. That means if you just say "make", it will only build the library, nothing else. The respective targets are called `unit`, `integrationtests` and `examples`.
- The target for local coverage has been removed. Just use codecov.
- There are now targets for Boost, Mumps, Arpack, Ann, Json and Fakeit. Use those instead of `${XYZ_LIBRARIES}`.
- A new macro for adding integrationtests has been added; `add_integrationtest()`.


applications/integrationtests/CMakeLists.txt
```diff
@@ -1,25 +1,14 @@
...
-target_link_libraries(SparseMatrixTests Math Base)
-add_test(integration::SparseMatrixTests.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/SparseMatrixTests)
-
-add_custom_target(integrationtests COMMAND ctest -R integration)
+add_subdirectory(metamodel)
+add_subdirectory(optimize)
+add_subdirectory(visualize)
+
+add_custom_target(integrationtests
```
TTitscher, line 10: Is it possible/reasonable to collect these target definitions somewhere? /srcipts/Targets.cmake or something?

applications/integrationtests/geometryConcrete/CMakeLists.txt
```diff
@@ -31,11 +31,14 @@ if(ENABLE_PYTHON)
 endif()
 
 if(GMSH_PATH)
-    add_executable(GeometryConcreteAPI GeometryConcreteAPI.cpp)
+    add_executable(GeometryConcreteAPI EXCLUDE_FROM_ALL GeometryConcreteAPI.cpp)
```
TTitscher, line 34: We should document some of the changes and the new features. This `EXCLUDE_FROM_ALL` removes the executable from `make all`, right? 
With those custom targets, all our executables have to be excluded from `all`. Is there a way to define that only once and not at every `add_executable` command?

applications/integrationtests/mechanics/MultipleConstitutiveLaws.cpp
```diff
@@ -869,12 +869,10 @@ void AdditiveInputImplicitTest(std::vector<int> rN, std::vector<double> rL,
 
     SetupMultiProcessor(S);
 
-#ifdef ENABLE_VISUALIZE
```
TTitscher, line 872: Removing that really cleans up the code. Thank you!

applications/integrationtests/mechanics/feti/CMakeLists.txt
```diff
@@ -7,52 +7,45 @@ add_subdirectory(meshes)
...
         target_link_libraries(${test} Boost::mpi ${LAPACK_LIBRARIES})
-        target_link_libraries(${test} ${ANN_LIBRARIES} ${MPI_LIBRARIES})
-        if(ENABLE_VISUALIZE)
-            target_link_libraries(${test} Visualize)
-        endif()
-        if(MUMPS_FOUND)
-            target_link_libraries(${test} ${MUMPS_LIBRARIES})
-            target_link_libraries(${test} ${CMAKE_THREAD_LIBS_INIT})
-        endif()
+        target_link_libraries(${test} ${MPI_LIBRARIES})
```
TTitscher, line 13: Is it advantageous to also convert MPI_LIBRARIES and LAPACK_LIBRARIES (and others...) into cmake targets? Is this manual coding or can we copy that from somewhere?

TTitscher (approved):
Again 600 lines less cmake/build code. Thank you!

applications/integrationtests/CMakeLists.txt
Psirus: I guess it would be possible, but I think it's quite appropriate there. In the `integrationtests` directory, you define your `integrationtests` target. Like you would with the `src/mechanics` and `Mechanics`, no?

applications/integrationtests/geometryConcrete/CMakeLists.txt
Psirus: Yes, `EXCLUDE_FROM_ALL` removes it from the default build target. Afaik there is no way to set this globally. This is also part of the reason why I've introduced the `add_integrationtest` macro.

applications/integrationtests/mechanics/feti/CMakeLists.txt
Psirus: I think it would be advantageous as well. This is mostly guesswork, reading the documentation and reading other FindXYZ modules, e.g. looking at FindPNG from CMake itself.

applications/integrationtests/CMakeLists.txt
```diff
@@ -1,25 +1,14 @@
...
-endif()
-
-if(ENABLE_METAMODEL)
-    add_subdirectory(metamodel)
-endif()
-
-if(ENABLE_OPTIMIZE)
-    add_subdirectory(optimize)
-endif()
+unset(all_integration_tests CACHE)
```
joergfunger, line 1: Why is this necessary? Is all_integration_tests still set in the Cachefile, or is that because of doing that in the oldVersion?

scripts/cmake/CheckForDependencies.cmake
```diff
@@ -0,0 +1,72 @@
+macro(check_for_dependencies)
```
joergfunger, line 1: Those files are just moved from scripts into scripts/cmake, right? Is there a way to see that in the pull request? I was just wondering when going through the list of deleted files, how to distinguish between moved and deleted files.

joergfunger (approved):
Thanks for the tedious work of removing the flags.

applications/integrationtests/CMakeLists.txt
TTitscher: @Psirus  is the expert here, but I'll try. The target `integrationtests` is basically a list of corresponding executables. In this case, this list is the variable `all_integration_tests`. In case you remove, rename or add new tests, this list has to be updated as well. For the target `unit` we are doing this technique for a while and several times I ran into problems. Especially if you delete a test, it will still be in the target list. I had to manually clear the variable in my CMakeCache.txt to fix that. I hope and guess that this `unset` command clears the variable automatically.

Psirus: Yes, unset is just for clearing that variable. Otherwise we would accumulate duplicate tests or try to run deleted tests, as @TTischer said.

scripts/cmake/CheckForDependencies.cmake
Psirus: In one commit they were just moved, yes. In the PR as a whole they were also changed, which is why they are not shown as move only.

PR #117, opened 2017-10-16
Add RungeKutta method independent of structure

pmueller2:
This is a first attempt to separate time Integration methods from structure/PDE related stuff. Time integration is performed by repeatedly calling the `doStep` method. The function f that is used to calculate the right hand side of the differential equation has a signature that meets the requirements of the integrators of Boost Odeint because I think it would be nice if one could exchange the self made integrator with the ones from this library.


src/mechanics/timeIntegration/ExplicitRungeKutta.h
```diff
@@ -0,0 +1,64 @@
...
+        {
+            c.push_back(elm);
+        }
+    }
+
+    //! @brief Performs one RungeKutta step
+    template <typename F>
+    state doStep(F f, state w0, double t0, double h)
+    {
+        std::vector<state> k;
```
TTitscher: `std::vector` can be initialized directly as
~~~c++
std::vector<state> k(c.size(), w0);
~~~
This is the second ctor from here http://en.cppreference.com/w/cpp/container/vector/vector

src/mechanics/timeIntegration/ExplicitRungeKutta.h
```diff
@@ -0,0 +1,64 @@
...
+            k.push_back(state(w0));
+        }
+        state result = w0;
+        state wni = w0;
+        double t;
+
+        for (std::size_t i = 0; i < c.size(); i++)
+        {
+            t = t0 + c[i] * h;
+            wni = w0;
```
TTitscher: I suggest initializing the variables `wni` and `t` in the loop. The compiler might pull it out of the loop for optimization. But generally I would keep the scope of variables as small as possible.

src/mechanics/timeIntegration/ExplicitRungeKutta.h
```diff
@@ -0,0 +1,64 @@
...
+                wni += h * a[i][j] * k[j];
+            }
+            f(wni, k[i], t);
+            result += h * b[i] * k[i];
+        }
+        return result;
+    }
+
+protected:
+    std::vector<std::vector<double>> a;
```
TTitscher, line 55: This really looks like a matrix. Some thoughts on this
- make it fixed size (all a, b and c have the same size (or size squared for a))
- even move it to a ButcherTable class?

really not sure what to do. But this looks strange.

In any case they should be renamed to something starting with `m`, e.g. `mA, mB, mC`

src/mechanics/timeIntegration/ExplicitRungeKutta.h
```diff
@@ -0,0 +1,64 @@
...
+{
+template <typename state>
+class ExplicitRungeKutta
+{
+
+public:
+    //! @brief Initialization with method specific parameters (butcher tableau)
+    ExplicitRungeKutta(std::vector<std::vector<double>> aa, std::vector<double> bb, std::vector<double> cc)
+    {
+        for (auto elm : aa)
```
TTitscher: Can you simply copy the whole vectors?
~~~c++
ExplicitRungeKutta(std::vector<std::vector<double>> aa, std::vector<double> bb, std::vector<double> cc)
  : a(aa)
  , b(bb)
  , c(cc) {}
~~~

The ctor itself should check, whether or not the butcher table is valid for explicit RK schemes. No diagonal entries in aa and no values on the upper triangle, right?

test/mechanics/timeIntegration/RK4.cpp
```diff
@@ -0,0 +1,57 @@
+#include "BoostUnitTest.h"
+
+#include <iostream>
+#include <Eigen/Core>
```
TTitscher: Travis error here. Use `<eigen3/Eigen/Core>` for now. This is IMO an issue in our cmakelists.  @Psirus ??? :)

test/mechanics/timeIntegration/RK4.cpp
```diff
@@ -0,0 +1,57 @@
+#include "BoostUnitTest.h"
+
+#include <iostream>
```
TTitscher: iostream should not be needed in unit tests. You could use a [`BOOST_TEST_MESSAGE`](http://www.boost.org/doc/libs/1_51_0/libs/test/doc/html/utf/user-guide/test-output/BOOST_TEST_MESSAGE.html), if you really need to print something.

test/mechanics/timeIntegration/RK4.cpp
```diff
@@ -0,0 +1,57 @@
...
+    //! first order system
+    void operator()(const Eigen::Vector2d& w, Eigen::Vector2d& dxdt, double t)
+    {
+        dxdt[0] = w[1];
+        dxdt[1] = -w[0] - 2. * m_gam * w[1];
+    }
+
+    //! Solution for damped harmonic oscillator at time t
+    //! Initial conditions [displacement, velocity] w0
+    double exactSolution(const Eigen::Vector2d& w0, double t)
```
TTitscher: Can you provide a link to this exact solution for further reading? Wiki or something. Plus: CamelCase please :)

src/mechanics/timeIntegration/ExplicitRungeKutta.h
```diff
@@ -0,0 +1,64 @@
...
+        }
+        for (auto elm : cc)
+        {
+            c.push_back(elm);
+        }
+    }
+
+    //! @brief Performs one RungeKutta step
+    template <typename F>
+    state doStep(F f, state w0, double t0, double h)
```
TTitscher: Can you provide a brief documentation here? Esp. how `f` has to look like. 

I am still struggeling with the name, `DoStep` would be our convention. But, I guess, you want to have a similar naming as the `boost::odeint` stuff. Still, not sure if we should stick to our naming convention.

TTitscher (changes requested):
Great idea to make the time integration more generic. My request mostly concerns documentation, naming and something like [RAII](http://en.cppreference.com/w/cpp/language/raii). IMO the latter one is very important. Try to initialize everything upon construction and in the smallest scope possible. This also makes the code much shorter.

test/mechanics/timeIntegration/RK4.cpp
Psirus: Yes, imho the Eigen include dir under Debian for example should be `/usr/include/eigen3`, so that you could do the above and be in agreement with the EIgen documentation. We could change that if we want to. Personally, I would like that.

src/mechanics/timeIntegration/ExplicitRungeKutta.h
```diff
@@ -0,0 +1,64 @@
...
+    {
+        std::vector<state> k;
+        for (size_t i = 0; i < c.size(); i++)
+        {
+            k.push_back(state(w0));
```
joergfunger: Is it really necessary to dynamically allocate  the state vectors for each time step?  It would be interesting to see the performance difference, if this allocation and the state vector would be moved to the constructor/member variable , the number of variables should be known there.

src/mechanics/timeIntegration/ExplicitRungeKutta.h
joergfunger: I would also stick to our naming convention. In case use the boost interface, there could be a similar method in the header that calls our method. This should be optimized by the compiler.

joergfunger (changes requested):

src/mechanics/timeIntegration/ExplicitRungeKutta.h
pmueller2: The size of the state vector is not known when the integrator is created. But of course once the integration starts things are different. This will probably only be relevant for large state vectors and I want to postpone optimizations of this kind to later stages. Maybe a function like integrateNsteps() could take care of the initial resizing.

Psirus:
I'm not sure I fully understand. The slightly weird syntax of the `operator()` is to be compatible with Boost. Okay, fine. But then why do this at all and not just "`#include <boost/numeric/odeint/stepper/runge_kutta4.hpp>`, done."?

TTitscher:
You write the RHS of your problem using this operator(). Now you can use all boost::odeint schemes. If there is one missing, you can implement a custom scheme in nuto. Or do you think that we do not need to implement explicit schemes on our own? Might be a valid point as well. Not sure.

TTitscher:
Thought about it again. I think it's nice to have a custom implementation of runge kutta without external dependencies. In contrast to boost::odeint, you can easily recognize the pseudo code from e.g. wiki, with butcher table and so on. Another more subtle, minor advantage: boost::odeint requires the state to be default constructable. This caused a lot of troubles when using our BlockVector. The current implementation does not have this requirement.

src/mechanics/timeIntegration/ExplicitRungeKutta.h
```diff
@@ -0,0 +1,60 @@
...
+    //! The return value is stored in dwdt
+    //!
+    //! @param w0 initial value
+    //! @param t0 start time
+    //! @param h step size (t-t0)
+    //! @return value after one RungeKutta step
+    template <typename F>
+    Tstate DoStep(F f, Tstate w0, double t0, double h)
+    {
+        std::vector<Tstate> k(c.size(), w0);
```
TTitscher: Did some measurements.

performance gain 1: make `c.size()` a constexpr (e.g. template argument `TNumStages`)
performance gain 2: `std::vector` --> `std::array<TState, TNumStages>`
The last one requires that `TState` is default constructable (IMO bad) OR a wierd template magic `make_array<TNumStages>(w0)` that initializes the array with copies of `w0`. 

As long as performance is not important, this ugliness may not be worth it.

TTitscher (approved):

Psirus:
Sorry, I don't want to be so negative, but

> [...] the right hand side of the differential equation has a signature that meets the requirements of the integrators of Boost Odeint because I think it would be nice if one could exchange the self made integrator with the ones from this library.

> Another more subtle, minor advantage: boost::odeint requires the state to be default constructable. This caused a lot of troubles when using our BlockVector. 

So either a) we're accepting a slightly weird syntax to be able to use boost as well or b) we're reimplementing this ourselves because we can't use boost. Which is it?

pmueller2:
Ok, so I guess your proposal if we want to implement an integrator ourselves would then be to use a simpler RightHandSide operator (get rid of the reference to the result dxdt) so that it is just ()(x,t) and wrap it, if we want to use boost right?

Psirus:
I don't have proposal, to be honest I don't have a strong opinion about how this should look. I was just wondering what the motivation was, since you mentioned Odeint yourself and it already has RK4.

joergfunger (approved):

Psirus:
Can we have a test for the `StructureRhsExplicit2ndOrder` thing? This seems the most interesting part to me.

pmueller2:
The only way to test it I can think of is an integration test like WaveEquation1D. There are also still some issues:
1. Currently no time dependent load (maybe not necessary because PDE...)
2. The update of the dependent velocities is not right because it does not include the time derivative of the right hand side of the constraint equation
But I do not know how to design a test that fails because of 2.

Psirus:
👏

Psirus:
Maybe you can modify the `FindEIGEN.cmake` to get back the include path including the `eigen3` part, so that you don't have that problem anymore. I'd prefer it as well, because that's how the Eigen documentation writes it. It would also be nice to Eigen as an imported target :) But preferably in a seperate PR. Feel free to send me a mail/reply to #98 if you want to tackle that and need help.

Issue #34, opened 2017-06-02
Implicit dependency of Interpolation and Intergration


TTitscher:
`eShapeType` defines the shape of an InterpolationType (`TRUSS`, `BRICK`, ...) and the `eTypeOrder` defines the actual interpolation type and its order (`EQUIDISTANT1`, `LOBATTO3`).  Independently, you have to pick integration types basically containing the same information hidden in another enum `eIntegrationType`, e.g.
`eIntegrationType::IntegrationType2D4NGauss4Ip`
- `eIntegrationType::IntegrationType...` (quite redundant, minor problem :) )
- `...2D3D...  --> eShapeType::TRIANGLE`
- `...Gauss4IP --> eTypeOrder::EQUIDISTANT2` (which on top of that is misleading since it is implemented as a [serendipity](https://ses.library.usyd.edu.au/bitstream/2123/709/8/adt-NU20060210.15574814appendixD.pdf) interpolation.

## Problem

- redundant information
- Combination of (e.g.) `LOBATTO` and `TETRAHEDRON` can be defined, but is impossible
    - many unnecessary run time checks
- Which integration type can accurately integrate triangles of order 4? `6IP`, `10IP`, `12IP?`, `15IP`?
    - error prone
- matching interpolation types and integration types have to be defined in the same natural coordinate system

## Solution (idea)

- `Shape` classes that defines the natural coordinate system and its bounds
    - `Quad` - xi, eta in [-1, 1]
    - `Triangle` - xi, eta in triange [0,0] -- [1,0] -- [0,1];
    - ....

- Define the general interpolation / integration concept with an enum containing e.g. `Equidistant`, `Lobatto`, `Serendipity`

- Somehow create the integration types and interpolation types from an integer value containing the `order`

Maybe like:
~~~cpp
auto triangle = NuTo::Shape::Triangle::Create()
element.SetInterpolationType(triangle.Interpolation<Equidistant>(2));
element.SetIntergrationType(triangle.Integration<Equidistant>(2));
//or
element.SetBoth<Equidistant>(triangle, 2);
~~~

## Urgency

Meh. If we find a concept, this could be implemented during the PDE change... :)

## Discussion

Do you see these problems as well? How should the API look like? How should the classes look like?

~~~cpp
// template everything
NuTo::Interpolation<Triangle, Equidistant,1> 
// template some
NuTo::InterpolationTriangle<Equidistant,1>
NuTo::InterpolationTriangleEquidistant<1> 
// some dynamic
[...]

PR #118, opened 2017-10-17
Pde mesh

TTitscher:
Implements a MeshFem that stores (owns) nodes, fem interpolation types and fem element collections. It also implements a `GetNodeAt()` as a primitive (poor performance expected) node selection method. This might be helpful for the constraints-sprint of @joergfunger and @vhirtham .


src/base/ValueVector.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <boost/ptr_container/ptr_vector.hpp>
+
+namespace NuTo
+{
+
+//! @brief container that stores values of T and keeps references to these values valid
```
Psirus, line 8: I'm not sure what this class is supposed to accomplish, compared to say `std::vector`. ~~The iterators of `boost::ptr_vector` are invalidated the same way as those of the `std::vector`, see [boost docs](http://www.boost.org/doc/libs/1_65_1/libs/ptr_container/doc/conventions.html#iterators-are-invalidated-as-in-the-corresponding-standard-container).~~ (**Edit:** Your test disproves this; hm...) Can you maybe add a few more lines of documentation/explanation?

src/mechanics/elements/ElementCollection.h
```diff
@@ -1,39 +1,82 @@
...
+    void AddDofElement(DofType dofType, TElement dofElement)
     {
-        mDofElements[dofType] = &dofElement;
+        mDofElements.Insert(dofType, dofElement);
+        // The alternative implementation with
+        //        mDofElements[dofType] = dofElement
+        // will fail. The operator[] must be able default construct a new TElement, if it does not exist. It would then
+        // return a reference to it and dofElement can be copied/moved into it. Our elements are not default
+        // constructable (may require nodes, interpolations, ...). Thus, use Insert here, that copies/moves the entity
+        // into the DofContainer without a temporary, default constructed TElement.
```
Psirus, line 52: If the compiler will complain about `mDofElements[dofType] = dofElement`, then I'd delete the comment.

src/mechanics/elements/ElementIga.h
```diff
@@ -34,36 +34,42 @@ class ElementIga : public ElementInterface
...
 
-    Eigen::VectorXd GetShapeFunctions(Eigen::VectorXd ipCoords) const override
+    ShapeFunctions GetShapeFunctions(NaturalCoords ipCoords) const override
     {
-        return mNurbsGeometry.BasisFunctionsAndDerivativesRational(0, Transformation(ipCoords));
+        return NurbsGeometry().BasisFunctionsAndDerivativesRational(0, Transformation(ipCoords));
     }
 
-    Eigen::MatrixXd GetDerivativeShapeFunctions(Eigen::VectorXd ipCoords) const override
+    DerivativeShapeFunctionsNatural GetDerivativeShapeFunctions(NaturalCoords ipCoords) const override
```
Psirus, line 50: I'm not so sure about these typedefs, as in I find them confusing. I'd start looking for `NaturalCcords.h` and wonder how the constructor might look like etc. And they are not *strong types* either, so no one is stopping me from passing say `GlobalCoords` to it. But then again `Eigen::VecUpTo3D` would probably just as confusing, so I don't know.

Psirus (approved):
Looks good to me.

src/mechanics/elements/ElementCollection.h
TTitscher: For me, it took like an hour to figure out the exact problem from the compilation error. That is why, I wrote it there....

src/mechanics/elements/ElementIga.h
TTitscher: I see your point. This is more a help to make it more verbose than an actual strong type. Maybe we should add a hint that it is still an Eigen::Matrix<>. There are some special features though. If you look at, say `NaturalCoords`, it is a dynamic vector with max size 3. So the type does matter and I guess this is the reason I had to change the line you marked. Eigen::VectorXd may be implicitly convertable to NaturalCoords, but it is not the same.

src/base/ValueVector.h
TTitscher: I just thought about this. IMO you could compare it to a `vector<unique_ptr>`. Values are stored somewhere in memory and the unique_ptr points to them. And my container returns references to those locations. So references to the `unique_ptr` are invalidated, but the underlying values in memory do not move. Is that understandable somehow? :) If so, I'll add it as a comment!

src/mechanics/mesh/MeshFem.h
```diff
@@ -11,13 +13,63 @@ namespace NuTo
...
 {
 public:
-
     InterpolationSimple& CreateInterpolation(const InterpolationSimple& interpolation)
     {
         mInterpolations.push_back(interpolation.Clone().release());
         return *mInterpolations.rbegin();
     }
 
+    //! @brief selects a node of type \p dofType at given \p coords
```
Psirus: Doxygen understands Markdown. I'd prefer

    selects a node of type `dofType` at given `coords`

src/mechanics/mesh/MeshFem.h
```diff
@@ -0,0 +1,82 @@
...
+                              "There is no node for dof type " + dofType.GetName() + " at " + coordsString.str());
+    }
+
+    //! @brief selects all nodes of type \p dofType where the \p coord in \p direction is within \tol
+    //! @param direction ::X, ::Y, or ::Z
+    //! @param coord coordinate value on the axis
+    //! @param dofType dof type
+    //! @param tol selection tolerance
+    //! @return group with selected nodes, the group may be empty if no nodes were found
+    Groups::Group<NodeSimple> NodeAtAxis(eDirection direction, double coord, DofType dofType, double tol = 1.e-10)
```
Psirus: Small naming thing: I think it should be *Nodes*, because it may return several. Also *Axis* suggests `coord == 0.0` to me, but then again I don't have a good alternative.

src/mechanics/mesh/MeshFem.h
TTitscher: Good point. I actually thought that `\p` is somehow related to `@param` and somehow creates a hyperref from the word to the parameter. If it is just about the font, we could leave it out completely. What do you think?

src/mechanics/mesh/MeshFem.h
TTitscher: Jip, i'll fix the naming. What about changing `coord` to `axisOffset = 0`. So it actually has the meaning of an axis, but you can provide an offset.

src/mechanics/mesh/MeshFem.h
Psirus: Unfortunately, it's just the font. In the generated HTML, the difference is almost unnoticeable. But I'm so used to this syntax that I find it useful in the source to differentiate between "prose" and "code".

src/mechanics/mesh/MeshFem.h
Psirus: 👍

src/mechanics/elements/ElementFem.h
```diff
@@ -12,49 +12,71 @@ namespace NuTo
...
-                         mInterpolation.GetDofDimension());
+        return Matrix::N(Interpolation().GetShapeFunctions(ipCoords), Interpolation().GetNumNodes(),
+                         Interpolation().GetDofDimension());
     }
 
-    Eigen::VectorXd GetShapeFunctions(Eigen::VectorXd ipCoords) const override
+    ShapeFunctions GetShapeFunctions(NaturalCoords ipCoords) const override
     {
-        return mInterpolation.GetShapeFunctions(ipCoords);
+        return Interpolation().GetShapeFunctions(ipCoords);
```
joergfunger, line 50: What is the advantage of using the additional method Interpolation() instead of mInterpolation?

joergfunger (approved):

src/mechanics/mesh/MeshFem.h
```diff
@@ -78,5 +57,25 @@ class MeshFem
     boost::ptr_vector<InterpolationSimple> mInterpolations;
 };
 
+namespace UnitMeshFem
```
Psirus: I think these should live in a separate file, `UnitMeshFem.h` maybe?

src/mechanics/elements/ElementFem.h
TTitscher: I changed `const&` to `reference_wrapper`. The reference wrapper syntax involves a `.get()`, which I found even uglier. Maybe easier to change later. Maybe only preference.

TTitscher:
Just for the record. For now (and for this pull request), I am happy with the functionality of the MeshFem. So (I think) I'll only do new commits, to answer/fix your comments. 

You can
 - Create nodes / elements / interpolations manually
 - Create a 2D unit mesh (from 0,0 to 1,1) for triangles and quads
 - transform the coordinates of a mesh 
   (_maybe TODO: provide some default transformations like translation, streching, rotation, ..._)
 - select dof nodes or coordinate nodes
    - at a given coordinate
    - along a given coordinate axis
 - add new dof interpolations by adding and merging dof nodes at the same coordinate (similar to ConvertToInterpolationType)

Psirus:
Sounds awesome! I think unit meshes for 1D and 3D would be nice as well, but this is low priority.

src/mechanics/interpolation/InterpolationTriangleQuadratic.h
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    std::unique_ptr<InterpolationSimple> Clone() const override
+    {
+        return std::make_unique<InterpolationTriangleQuadratic>(*this);
+    }
+
+    ShapeFunctions GetShapeFunctions(const NaturalCoords& naturalIpCoords) const override
+    {
+        return ShapeFunctions2D::ShapeFunctionsTriangleOrder2(naturalIpCoords);
```
Psirus, line 23: Can we start moving this code here, instead of having this huge `ElementShapeFunctions.cpp`?

TTitscher: I left it as it is since I was not sure how it should look like in an ideal world. And wanted to save some work. 

Maybe
~~~cpp
Triangle<TOrder>
Quad<TOrder, TType>
// with TType being on of
struct Equidistant{};
struct Lobatto{};
struct Serendipity{};
~~~
or "simply"
~~~cpp
Interpolation<TShape, TOrder, TType>
// with a corresponding
Integration<TShape, TOrder, TType>
// where TShape has to match
~~~
(maybe see #34)

Who knows? Not me. (We never lost control... :) )

PR #120, opened 2017-10-20
Hygro

Psirus:
I'd like to merge the stuff I've done for Hygro into master. It's far from done, and I'll probably continue working on this branch. It's just that I don't want to work in parallel forever, and if there is something that would keep this from being merged, I'd rather know sooner than later.

Specifically, I've added stuff quite ugly into the `ContinuumElement` and written a `PorousMediaAdapter` (a fake law so to say), as to circumvent all the `ConstitutiveMaps` and `Evaluate`, in view of upcoming PDE :) (and because it's very, very annoying).

Also, there is a new subdirectory `pybind`, where I've tried [pybind11](https://pybind11.readthedocs.io/en/stable/), an alternative to SWIG. We can discuss whether we want to switch completely later, but for now I find it useful.


pmueller2 (approved):
I think I cannot be of much help here...looks good? You referenced Gawin et al. - Can you provide a citation so that one can look it up (since you found two mistakes...)?
...
there is this ConstitutiveEnum...

Psirus:
Yes, will do. Proper documentation and testing of many of these classes is still on the to-do list ;)

src/mechanics/PDEs/DryAirMassBalance.h
```diff
@@ -0,0 +1,132 @@
...
+//! Corresponds to \f$ \mathbf{K}_{gc} \f$ in Gawin et al.
+Eigen::MatrixXd DiffusionCapillaryPressure(const PoreState& poreState, const PorousMedium& medium,
+                                           const Eigen::MatrixXd& dN)
+{
+    const double gasDensity = poreState.GasDensity;
+
+    const double M_a = Hygro::AirMolarMass;
+    const double M_w = Hygro::WaterMolarMass;
+    const double M_g = poreState.GasMolarMass;
+    const double molarRatio = M_a * M_w / std::pow(M_g, 2);
```
TTitscher, line 118: Could you store the constant molarRatio at `Hygro` as well? would save a lot of lines in this law. If you provide the formula as a comment and write down its result, it could even be constexpr.

src/mechanics/PDEs/DryAirMassBalance.h
```diff
@@ -0,0 +1,132 @@
...
+
+    const double temperature = poreState.Temperature;
+    const double capillaryPressure = poreState.CapillaryPressure;
+    const double gasPressure = poreState.GasPressure;
```
TTitscher, line 122: The "shortened" names are identical to the poreState members. If you rename `poreState` to `ps` or something similarly short, you might not need these variables.

src/mechanics/PDEs/HygroHelpers.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+#include "HygroHelpers.h"
+
+using namespace Hygro;
+
+double Hygro::DensityOfWater(const double temperature)
+{
+    using std::pow;
+    const double t = temperature - 273.15; // celsisus temperature
+    const std::array<double, 6> a = {4.89e-7, -1.65e-9, 1.86e-12, 2.43e-13, -1.6e-15, 3.37e-18};
```
TTitscher, line 13: you could make those arrays `constexpr`. Dunno, if that actually makes things faster.

src/mechanics/PDEs/HygroHelpers.cpp
```diff
@@ -0,0 +1,94 @@
...
+    const std::array<double, 6> b = {1.02e3, -7.74e-1, 8.77e-3, -9.21e-5, 3.35e-7, -4.4e-10};
+    const double fst = b[0] + b[1] * t + b[2] * pow(t, 2) + b[3] * pow(t, 3) + b[4] * pow(t, 4) + b[5] * pow(t, 5);
+    const double snd = a[0] + a[1] * t + a[2] * pow(t, 2) + a[3] * pow(t, 3) + a[4] * pow(t, 4) + a[5] * pow(t, 5);
+    return fst - 1e7 * snd;
+}
+
+
+std::tuple<double, double, double> KelvinEq(const double capillaryPressure, const double temperature)
+{
+    const double R = MolarGasConstant;
```
TTitscher, line 23: You often rename MolarGasConstant to R, WaterMolarMass to M_w and so on. Not only here, but also in you weak form. If that is the notation in the paper, I see no problem in using those names directly. You could add their meaning in the documentation.

src/mechanics/PDEs/HygroHelpers.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+double Hygro::KelvinEquation::dTemperature(const double capillaryPressure, const double temperature)
+{
+    return std::get<2>(KelvinEq(capillaryPressure, temperature));
+}
+
+
+double Hygro::SaturationPressure(const double temperature)
+{
+    const double criticalPointTemperature = 647.096; // Kelvin
```
TTitscher, line 60: The compiler will make an compile time constant out of this, right? You could be sure, if you use `constexpr`

src/mechanics/PDEs/WaterMassBalance.h
```diff
@@ -0,0 +1,129 @@
...
+    return N.transpose() * n * (1 - S_w) * M_w * dpv_dpc / (R * T) * N;
+}
+
+//! Corresponds to first term in \f$ \mathbf{K}_{cg} \f$ in Gawin et al.
+Eigen::MatrixXd DiffusionGasPressure(const PoreState& poreState, const PorousMedium& medium,
+                                     const Eigen::MatrixXd& dN)
+{
+    const double airDensity = poreState.AirDensity;
+
+    const double M_a = Hygro::AirMolarMass;
```
TTitscher: This is what I mentioned above. Maybe `Hygro::M_a` is also clear enough (and short enough) and you can avoid the renaming in a lot of methods.

src/mechanics/constitutive/laws/PorousMediaAdapter.h
```diff
@@ -0,0 +1,71 @@
...
+#include "mechanics/constitutive/ConstitutiveEnum.h"
+
+namespace NuTo
+{
+
+class PorousMediaAdapter : public ConstitutiveBase
+{
+public:
+    PorousMediaAdapter(double a, double b, double c, double d)
+        : mProperLaw(a, b, c, d)
```
TTitscher: xD

TTitscher (approved):
Well, these laws obviously involve a lot of terms and constants. I like the Wrapping of the law. The dynamic_cast in continuum element is a bit against the rules, but it makes everything so much clearer.

What is your experience with pybind11? Any problems that involved _special treatment_ (as we know it from SWIG)?
 
Some questions/suggestions in the comments.

src/mechanics/PDEs/DryAirMassBalance.h
Psirus: Nope, the molar mass of the mixture (`M_g`) depends on the "pore state". Has to be computed at each IP.

TTitscher: Ok, seems like your renaming is confusing :)

src/mechanics/PDEs/HygroHelpers.cpp
Psirus: Well, `WaterMolarMass` is clear in what it represents, even if you don't know the paper. In the formulas, I find that the structure goes missing if you say `longVariable / (anotherLongVariable * thisIsAlsoLong)` compared to `M_w / (R * T)`. Sure I could just name it `R` and add a `// Molar gas constant` everywhere, but then I've not gained much.

TTitscher: I see, makes sense. I thought it is enough if you add `// molar gas constant` once at the definition of `R`. Was just an idea.

src/mechanics/PDEs/HygroHelpers.cpp
Psirus: `const`:
~~~
|                                          |        [ns]       |     runs    | std dev[ns] |
|------------------------------------------|-------------------|-------------|-------------|
| Saturation::CompleteRun                  |               285 |     1000000 |          61 |
~~~

`constexpr`:
~~~
|                                          |        [ns]       |     runs    | std dev[ns] |
|------------------------------------------|-------------------|-------------|-------------|
| Saturation::CompleteRun                  |               289 |     1000000 |         151 |
~~~

Compilers aren't that daft.

src/mechanics/PDEs/HygroHelpers.cpp
Psirus: Nope, see below.

Psirus:
I will probably rave on about pybind elsewhere as well, but just quickly:
- proper CMake integration; it's just another library target, no intermediate output, automatic dependency tracking
- it's all C++; no more guessing what happens to these `Module.i` file, because **I did not understand...**
- compile time…hard to judge right now, because I've only exported a couple of classes, but seems quick
- docstrings! 🕺

src/mechanics/PDEs/DryAirMassBalance.h
Psirus: Hm…

Psirus:
Thanks for the reviews, but a) I'm not sure that this is the model I will use, b) I don't really want to do it in old NuTo anyway, c) I don't see myself merging this.

PR #121, opened 2017-10-20
Pde patch test (wooohooo)

TTitscher:
Patch test using
- linear quad elements
- linear elastic material
- manual dof numbering
- neumann boundary conditions (already merged `PDE_newmann_bc`, sorry...)

Problems:
- `AddDofInterpolation()` API may be wrong. If you have a mesh with different interpolations, as we have here, you cannot simply call it. Here we
    1) add volume coordinate elements
    2) add a layer of displacement elements
    3) manually select the boundary coordinate nodes
    4) manually select the boundary displacement nodes
    5) manually add a boundary element.

Suggestions?


test/mechanics/cell/CMakeLists.txt
```diff
@@ -14,3 +14,13 @@ add_unit_test(Matrix)
 
 add_unit_test(CellIpData
     mechanics/elements/ElementShapeFunctions.cpp)
+
+
+add_unit_test(PatchTest
+    mechanics/integrationtypes/IntegrationTypeBase.cpp
+    mechanics/integrationtypes/IntegrationTypeTensorProduct.cpp
+    math/Legendre.cpp # oh WHY? I do gauss... #SomeoneTODO
```
Psirus: The "standard" Gauss is more correctly called Gauss-Legendre quadrature, and your integration points are the roots of Legendre polynomials, implemented in `math/Legendre.cpp`. So you need this here. #NobodysTodo.

test/mechanics/cell/PatchTest.cpp
```diff
@@ -0,0 +1,196 @@
...
+     * /| |     -7------------6   | -->
+     * /| | e5  /     e4      |   | -->
+     * /| |    /              |e1 | -->
+     * /| |   /    _____------5   | -->
+     * /| |  4-----            \  | --> p
+     * /| | /      e0           \ | -->
+     * /| |/                     \| -->
+     *    0-----------------------1
+     *   /_\
+     *   ///
```
Psirus, line 40: I've seen better diagrams from you :P

Psirus (approved):

Psirus:
The PatchTest is really not a unit test anymore, but I think you know that. And I'm not sure what you're trying to tell me with the `AddDofInterpolation` comment above.

TTitscher:
Indeed, that is an integration test - like Cell.cpp or maybe even NeumannBc.cpp. But for now, I do not want to activate the integration tests. Maybe I can make a new target, make integrationPde.

Issue #57, opened 2017-06-18
Simplify BlockVector/matrix #2


TTitscher:
The `Block` types implement arithmetic operators without any state of the art expression template lazy evaluation. If we derive the `BlockVector` directly from `Eigen::VectorXd`, we only need to provide operators to access a certain [`.segment`](https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#a6212eee4095a0ab9ac3a2202274ea8d7) for a specific dof type. This may involve some index magic but sounds feasible.

Maybe we can even benefit from the lazy evaluations provided by `Eigen::SparseMatrix`. [But the block operations are limited.](https://eigen.tuxfamily.org/dox/group__TutorialSparse.html#title7) Read access is no problem though.

joergfunger:
I'm not sure if these lazy evaluation optimization really improves the speed, due to our usage of the vectors/matrix. However, I agree that writing our own interface for matrix/vector multiplications is sub optimal. However, I don't see a nice solution in particular for the sparse matrices. Even when we switch to `Eigen::SparseMatrix`, I would guess that accessing this via block routines is suboptimal (after having checked the documentation, I'm not sure if there are block methods available for sparse matrices anyway)

Issue #42, opened 2017-06-08
Benchmark lib


TTitscher:
We currently use a handwritten benchmark class. I used it as a training to implement self registered methods that do not require a `main()` method in the \*.cpp file. [Have a look, if you're interested.](https://github.com/nutofem/nuto/blob/master/applications/benchmark/Benchmark.h)

But it has obvious limitations:

1) Handwritten. By an amateur. Pure intuition. No theoretical benchmark background.
2) Kind of unstable results. High inconsitant standard deviation
3) Suffers (maybe, as I said, not an expert) from undesired compiler optimizations
4) Missing features, most important IMO to only run a subset of benchmarks from the CLI without recompiling

Here is a nice overview of popular libs:
http://www.bfilipek.com/2016/01/micro-benchmarking-libraries-for-c.html

Any opinions? [Google Benchmark](https://github.com/google/benchmark) maybe?

Psirus:
I'd vote for Google benchmark, simply because it seems the most widely used. The output of the library should be compatible with the [Jenkins Performance Plugin](https://wiki.jenkins.io/display/JENKINS/Performance+Plugin), so that we can track performance over time. A proof of concept can be found [here](https://webentw1.bam.de:18444/job/nuto_performance/lastBuild/performance/). It seems all of the mentioned libs it the article provide some sort of output, but I'm not sure which is compatible. Sadly, none of them seem to be in the Debian repos.

TTitscher:
Since the number of benchmarks increases, we should really find a way to include it in our CI. Do we need another server for that? Or just run it on sv2212 and hope that one core is idle?

TTitscher:
Alternative lib: https://github.com/ivafanas/sltbench/blob/master/README.md

Psirus:
I've just noticed that the boost test library also produces JUnit output (compatible with Jenkins). Just run your test with `myUnitTest -f JUNIT`. So if we just want to run regression testing on, say, all integration tests, we wouldn't need another library. Of course this is not suitable for microbenchmarking, standard deviations etc, but just fine for automated regression testing.

vhirtham:
Just made a new branch ([googleBenchmark](https://github.com/nutofem/nuto/tree/googleBenchmark)) that implements google benchmark via git submodules. I also wrote some setup scripts, because submodules are not cloned out of the box. The library is compiled together with the nuto libs.

I added a benchmark that uses this lib as example (took one from the documentation) .
[...]

PR #123, opened 2017-10-22
Pde block refactoring

TTitscher:
__TL/DR__: Activate unit tests for `mechanics/dofs/` and introduce `DofCalcContainer` as uniform interface for all dof containers for arithmetic.

Small refactoring of DofVector/DofMatrixContainer.

- activate unit tests for directory dofs
- introduce `DofCalcContainer` that provides + and *(scalar) operators
    - `std::vector` as underlying data structure
    - use/derive `DofVector/DofMatrix/DofSparseMatrix` from that (avoid code duplication)
- add benchmark for various dof vector implementations
    - current version (`std::vector`) is 3x or so faster than previous `std::map`
    - performance does not matter as soon as some more expensive calculations are involved.

---

The actual goal was to solve issue #57, but this will not work as intended. We can build a `DofVector` that inherits from `Eigen::VectorXd`. The idea is to use the arithmetic operators from eigen (including lazy evaluation). The access to the dof types (segments of the vector) requires some additional members that keep track of indexes. We could easily write things like
~~~
Eigen::VectorXd v = dofVector0 + dofVector1 * detJ;
~~~
that are lazy evaluated, super fast, and so on. But we actually want to have
~~~
DofVector v = dofVector0 + dofVector1 * detJ;
~~~
to have access to the single segments/subvectors of v. This requires specific, manually implemented operators like `operator*(DofVector, double)` and we win nothing. 

Another option would be to have an external indexing object in the fashion of
~~~
DofIndex index(someInfoOnTheDofTypesAndTheirSizes);
Eigen::VectorXd v;
intex.GetSubVector(v, dofType) = B^T * sigma;
~~~
But then, we have to pass something like a pair of `index` object and vector. Since performance does not matter much, this is only provides ugly code.


applications/benchmark/CMakeLists.txt
```diff
@@ -38,4 +38,11 @@ target_include_directories(NewtonRaphsonBenchmark
     PUBLIC ${CMAKE_SOURCE_DIR}/src)
 append_to_benchmarks(NewtonRaphsonBenchmark)
 
+add_executable(BlockVectors
+    EXCLUDE_FROM_ALL BlockVectors.cpp)
+target_include_directories(BlockVectors
+    PUBLIC ${CMAKE_SOURCE_DIR}/src)
```
Psirus: Can be made `PRIVATE`. Not that it matters much, but `PUBLIC` means "this target needs these includes to compile, and targets that link to it do as well", and since you're probably not linking this executable to anything else, `PRIVATE` ("this target needs these includes to compile") is enough. Not terribly important, just FYI.

src/mechanics/dofs/DofCalcContainer.h
```diff
@@ -0,0 +1,115 @@
...
+#include <eigen3/Eigen/Core>
+#include "mechanics/dofs/DofType.h"
+
+namespace NuTo
+{
+
+//! @brief dof container that is also capable of performing calculations.
+//! @tparam T type of the data that supports all the arithmetic operations
+template <typename T>
+class DofCalcContainer
```
Psirus, line 13: If I understand this correctly, we now have `DofContainer` for things like nodes and elements, and `DofCalcContainer` for vectors and matrices, right? I find the names `DofMatrix` and `DofVector` are fine, but the base class…
> dof container that is also capable of performing calculations

sounds like it stores `DofType`s, and the name does too. I can't come up with a better name at the moment (`DofLinAlgBase`^^); I'll get back to you when I do.

src/mechanics/dofs/DofCalcContainer.h
```diff
@@ -0,0 +1,115 @@
...
+    const T& operator[](const DofType& dofType) const
+    {
+        return mData.at(dofType.Id());
+    }
+
+    //! @brief nonconst access
+    //! @param dofType dof type
+    //! @return reference to either
+    //          an existing value
+    //          an newly default constructed value
```
Psirus: You need `//!` here for these lines to show up in doxygen.

Psirus (approved):
Nice, there seems to be less code duplication 👍 

This not particularly tied to this PR, but I've noticed it: there are more and more benchmarks, without any explaination what they are benchmarking or why (the same is in some sense also true of many of the integration tests). Maybe you (or whoever wrote the benchmark - so you :P) could add a couple of comments to each of them and tell us what was benchmarked and why, and maybe even a quick overview of the results at the time. No need to write a lot, and especially no need to do it now, just to keep in mind.

applications/benchmark/CMakeLists.txt
TTitscher: Good point, thanks. I just copied it from somewhere without understanding it.

src/mechanics/dofs/DofCalcContainer.h
TTitscher: I have no idea either and thanks for pointing out. Since this is not the only case of unsuitable names, I created #124.

TTitscher:
Thanks for your comments! 
I have the feeling that no one is really interested in these benchmarks and I mainly do them for me. And we always say that performance doesn't really matter, so most of them are pointless. (But I cannot help myself, sorry, in the end, this is c++...) So I do not invest much effort into them.  But sometimes they are helpful to find out if adding adding a million items into a badly designed container will exceed the age of the universe.

What I realized though is that we actually need a real benchmark lib. You may want to profile a single benchmark that you select via CLI - not possible. More and more benchmark concern algorithms. Questions like is _Is Group::Add(...) O(n) or O(log n)? Or is  AddDofInterpolation O(n) or O(n*n)?_ To avoid code duplication, this may require additional arguments for the benchmark function. And so on... 
This requires major modifications in the benchmarks and demotivates me even more to put comments in there.

__Edit:__ Oh, there already is an issue on that #42.

PR #125, opened 2017-10-24
Pde refactor convert

TTitscher:
This performs DofConvert without the ANN library. ANN based on binary trees or other kinds of spatial trees are _very_ useful when their entries are sparse and clustered. A much simpler alternative, a 3D uniform grid structure, is not suitable in those cases. It produces either a lot of memory overhead for empty grid cells or many entries in the grid cells makes ANN search slow. 

In our case, however, we have nice FE elements. Even with local mesh refinement, the element sizes vary in one or two orders of magnitude. For these cases, the 3D uniform grid structure is applicable. Advantages:
- no dependency on libann
- much nicer interface 
  (ann libs often have C style API because they are either written in C or they support ancient C-interfaces of other libs. This may not be 100%, but I got the impression, when I looked at popular libs.)
- short and simple
- faster 
   something like O(1) for addition and access (if the number of grid cells scales with the number of entries.) - in contrast to O(log n) in a tree.


src/mechanics/mesh/UnitMeshFem.cpp
```diff
@@ -60,6 +60,8 @@ MeshFem UnitMeshFem::CreateQuads(int numX, int numY)
 
 void UnitMeshFem::Transform(MeshFem* rMesh, std::function<Eigen::VectorXd(Eigen::VectorXd)> f)
 {
-    auto nodes = rMesh->NodesTotal();
+    // Build a group to avoid duplicates. Otherwise, the transformation is applied multiple times.
+    // This is, however, a bit of a bottleneck for big meshes. 
+    Groups::Group<NodeSimple> nodes = rMesh->NodesTotal();
```
Psirus: I don't fully understand. What is the difference between `mesh.Nodes` and `mesh.NodesTotal()`? What *duplicates* would show up *otherwise*?

applications/benchmark/MeshFemBenchmark.cpp
```diff
@@ -0,0 +1,73 @@
...
+    NuTo::MeshFem mesh = NuTo::UnitMeshFem::CreateTriangles(100, 100);
+    while (runner.KeepRunningTime(1))
+        NuTo::UnitMeshFem::Transform(&mesh, f);
+}
+
+BENCHMARK(Mesh, Transform1000x1000, runner)
+{
+    NuTo::MeshFem mesh = NuTo::UnitMeshFem::CreateTriangles(1000, 1000);
+    while (runner.KeepRunningTime(1))
+        NuTo::UnitMeshFem::Transform(&mesh, f);
```
Psirus: I'm not sure about this syntax, I still find it kind of clunky. If `Transform` has to be a free function, the ideal *syntax* to me would be `mesh2 = Transform(mesh, f)`, but I can see how that would not be ideal either. In this case, I'd vote to make this a member function, and I'd not be surprised that `mesh.Transform(f)` changes the mesh.

src/mechanics/mesh/MeshFemDofConvert.cpp
```diff
@@ -1,96 +1,176 @@
...
+                        continue;
+                    if (kk < 0 || kk >= mDomain.subs[2])
+                        continue;
+
+                    // search the subbox (ii,jj,kk) for matches
+                    for (T& entry : mData[Index(ii, jj, kk)])
+                        if ((entry - coords).squaredNorm() < mDomain.searchRadiusSquared)
+                            return &entry;
+                }
+        return nullptr; // this is a classic case for the c++17 feature std::optional<T>. Here we have to introduce
```
Psirus: Does it make sense to collect these somewhere? Or add `// TODO17`?

src/mechanics/mesh/MeshFemDofConvert.cpp
```diff
@@ -1,96 +1,176 @@
...
-    auto dummyNodes = GetNewDummyNodes(*rMesh, interpolation);
-    NuTo::SpatialContainer<DummyNode, DummyNodeCoordinate> dummyNodesSpatial(dummyNodes);
-
-    // Plan:
-    // If those dummynodes have _NO_ existingNodePtr, a new node
-    // is added to the rMesh->Nodes container. A reference to rMesh newly created node
-    // is saved at all of those duplicate dummy nodes.
+    // Setup subbox. These argument values are quite arbibrary and should maybe be chosen based on 
+    // the dimensions of the mesh.
+    SubBoxes<NodePoint> subBoxes(SetupSubBoxDomain(*rMesh, /*numBoxesPerDirection=*/300, /*eps=*/1.e-10));
```
Psirus: Maybe we should open a new issue to benchmark this and find a mapping between the mesh & element size to these values?

Psirus (approved):
Me like! The subbox algorithm is surprisingly understandable.

In the future though, I would still like to have a benchmark for different meshes, i.e. what happens for large meshes, what happens for meshes with large discrepancies in element size, what happens with meshes with high "porosity" (for example a discrete geometry of bone tissue).

src/mechanics/mesh/UnitMeshFem.cpp
TTitscher: Definitely a point for the renaming issue. `NodesTotal()` returns all (unique - because of the Group) coordinate nodes. In contrast to `NodesTotal(dofType)` that returns all unique nodes for `dofType`.

I actually think of a different approach for the node selection. Maybe only those four method:
~~~
Group<NodeSimple> Mesh::SelectDofNodes(DofType dof, std::function<bool(Eigen::VectorXd)> coordinateSelect);
Group<const NodeSimple> Mesh::SelectDofNodes(DofType dof, std::function<bool(Eigen::VectorXd)> coordinateSelect) const;

Group<NodeSimple> Mesh::SelectCoordinateNodes(std::function<bool(Eigen::VectorXd)> coordinateSelect);
Group<const NodeSimple> Mesh::SelectCoordinateNodes(std::function<bool(Eigen::VectorXd)> coordinateSelect) const;
~~~

And then provide a variety of `coordinateSelect` functions in some helper file. 

__Edit:__ But another pull request IMO. Sry for mixing stuff.

Issue #109, opened 2017-09-26
Manual constraints


TTitscher:
This is actually the issue of @joergfunger, I'll just write it down :ghost:

## Context

We currently allow arbitrary constrains resulting in a system of constraint equations like
~~~
[T1] (d1) + [T2](d2) = (RHS(t))
~~~
d1 are independent dofs, d2 are dependent dofs and T1 and T2 are coefficient matrices.

We then invert `T2` resulting in
~~~
d2 = -[T2^-1 T1] d1 + [T2^-1] (RHS(t))
~~~
This allows removing d2 from the global system of equations and we only have to solve for the d1 dofs. The matrix `[T2^-1 T1]` is called [mConstraintMatrix](https://github.com/nutofem/nuto/blob/master/src/mechanics/structures/Assembler.h#L96) and the `[T2^-1]` is called [mConstraintMappingRhs](https://github.com/nutofem/nuto/blob/master/src/mechanics/structures/Assembler.h#L114).

## Problem

- The total dofs `d` are separated into `d1` and `d2` via a manually implemented gauss implementation. This works in our custom matrix format and is hard to apply to arbitrary formats.
- The `[T2]` is sparse and the calculation of the inverse is not trivial
- Each evaluation of the `RHS(t)`, requires the matrix multiplication `[T2^-1] (RHS(t))`

## Solution

We restrict the way, constraints can be created in the following way:

1) Each constraint equation must only contain _exactly one_ dependent dof. If you need to define a constraint equation with more than one dependent dof, you have to perform an elimination manually. 
So the following would be invalid:
~~~
1) u0 - u1 = 42  [+ u0 is dependent]
2) u1 - u2 = 12  [+ u1 is dependent] 
~~~
This is invalid since `u0` _and_ `u1` are dependent and both are present in the first equation. Instead, you'll have to manually eliminate one of them in your code, e.g.
~~~
1) u0 - u2 = 54  [+ u0 is dependent]
2) u1 - u2 = 12  [+ u1 is dependent] 
~~~

2) The coefficient in front of the dependent dof term has to be `1`. 

As a result of both restrictions, it is totally clear whether or not a dof is in `d1` or `d2` and `[T2]` will be an identity matrix. Thus we can completely avoid all the [problems](#problem).

Psirus:
Solved in PDE.

PR #126, opened 2017-10-24
Pde constraints

TTitscher:
Adds constraints to PDE with the approach of #109. The features below are still not ready, but I am happy to see some early feedback.

- [x] proper documentation including all that CMat, T1, T2, identity stuff
- [x] `ConstraintCompanion` for PDE
- [x] Test case with a nonempty constraint matrix. E.g. periodic boundary conditions
- [x] Check for valid constrains on creation rather than on DofNumbering
- [x] check above may be expensive. Add benchmark to evaluate cost.


src/mechanics/constraintsPde/ConstraintCompanion.cpp
```diff
@@ -0,0 +1,106 @@
...
+            throw Exception(__PRETTY_FUNCTION__, "Dimension mismatch");
+
+        eqs.push_back(Equation({Term(node, component, 1)}, rhs));
+    }
+    return eqs;
+}
+
+
+std::vector<Equation> Component(const Group<NodeBase>& nodes, std::vector<eDirection> directions, double value)
+{
```
pmueller2: I think you will use the new Group here, yes?

src/mechanics/dofs/DofNumbering.cpp
```diff
@@ -0,0 +1,95 @@
...
+                    __PRETTY_FUNCTION__,
+                    "Dof number " + std::to_string(dependentDofNumber) +
+                            " already constrained. You are only allowed to have a dependent dof in one equation.");
+
+        // add the equation index to the status.
+        dofStatus[dependentDofNumber] = iEquation;
+    }
+
+    // go through all terms that are not dependent and check if they contain dependent dofs
+    for (int iEquation = 0; iEquation < constraints.GetNumEquations(dof); ++iEquation)
```
pmueller2: Would it be nice to have such a check when a constraint is created or a term is added?

src/mechanics/constraintsPde/ConstraintCompanion.cpp
TTitscher: The ConstraintCompanion is a point on the bucket list above. Will fix that soon.

src/mechanics/dofs/DofNumbering.cpp
TTitscher: That is a _very_ good point that bothers me as well. I add it to the bullet list.

This method is called when all dofs have a unique, arbitrary number. So we can do all these checks based on this number. 
If we want to do it earlier, and it _definitely makes perfect sense_, we do not have a unique dof numbering. Thus, we have do deal with something like ´std::pair<Node, component>` do uniquely identify the dof. I hope @vhirtham and I can think of a clean solution.

src/mechanics/constraintsPde/ConstraintCompanion.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+Equation Direction(const NodeSimple& node, Eigen::VectorXd direction, RhsFunction rhs)
+{
+
+    assert(direction != Eigen::VectorXd::Zero(direction.rows()) && "the direction vector contains only zeros");
+
+    // Normalization necessary, otherwise rhs depends on length of the direction vector
+    direction.normalize();
+
+    int firstNonZeroComp = -1;
```
TTitscher: I would prefer using `direction.cwiseAbs().maxCoeff(&maxComponent)` instead of this loop/if thing. First, it is shorter. Second, it is more reasonable. In the gauss elimination we had before, we also chose the highest coefficient in the equation. 
In the loop below you additionally check (`iComponent != maxComponent`).

src/mechanics/constraintsPde/ConstraintCompanion.cpp
```diff
@@ -0,0 +1,125 @@
...
+        eqs.insert(eqs.end(), tmpEqs.begin(), tmpEqs.end());
+    }
+    return eqs;
+}
+
+
+Equation Direction(const NodeSimple& node, Eigen::VectorXd direction, RhsFunction rhs)
+{
+
+    assert(direction != Eigen::VectorXd::Zero(direction.rows()) && "the direction vector contains only zeros");
```
TTitscher: It is always a bit risky to compare floating point numbers without a tolerance. If you change that assert into an actual exception, you could properly test it.

src/mechanics/constraintsPde/ConstraintCompanion.cpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+std::vector<Equation> Direction(const Groups::Group<NodeSimple>& nodes, Eigen::VectorXd direction, double value)
+{
+    return Direction(nodes, direction, RhsConstant(value));
+}
+
+Equation Value(const NodeSimple& node, double value)
+{
+    return Value(node, [=](double) { return value; });
```
TTitscher: RhsConstant()?

src/mechanics/constraintsPde/ConstraintCompanion.cpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+Equation Value(const NodeSimple& node, RhsFunction rhs)
+{
+    assert(node.GetNumValues() == 1 && "This function is ment to be used with single value nodes only");
```
TTitscher: Can we change that into an Exception and test it? IMO exceptions should be preferred over assertions for user input.

test/mechanics/constraintsPde/ConstraintCompanionPde.cpp
```diff
@@ -0,0 +1,195 @@
...
+    if (a.GetTerms() != b.GetTerms())
+        return false;
+    return true;
+}
+
+} // Constraint
+} // NuTo
+
+
+// quite the stupid tests, but oh well...
```
TTitscher: ??

test/mechanics/constraintsPde/ConstraintCompanionPde.cpp
```diff
@@ -0,0 +1,195 @@
...
+        expectedEquations.emplace_back(node, 0, sinFunction);
+    }
+    BOOST_CHECK(eqs == expectedEquations);
+}
+
+
+BOOST_FIXTURE_TEST_CASE(Direction_test, Helpers)
+{
+    Eigen::Vector3d someDirection;
+    someDirection << 1, 1, 1;
```
TTitscher: I find this direction test hard to follow. Can you change the direction to something more unique, say `(42, -6174, 0`)?. Or give them names. Then I could follow the single terms in the `expectedEquations` below.

A major part of the `ConstraintCompanion::Direction()` is about finding the first nonzero entry. And not constraining zero entries. These are obviously edge cases you thought about and, thus, they should be tested.

test/mechanics/constraintsPde/ConstraintCompanionPde.cpp
```diff
@@ -0,0 +1,195 @@
...
+    return true;
+}
+
+bool operator==(const Equation& a, const Equation& b)
+{
+    if (a.GetRhs(0.0) != b.GetRhs(0.0))
+        return false;
+    if (a.GetRhs(42.0) != b.GetRhs(42.0))
+        return false;
+    if (a.GetTerms() != b.GetTerms())
```
TTitscher: This checks the order of terms, right? Do we want that? The first term somehow determines the dependent dof and is important. The order of the other ones is IMO not important.

TTitscher (commented):
Thanks for taking the precious ConstraintCompanion to Pde! Apart from minor things I commented on, I would like to see:

- Exceptions instead of assertions including tests for them (or proper reasons why not use Exceptions)
- Better testing of the `Direction` function. The code indicates several pitfalls concerning zero entries in the direction. This should be checked in the tests as well.

src/mechanics/constraintsPde/ConstraintCompanion.cpp
vhirtham: Will change it. However, you can't destroy anything by using this function on a 2+ value node, that is the reason why I picked an assertion. :) 

**DONE**

src/mechanics/constraintsPde/ConstraintCompanion.cpp
vhirtham: If you prefer this name. But I think the fact that the function is meant for single value nodes is more important for the user. That should be addressed in the name as well.

src/mechanics/constraintsPde/ConstraintCompanion.cpp
vhirtham: Yep I know. However, since the vector is normalized even the tiniest number should be enough to get everything back on track. But I ll replace that with a tolerance. 
Something like: max(abs(v)) > 1e-6;  

**DONE**

src/mechanics/constraintsPde/ConstraintCompanion.cpp
vhirtham: As we discussed before, this is an option too and I ll just do that. 


**DONE**

src/mechanics/constraintsPde/ConstraintCompanion.cpp
TTitscher: Not sure if we are talking about the same. I thought of `return Value(node, RhsConstant(value));` :)

test/mechanics/constraintsPde/ConstraintCompanionPde.cpp
vhirtham: The order is not important, but this is just a helper function for the test (I just took it from the original testfile). I mean, if somebody changes the internals of the companion so that the current order is messed up, the test will probably fail, even though might not. Other option would be to compare every term of A to every other term of B etc.

test/mechanics/constraintsPde/ConstraintCompanionPde.cpp
vhirtham: --> git blame in the original testfile 
Just copied, didn't dare to do much more than making it run again ;p

test/mechanics/constraintsPde/ConstraintCompanionPde.cpp
vhirtham: As I said above, goal was to get the original test to run again. This is absolutely right, but at the moment I hear so much about "stupid" (have a look above) tests, that I am not sure what is considered worth of being tested and what not :laughing:

src/mechanics/constraintsPde/ConstraintCompanion.cpp
vhirtham: k... didn't get what you mean from your first comment xD

Will change that

**DONE**

test/mechanics/constraintsPde/ConstraintCompanionPde.cpp
TTitscher: test everything! :)
Tests are a nice opportunity to make the behavior of the code clear. Imagine a test like
~~~
BOOST_CHECK(Direction(node, Eigen::Vector3d(1,2,3), RhsConstant(5)) == Direction(node, Eigen::Vector3d(3,6,9), RhsConstant(5));
~~~
That would perfectly document the expected behavior.

src/mechanics/constraintsPde/ConstraintCompanion.cpp
```diff
@@ -48,29 +50,25 @@ std::vector<Equation> Component(const Groups::Group<NodeSimple>& nodes, std::vec
 Equation Direction(const NodeSimple& node, Eigen::VectorXd direction, RhsFunction rhs)
 {
 
-    assert(direction != Eigen::VectorXd::Zero(direction.rows()) && "the direction vector contains only zeros");
+    int maxComponentIndex = -1;
+    if (direction.cwiseAbs().maxCoeff(&maxComponentIndex) < 1.e-6)
+        throw Exception(__PRETTY_FUNCTION__,
+                        "Your direction vector is composed of zeros only! The direction is unspecified!");
+
+    assert(maxComponentIndex > -1 && maxComponentIndex < direction.rows());
```
TTitscher: You don't trust `Eigen`, right? :P

src/mechanics/constraintsPde/ConstraintCompanion.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+Equation Value(const NodeSimple& node, double value)
+{
+    return Value(node, RhsConstant(value));
+}
+
+Equation Value(const NodeSimple& node, RhsFunction rhs)
+{
+    if (node.GetNumValues() != 1)
+        throw Exception(__PRETTY_FUNCTION__, "This function is ment to be used with single value nodes only");
```
Psirus: Should be "meant to be".

src/mechanics/constraintsPde/Constraints.cpp
```diff
@@ -0,0 +1,124 @@
...
+                                             "is already constrained as a dependent dof in another equation.");
+
+    if (Contains(mOtherTerms, newDependentTerm))
+        throw Exception(__PRETTY_FUNCTION__, "The dependent dof of the new equation "
+                                             "is already constrained in another equation.");
+
+
+    // Any term in the new equation must not constrain the same dof as any existing _dependent_ terms
+    // since the dependent term of the new equation is already checked above, we omit it here and start loop at 1.
+    for (int iNewTerm = 1; iNewTerm < e.GetTerms().size(); ++iNewTerm)
```
Psirus, line 113: > warning: comparison between signed and unsigned integer expressions

src/mechanics/constraintsPde/Constraints.cpp
```diff
@@ -0,0 +1,124 @@
...
+    Eigen::SparseMatrix<double> matrix(numEquations, numIndependentDofs);
+
+    for (int iEquation = 0; iEquation < numEquations; ++iEquation)
+    {
+        const auto& equation = equations[iEquation];
+        for (Term term : equation.GetTerms())
+        {
+            double coefficient = term.GetCoefficient();
+            int globalDofNumber = term.GetConstrainedDofNumber();
+            if (globalDofNumber == -1 /* should be NodeSimple::NOT_SET */)
```
Psirus, line 50: Is there an issue or something tracking this `NodeSimple::NOT_SET` "problem"?

src/mechanics/constraintsPde/Constraints.cpp
```diff
@@ -0,0 +1,124 @@
...
+                                "There is no dof numbering for a node in equation" + std::to_string(iEquation) + ".");
+
+            if (globalDofNumber >= numIndependentDofs)
+            {
+                // This corresponds to the last block of size numDependentDofs x numDependentDofs that should be an
+                // identity matrix
+                int transformedDof = globalDofNumber - numIndependentDofs;
+                if (transformedDof != iEquation)
+                    throw Exception(__PRETTY_FUNCTION__, "The numbering of the dependent dofs "
+                                                         "is not in accordance to the equation numbering.");
```
Psirus: Should be "in accordance with ...".

src/mechanics/constraintsPde/Constraints.h
```diff
@@ -0,0 +1,83 @@
...
+    //! @return constraint rhs vector
+    Eigen::VectorXd GetRhs(DofType dof, double time) const;
+
+    //! @brief builds a sparse matrix containing the constraint terms for a specific dof type
+    //! @param dof dof type
+    //! @param nIndependentDofs number of independent dofs for the dof type dof, required for a proper resize
+    //!        of the sparse matrix
+    //! @return sparse matrix containing the constraint terms where the last block with size (numDependent x
+    //! numDependent is removed)
+    Eigen::SparseMatrix<double> BuildConstraintMatrix(DofType dof, int numIndependentDofs) const;
```
Psirus: Name of the parameter `numIndependentDofs` is different in code and doxygen comment.

src/mechanics/constraintsPde/Constraints.h
```diff
@@ -0,0 +1,83 @@
...
+
+    private:
+        struct TermCompare
+        {
+            bool operator()(const Term& lhs, const Term& rhs) const;
+        };
+
+        std::set<Term, TermCompare> mDependentTerms;
+        std::set<Term, TermCompare> mOtherTerms;
+    } mTermChecker;
```
Psirus, line 78: Hah, I did not know that one could do this. Cool.

src/mechanics/constraintsPde/Equation.h
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    int GetDependentDofNumber() const
+    {
+        return mTerms.front().GetConstrainedDofNumber();
+    }
+
+    //! @brief nonconst getter for mTerms
+    //! @remark IMO only used for ExchangeNodePtr which is very questionable...
+    std::vector<Term>& GetTerms()
```
Psirus: Is there an `ExchangeNodePtr` in PDE? Could this be deleted?

src/mechanics/constraintsPde/Term.h
```diff
@@ -0,0 +1,66 @@
...
+
+    int GetConstrainedDofNumber() const
+    {
+        return mNode.get().GetDofNumber(mComponent);
+    }
+
+private:
+    //! @brief node reference, std::reference_wrapper is neat since it provides copy
+    //! ctor and assignment
+    std::reference_wrapper<const NodeSimple> mNode;
```
Psirus: What is the advantage over a normal reference?

test/mechanics/cell/PatchTest.cpp
```diff
@@ -1,16 +1,21 @@
 #include "BoostUnitTest.h"
 
+#include <numeric>
```
Psirus: I don't think this is needed (anymore).

test/mechanics/cell/PatchTest.cpp
```diff
@@ -194,3 +185,120 @@ BOOST_AUTO_TEST_CASE(PatchTest)
...
+
+        auto analyticSolution = analyticDisplacementField(coord);
+
+        BOOST_TEST_MESSAGE("Node at " << coord.transpose() << " with dofs " << displNode.GetDofNumber(0) << ","
+                                      << displNode.GetDofNumber(1));
+        BOOST_TEST_MESSAGE("( " << displNode.GetValues().transpose() << " )");
+
+        BoostUnitTest::CheckEigenMatrix(displNode.GetValues(), analyticSolution);
+    }
+}
```
Psirus, line 295: This should become an integration test in the near future.

test/mechanics/constraintsPde/ConstraintCompanionPde.cpp
```diff
@@ -0,0 +1,197 @@
+#include "BoostUnitTest.h"
+#include <fakeit.hpp>
+//#include <cmath>
+//#include "mechanics/groups/Group.h"
```
Psirus: Just delete those lines.

test/mechanics/dofs/DofNumbering.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+    std::set<int> dofNumbers{nodeUnconstrained0.GetDofNumber(0), nodeUnconstrained0.GetDofNumber(1),
+                             nodeUnconstrained1.GetDofNumber(0), nodeUnconstrained1.GetDofNumber(1)};
+    BOOST_CHECK_EQUAL(dofNumbers.size(), 4);
+    BOOST_CHECK_EQUAL(*dofNumbers.rbegin(), 3);
+
+    dofNumbers.insert(nodeConstrained.GetDofNumber(0));
+    dofNumbers.insert(nodeConstrained.GetDofNumber(1));
+    BOOST_CHECK_EQUAL(dofNumbers.size(), 6);
+    BOOST_CHECK_EQUAL(*dofNumbers.rbegin(), 5);
```
Psirus: Instead of using `set` to test the uniqueness of dof numbers, which was not immediately obvious to me, why not use `std::unique_copy` and compare the lengths? This would also result in less "magic numbers". And then check `BOOST_CHECK_EQUAL(nodeUnconstrained1.GetDofNumber(1), 3)` and `BOOST_CHECK_EQUAL(nudoConstrained.GetDofNumber(1), 5)` separately.

Psirus (approved):
Looks good. Nicely tested and commented. I've only noticed minor things. @vhirtham In the future, please format your commit messages according to https://nuto.readthedocs.io/en/latest/CodingStyle.html, section "Commit messages", especially with regard to the imperative case. Thanks guys! ❤️

src/mechanics/constraintsPde/Constraints.cpp
TTitscher: These warnings are all over the place. I will fix them in a dedicated PR. Any thoughts about how to change it? Simply `int --> size_t`?

src/mechanics/constraintsPde/Term.h
TTitscher: Try to replace the like with a const reference and look at the compiler output. 

I tried to explain it in the comments, better believe it :) Think about a class with a const reference member. You now have to define custom copy/move ctor/assigment. And now try to write assignment operators. For me, this is not trivial and the fact that I have to provide them at all is annoying. `reference_wrapper` takes care of this.

src/mechanics/constraintsPde/Constraints.cpp
Psirus: I guess so.

src/mechanics/constraintsPde/Constraints.cpp
```diff
@@ -58,7 +58,7 @@ Eigen::SparseMatrix<double> Constraints::BuildConstraintMatrix(DofType dof, int
                 int transformedDof = globalDofNumber - numIndependentDofs;
                 if (transformedDof != iEquation)
                     throw Exception(__PRETTY_FUNCTION__, "The numbering of the dependent dofs "
-                                                         "is not in accordance to the equation numbering.");
+                                                         "should be in accordance to the equation numbering.");
```
Psirus: Sorry, this is not what I meant. Compare "according **to**" and "in accordance **with**".

src/mechanics/constraintsPde/Term.h
Psirus: Hm. maybe I just misread the comment then, but still don't understand :) Why does the `reference_wrapper` provide a copy constructor that the `NodeSimple` doesn't provide? Or put differently, why is this a problem here and not at the other places where we use reference members? (Also, I've tried replacing it with a normal reference, and at least the unit test `ConstraintsPde` compiles just fine. Can't try the rest atm because old laptop.)

TTitscher: Since I think this is more about c++ than NuTo, I chose not to change the comment in the code. The theory is [here](http://en.cppreference.com/w/cpp/language/copy_assignment). And [this](http://en.cppreference.com/w/cpp/utility/functional/reference_wrapper) is also worth reading.

The compiler (normally) implicitly creates copy/move ctor/assign for you. A const reference member (e.g. `const NodeSimple& mNode`) implicitly deletes the assignment operators for the class. (Maybe this was the missing piece of information?) This makes `Term` not assignable. Note that the default copy/move _constructor_ are implicitly created. 
If we want to use it with the STL, we can now have a look at the specific __type requirements__. E.g. [`std::vector<>::push_back`](http://en.cppreference.com/w/cpp/container/vector/push_back) that might be the only method used in `ConstraintsPde` requires the type to be `CopyInsertable` (for non language lawyers like me: = copy constructable). This is the case for `Term`, no problem here.
In other cases, e.g. `ConstraintCompanionPde` we use [`std::insert`](http://en.cppreference.com/w/cpp/container/vector/insert). This requires Copy/MoveAssignable` and will fail.

If we use a pointer `NodeSimple* mNode`, assignment is not a problem. And I think this is what `std::reference_wrapper` does, with value semantics.

> Or put differently, why is this a problem here and not at the other places where we use reference members?

1) We want `MeshFem` to be 
    - MoveConstructable (`return tmpMesh`) and 
    - MoveAssignable (`mesh = Transform(mesh, f)`). 

     Thus all classes below have to fulfill this requirement and you will see `reference_wrapper` there. The code line above might be the only place where I specifically commented on that in the code.
2) We rarely deal with values types. We mostly work on polymorphic types that have no assignment at all and require a `clone()`. Here, we do not have that problem. Either an owning `ptr_container<Interface>` or a non-owning `container<Interface*>` deals with assignments.

---
I hope this clears things up. Maybe you can add the required comments to the code that would have made it understandable without this PR comments.

Psirus: Thanks, that was very helpful indeed. What do you think of

    //! @brief node reference
    //! @remark `std::reference_wrapper` is used instead of a reference to
    //! make `Term` default CopyConstructible and CopyAssignable

which says less about what `reference_wrapper` does and maybe a bit more about what it means for the current class? I'll add a commit if you agree.

TTitscher: Way better code comment, go ahead! For correctness - since a reference is `CopyConstructable` by default, we should remove this word from the comment, right? 

And while you're at it, can you fix the exception message with the _accordance_ thing as well? xD

test/mechanics/constraintsPde/ConstraintCompanionPde.cpp
vhirtham: Will add this soon. At the moment I just added the edge cases with one or more zeros in the direction vector.

src/mechanics/constraintsPde/ConstraintCompanion.cpp
vhirtham: I don't trust anybody!!! Especially not myself. Not sure if this was just a relict from previous commits. Anyhow, I removed it.

test/mechanics/constraintsPde/ConstraintCompanionPde.cpp
```diff
@@ -145,12 +145,36 @@ BOOST_FIXTURE_TEST_CASE(Direction_test, Helpers)
...
         expectedEquations.push_back(expectedEquation);
     }
     BOOST_CHECK(eqs == expectedEquations);
+
+
+    // Tests if terms with a coefficient of 0 are ignored correctly
+
+    someDirection << 0., 3., -4.;
+    eq = ConstraintPde::Direction(node1, someDirection, sinFunction);
+    expectedEquation = ConstraintPde::Equation(node1, 2, rhsValueAdjustedFunc(sinFunction, -5. / 4.));
```
TTitscher, line 164: `-5/4`. For someone, who does not know anything about this internal normalization, this may be confusing. I dont't know if
`rhsAdjustedFactor = someDirection.norm() / someDirection[2]`¸
would be more understandable... Well, maybe its fine... You decide.

src/mechanics/constraintsPde/Equation.h
```diff
@@ -0,0 +1,60 @@
...
+    {
+        return mTerms.front().GetConstrainedDofNumber();
+    }
+
+private:
+    //! @brief rhs function
+    RhsFunction mRhs;
+
+    //! @brief terms
+    std::vector<Term> mTerms;
```
joergfunger, line 56: Why did you decide to put all terms (including the dependent one) into a single vector? When building the constraint matrix or when checking for the terms (only independent dofs in the terms >1) there has to be made this distinction anyway. And the check/assert for the factor of the dependent term could be avoided.

src/mechanics/constraintsPde/Constraints.cpp
```diff
@@ -0,0 +1,124 @@
...
+        const auto& equation = equations[iEquation];
+        for (Term term : equation.GetTerms())
+        {
+            double coefficient = term.GetCoefficient();
+            int globalDofNumber = term.GetConstrainedDofNumber();
+            if (globalDofNumber == -1 /* should be NodeSimple::NOT_SET */)
+                throw Exception(__PRETTY_FUNCTION__,
+                                "There is no dof numbering for a node in equation" + std::to_string(iEquation) + ".");
+
+            if (globalDofNumber >= numIndependentDofs)
```
joergfunger, line 54: This works well for the current (serial) implementation. In the future (no problem for the current version, just a remark), I think it would be advantageous not to identify the dependent/independent terms based on the inequality dof>numIndependent dofs. In particular, this test could be removed if the dependent terms are stored separately. Regarding the same issue, I think it would also be better to distinguish between a global dof number and a domain dof number.

src/mechanics/constraintsPde/Constraints.cpp
```diff
@@ -0,0 +1,124 @@
...
+    Term newDependentTerm = e.GetTerms()[0];
+
+    if (Contains(mDependentTerms, newDependentTerm))
+        throw Exception(__PRETTY_FUNCTION__, "The dependent dof of the new equation "
+                                             "is already constrained as a dependent dof in another equation.");
+
+    if (Contains(mOtherTerms, newDependentTerm))
+        throw Exception(__PRETTY_FUNCTION__, "The dependent dof of the new equation "
+                                             "is already constrained in another equation.");
+
```
joergfunger, line 109: Shouldn't there be the check for the factor to be one here as well (not only when building the constraint matrix)?

joergfunger: The error message is misleading. Proposition "The dependent dof of the new equation is already used as an independent term in another constraint equation"

src/mechanics/constraintsPde/Equation.h
```diff
@@ -0,0 +1,60 @@
...
+
+//! @brief stores a constraint equation
+class Equation
+{
+public:
+    //! @brief ctor with constant rhs, defaults to 0
+    //! @param dependentNode node reference
+    //! @param dependentComponent component in the dof vector of the node
+    //! @param rhs value for the constant rhs
+    Equation(const NodeSimple& dependentNode, int dependentComponent, RhsFunction rhs)
```
joergfunger, line 19: What happens, if someone tries to add an equation of dof type e.g. displacements with a node that has temperature dofs?

src/mechanics/dofs/DofNumbering.cpp
```diff
@@ -0,0 +1,69 @@
+#include "mechanics/dofs/DofNumbering.h"
+
+using namespace NuTo;
+
+//! @brief build dof numbering, starting at 0, for all `nodes` regardless of constraints
+//! @return total number of dofs in `nodes`
+int InitialUnconstrainedNumbering(const Groups::Group<NodeSimple>& nodes)
```
joergfunger, line 7: The dof numbering is separate for each dof_type, shouldn't there be a check for nodes to belong to this dof_type?

joergfunger (commented):
Thanks for the first constraint implementation.

src/mechanics/constraintsPde/Constraints.cpp
TTitscher: Proposition accepted. And the check for the factor to be one is not needed since there is no way to construct an equation with a dependent term factor != 1.

src/mechanics/constraintsPde/Equation.h
TTitscher: This will cause an evil bug. 
We could have a method `AreAllConstrainedTermOfDofType(DofType, mesh)` or something.

src/mechanics/constraintsPde/Equation.h
TTitscher: An equation can only be created via one ctor. This ctor makes sure that
- the dependent term always exists
- is the first term in the vector mTerms
- has the factor 1.

src/mechanics/constraintsPde/Equation.h
TTitscher: Btw, I would not check that automatically since this would introduce a lot of dependencies. We could have:
~~~cpp
// in main file, if needed:
Group<Node> dofNodes = mesh.NodesTotal(dof);
Group<Node> constrainedDofNodes = constrains.ConstrainedNodes(dof);
if (not Group::Difference(constrainedDofNodes, dofNodes).Empty())
   throw "You messed up.";
~~~

PR #127, opened 2017-10-25
Remove running examples/integr tests from targets

Psirus:
Previously, `make examples` and `make integrationtests` not only build
the respective executables, but also ran them. This commit changes that,
so that they are only build. If you want to run the tests, use `ctest -R
examples` and `ctest -R integration`, respectively.


TTitscher (approved):
My understanding of `make` is more like _build something_ rather than _run_ something. So I like. Very much.

PR #128, opened 2017-10-26
Pde mesh move problem

TTitscher:
`ValueVector.h:`
~~~
//! @brief container that stores values of T and keeps references to these values valid
//! @remark Normally, it is a no-brainer to use boost::ptr_vector here. It provides all
//! the nice value semantics for free. However, there is an issue regarding move.
//!         https://stackoverflow.com/questions/44130076/moving-a-boostptr-vector
//! Therefore, here is a custom implementation with only the basic features.
~~~

We want our mesh to be movable. `boost::ptr_vector` does not behave as expected and is replaced.

### DIFF
--- a/src/base/ValueVector.h
+++ b/src/base/ValueVector.h
@@ -1,27 +1,75 @@
 #pragma once
 
-#include <boost/ptr_container/ptr_vector.hpp>
+#include <vector>
+#include <memory>
+#include <boost/iterator/indirect_iterator.hpp>
+
 
 namespace NuTo
 {
 
 //! @brief container that stores values of T and keeps references to these values valid
+//! @remark Normally, it is a no-brainer to use boost::ptr_vector here. It provides all
+//! the nice value semantics for free. However, there is an issue regarding move.
+//!         https://stackoverflow.com/questions/44130076/moving-a-boostptr-vector
+//! Therefore, here is a custom implementation with only the basic features.
 template <typename T>
-class ValueVector : public boost::ptr_vector<T>
+class ValueVector
 {
-public:
+    using Data = std::vector<std::unique_ptr<T>>;
+    //! @brief indirect (dereferencing) iterator to provide value semantics for the iterators
+    typedef boost::indirect_iterator<typename Data::iterator> IndirectIterator;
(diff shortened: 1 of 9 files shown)


src/base/ValueVector.h
```diff
@@ -1,27 +1,75 @@
...
+    typedef boost::indirect_iterator<typename Data::const_iterator> ConstIndirectIterator;
 
+public:
     template <typename... TArgs>
     T& Add(TArgs&&... args)
     {
-        this->push_back(new T(std::forward<TArgs>(args)...));
-        return this->back();
+        mData.push_back(std::make_unique<T>(std::forward<TArgs>(args)...));
+        return *mData.back();
```
Psirus, line 29: I'm uncertain about the operator precedence is such cases (obviously this is correct), so I'd prefer `return *(mData.back());`. What do you think?

src/base/ValueVector.h
```diff
@@ -1,27 +1,75 @@
...
+    }
+
+    typename Data::size_type Size() const
+    {
+        return mData.size();
+    }
+
+    const T& operator[](int i) const
+    {
+        return *(mData[i]);
```
Psirus: See! :) [BTW: `*mData[i]` would be the same.]

src/mechanics/mesh/UnitMeshFem.cpp
```diff
@@ -54,10 +54,11 @@ MeshFem UnitMeshFem::CreateQuads(int numX, int numY)
...
+    MeshFem newMesh = std::move(oldMesh);
     // Build a group to avoid duplicates. Otherwise, the transformation is applied multiple times.
-    // This is, however, a bit of a bottleneck for big meshes. 
-    for (auto& node : oldMesh.NodesTotal())
+    // This is, however, a bit of a bottleneck for big meshes.
+    for (auto& node : newMesh.NodesTotal())
         node.SetValues(f(node.GetValues()));
 
-    return oldMesh;
+    return newMesh;
```
Psirus: I'd rather have `return std::move(oldMesh);` here instead of the extra line at the beginning. And the comment is not applicable anymore.

Psirus (approved):

PR #131, opened 2017-11-06
Pde neumann3 d

potto1:
3D Neumann BC. The BC is applied by boundary elements, which are facets of 3D elements. Actually only a bit of code (Jacobian.h), but a lot of testing.


src/mechanics/cell/Jacobian.h
```diff
@@ -2,6 +2,7 @@
 
 #include <eigen3/Eigen/Dense> // for determinant
 #include "mechanics/interpolation/TypeDefs.h"
+#include <iostream>
```
TTitscher: Not needed here.

src/mechanics/integrands/NeumannBc.h
```diff
@@ -5,6 +5,7 @@
 #include "mechanics/interpolation/TypeDefs.h"
 #include "mechanics/cell/CellIpData.h"
 #include "mechanics/cell/CellData.h"
+#include <iostream>
```
TTitscher, line 8: Not needed here.

src/mechanics/cell/Jacobian.h
```diff
@@ -33,20 +34,34 @@ class Jacobian
...
             }
             case 3:
             {
+                const int numRows = derivativeShapeFunctions.rows();
+                NodeValues nodeCopy = nodeValues;
+                // see CalculateFixedSize(...) for the Eigen::Map functionality
+                auto nodeBlockCoordinatesMap =
+                        Eigen::Map<Eigen::Matrix<double, 3, Eigen::Dynamic>>(nodeCopy.data(), 3, numRows);
+
+                Eigen::MatrixXd nodeBlockCoordinates = nodeBlockCoordinatesMap;
```
TTitscher, line 57: Can we eliminate the `nodeBlockCoordinates` object? The point of this `Eigen::Map` is to provide a _view_ to some data. So the `nodeCopy.data()` is interpreted as a `Matrix3Xd`, without a copy. Copying the `Map` into a `Matrix` somehow eliminates this advantage. Or is there a problem with the multiplications on the lines below?

TTitscher (changes requested):
The `Jacobian.h` code looks fine to me, thanks! 

However, I would like to rework the tests... 
Short explaination: In our _old NuTo_, we have two transformations in the boundary elements: 
- 2D natural --> 3D natural
- 3D natural --> 3D xyz

I implemented it this way to build boundary elements that can calculate a derivative in normal surface direction. It turns out that is rarely needed and produces a lot of complications, like a lot of zeros in the resulting output vectors/matrices.

For the neumann integrand we do not need this normal surface derivatives and we only want one transformation:
- 2D natural --> 3D xyz 

Your tests seem to mimic our _old_ style of the transformation and use it as a _reference solution_ to compare against the new style. I see several problems with that:
- IMO the _reference solution_ is (as well as our _old NuTo_ approach) hard to understand. Imagine someone who is not familiar with the _old_ appraoch. Could he follow this test? Where do all those numbers/loops/transformations come from?
- How can we be sure that there are no bugs in your newly implemented _reference solution_? There is no test against a hardcoded value. 
- All tests work on a unit tetrahedron where the xyz coordinates coincide with the natural coordinates. This seems like a trivial test case since all jacobians should be 1, right?
- All this 2D3D transformation code seems to be duplicated in `Jacobian.cpp` and `NeumannBc.cpp`

I would expect a test to look like this: Have a 2D triangle somehow rotated in 3D. Calculate its area `area_xyz` manually. This is the reference. Its area in the natural coordinate system is `area_natural = 0.5`. Now test if the jacobian returns the right ratio of `area_xyz` to `area_natural`.

For the NeumannBC test, the situation is similar. I would expect a 2D element in 3D with `area` and a surface `pressure` applied in a certain direction. The (summed up) resulting force calculated by the newmann integrand should somehow be `area * pressure`.

potto1:
The tetrahedron compared with the facet solution is not a reference one. I just wanted to be sure that the 3D jacobian and the facet jacobian are the same. IMO the analytical proof is not straight forward.

TTitscher:
My point is more that we do not need the analytical proof. We just need the correct test result. And we can obtain that much easier. Maybe you can have a look at 1e2e829 where I implemented what I meant for a linear triangle in 3D. Do you think such a test is sufficient or do we need test higher order elements as well?

potto1:
Looks nice ...
I think we also need higher order tests, cause the shape functions are needed for jacobian calculations.

TTitscher:
If it is just about the shape functions (and their derivatives) itself: They are tested in the unit tests for the interpolation types (directory `test/mechanics/interpolation/`). So we can assume they are correct.

src/mechanics/cell/Jacobian.h
```diff
@@ -33,20 +34,34 @@ class Jacobian
         }
         else
         {
-            assert (globalDimension = interpolationDimension + 1);
+            assert(globalDimension = interpolationDimension + 1);
```
Psirus: Did you mean `==`?

test/mechanics/integrands/NeumannBc.cpp
```diff
@@ -80,324 +80,61 @@ BOOST_AUTO_TEST_CASE(NeumannBc1Din2D)
...
 {
     std::vector<NodeSimple> nodesDisp;
-    for (int i = 0; i < 4; i++)
+    for (int i = 0; i < 3; i++)
         nodesDisp.push_back(NodeSimple(Eigen::Vector3d(0, 0, 0)));
 
     std::vector<NodeSimple> nodes;
     nodes.push_back(NodeSimple(Eigen::Vector3d(1, 0, 0)));
     nodes.push_back(NodeSimple(Eigen::Vector3d(0, 1, 0)));
     nodes.push_back(NodeSimple(Eigen::Vector3d(0, 0, 1)));
```
TTitscher: The element has a `std::initializerlist` ctor. So you could do
~~~cpp
NodeSimple n0(Eigen::Vector3d(1, 0, 0));
NodeSimple n1(Eigen::Vector3d(0, 1, 0));
NodeSimple n2(Eigen::Vector3d(0, 0, 1));
// ...
ElementCollectionFem element({{n0, n1, n2}, coordinateInterpolation});
~~~
which is a bit shorter and avoids the `notePtrs` vector

test/mechanics/cell/Jacobian.cpp
```diff
@@ -59,6 +61,42 @@ BOOST_AUTO_TEST_CASE(Jacobian1Din2D)
...
+BOOST_AUTO_TEST_CASE(Jacobian2Din3D)
+{
+    // some manual test cases
+    CheckJacobian2Din3D(Eigen::Vector3d(0, 0, 0), Eigen::Vector3d(1, 0, 0), Eigen::Vector3d(0, 1, 0), 0.5);
+    CheckJacobian2Din3D(Eigen::Vector3d(0, 0, 0), Eigen::Vector3d(5, 0, 0), Eigen::Vector3d(0, 0, 7), 5. * 7 / 2);
+    CheckJacobian2Din3D(Eigen::Vector3d(0, 4, 0), Eigen::Vector3d(0, 0, 3), Eigen::Vector3d(0, 4, 3), 3. * 4 / 2);
+
+    // some random test cases
+    for (int i = 0; i < 100; ++i)
+    {
```
Psirus: Randomized tests might be nice, but I don't think it is going to catch any problems that the above lines don't catch. Maybe we could have a couple of tests with problematic input and how it is being handled: "Triangle" with two points that coincide, a quad with negative jacobian, this sort of thing.

test/mechanics/cell/Jacobian.cpp
```diff
@@ -59,6 +61,42 @@ BOOST_AUTO_TEST_CASE(Jacobian1Din2D)
     }
 }
 
+//! @brief calculate the area of a triangle (points a,b,c) element based on ||J|| and check vs. `correctArea`
+void CheckJacobian2Din3D(Eigen::Vector3d a, Eigen::Vector3d b, Eigen::Vector3d c, double correctArea)
```
Psirus: Can we name this `TriangleCheck2Din3D` or something, `CheckJacobian` seems more general than it is.

test/mechanics/integrands/NeumannBc.cpp
```diff
@@ -80,324 +80,61 @@ BOOST_AUTO_TEST_CASE(NeumannBc1Din2D)
...
+        CellIpData cellIpData(element, dummyJac, points[ip]);
+
+
+        auto gradient = neumannIntegrand.Gradient(cellData, cellIpData);
+
+        Eigen::VectorXd referenceResultOnIP = (1. / 6.) * Eigen::VectorXd::Ones(9);
+        referenceResultOnIP.block(3 * ip, 0, 3, 1) = (2. / 3.) * Eigen::VectorXd::Ones(3);
+
+        // results on each ip
+        BoostUnitTest::CheckEigenMatrix(gradient[dof], referenceResultOnIP);
```
TTitscher, line 138: My idea was to actually integrate the neumann integrand over the cell and compare the integrated result with something like `p * area(element)`. This would require the jacobian which is already tested in `Jacobian.cpp`. So I really like this test, since it is independent from the jacobian and it is not tested twice.

But the values of `refenceResultOnIP` are not obvious to me. Can you comment on where these values come from? Or what do you think about mixing both testing approaches? Actually perform the integration with a _prescribed_ jacobi determinant. So in this case the area should be `sqrt(3)/2`. The resulting forces are just `sqrt(3)/2 * ones`. So the sum of all x components should be `sqrt(3)/2`, same for y and z.

Psirus (changes requested):
Looking good. Only minor changes necessary (`==` and `iostream`) imho, the rest is optional. In the future, please format your commit messages according to https://nuto.readthedocs.io/en/latest/CodingStyle.html, section "Commit messages", especially with regard to the imperative case. Thank you!

TTitscher (approved):
IMO the tests are much better now. 
(optional) Maybe you can think about my comments to the test.

I would like you to fix the unused includes before merges.

test/mechanics/cell/Jacobian.cpp
TTitscher: You are right, these random tests provide no benefit. But I think the jacobian itself should not check invalid inputs.

Psirus: Fair enough.

potto1:
Thanks for reviewing

potto1:
Unfortunately the site describing the format of commit messages is not reachable on  https://nuto.readthedocs.io/en/latest/CodingStyle.html

Psirus (approved):

PR #133, opened 2017-11-14
Fix eigen include

TTitscher:
Until this PR, we had the cmake eigen include path set as
`/usr/include/`  (for debian)
instead of 
`/usr/include/eigen3/`
We "fixed" that by manually adding `eigen3` in our `#include` macros.
`#include <eigen3/Eigen/Core>`
instead of
`#include <Eigen/Core>`

Now, the include path is properly set and this is workaround is no longer needed.

[Leaving out the `boost` in e.g. `#include <boost/ptr_container/...>` would cause confusion. But `Eigen` remains `Eigen`.]

---
Since v3.3.1, Eigen itself provides configuration files to generate the _modern cmake_ target `Eigen3::Eigen`. Travis (Ubuntu14.04) ships with v3.2 and does not support this feature. So we stick to our custom `FindEIGEN.cmake` for now.


pmueller2:
Ok, good. 
I have a symlink in /usr/local/include to the Eigen folder which is the reason for the 2 non working commits I did ... Ahem.

PR #134, opened 2017-11-15
Pde remove operations

TTitscher:
- replace the `NuTo::Operation` with a `std::function` that includes the integrand like
~~~cpp
auto functionToIntegrate = [&](...args...) 
{ 
  return integrandInMainFile.FunctionToIntegrate(...args..., possibleAdditionalArgs);
};
cell.Integrate(functionToIntegrate);
~~~

  This can either be done [manually](https://github.com/nutofem/nuto/blob/PDE_remove_operations/test/mechanics/cell/Cell.cpp#L52) or with the help of a generic wrapper [`Bind`](https://github.com/nutofem/nuto/blob/PDE_remove_operations/test/mechanics/cell/PatchTest.cpp#L122).

- remove integrands from the cell
- remove `Integrand::Base` and `Integrand::TimeDependent::Interface`


test/mechanics/cell/PatchTest.cpp
```diff
@@ -132,18 +154,23 @@ BOOST_AUTO_TEST_CASE(PatchTestForce)
...
     // ************************************************************************
     SimpleAssembler assembler(dofInfo.numIndependentDofs, dofInfo.numDependentDofs);
 
-    GlobalDofVector gradient = assembler.BuildVector(cellGroup, {&displ}, Integrands::TimeDependent::Gradient());
-    GlobalDofMatrixSparse hessian = assembler.BuildMatrix(cellGroup, {&displ}, Integrands::TimeDependent::Hessian0());
+    GlobalDofVector gradient = assembler.BuildVector(momentumBalanceCells, {displ}, MomentumGradientF);
+    gradient += assembler.BuildVector({neumannCell}, {displ}, NeumannGradientF);
+
+    GlobalDofMatrixSparse hessian = assembler.BuildMatrix(momentumBalanceCells, {displ}, MomentumHessian0F);
+    hessian += assembler.BuildMatrix({neumannCell}, {displ}, NeumannHessian0F);
```
TTitscher: Remark from me. This is not an optimal solution. Adding two SparseMatrices with possibly different sparsity patterns can be very expensive. I added this issue to [cleanup after completion](https://github.com/nutofem/nuto/projects/1).

src/mechanics/cell/SimpleAssember.h
```diff
@@ -11,58 +11,58 @@ namespace NuTo
...
-    SimpleAssembler(const NuTo::DofContainer<int>& numIndependentDofs, const NuTo::DofContainer<int>& numDependentDofs)
+    SimpleAssembler(NuTo::DofContainer<int> numIndependentDofs, NuTo::DofContainer<int> numDependentDofs)
         : mNumIndependentDofs(numIndependentDofs)
         , mNumDependentDofs(numDependentDofs)
     {
     }
 
-    GlobalDofVector BuildVector(const Groups::Group<NuTo::CellInterface>& cells,
-                                const std::vector<NuTo::DofType*>& dofTypes, const VectorOperation& op) const
+    GlobalDofVector BuildVector(const Groups::Group<NuTo::CellInterface>& cells, Groups::Group<NuTo::DofType> dofTypes,
```
Psirus: What advantage has using group over `std::vector` in this case?

src/mechanics/integrands/NeumannBc.h
```diff
@@ -10,39 +11,30 @@ namespace NuTo
...
         NuTo::DofVector<double> gradient;
-        
+
         gradient[mDofType] = N.transpose() * mFactor;
         return gradient;
     }
 
-    NuTo::DofMatrix<double> Hessian0(const NuTo::CellData&, const NuTo::CellIpData& cellIpData, double = 0,
-                                     double = 0) override
+    NuTo::DofMatrix<double> Hessian0(const NuTo::CellData&, const NuTo::CellIpData& cellIpData)
```
Psirus: Is this method still necessary?

Psirus (approved):
Of course, I like this change very much. Thank you ❤️

src/mechanics/cell/SimpleAssember.h
TTitscher: Value semantics. Avoidance of (accidentally?) duplicate dof types may also be helpful. But mainly value semantics. We could also have that with `std::vector<std::reference_wrapper<DofType>>`. `Group<DofType>` is much shorter.

Psirus: In other places you're moving towards using `DofType` as a value, so why not here?

TTitscher: Right, I haven't thought of that. Tell me, if you prefer `std::vector` over `NuTo::Group`. I can change that, but have no strong opinion myself :)

Psirus: I have no strong allegiance to either; I was just curious.

TTitscher: You decide what to do!

Psirus: Is that so? 😬 Then I decree `std::vector`.

PR #135, opened 2017-11-16
Google benchmark

vhirtham:
As I said in the issue, I will write another macro for benchmarks based on the unit test macro. Apart from that, everything is working so far. Maybe we should also replace the benchmark example to a NuTo related one.

Open Questions:

- Where do we run the benchmarks? Jenkins?
- Should they be disabled/enabled by an own option?

Feel free to add more questions/suggestions.


Psirus:
So far this is looking good imho, the only thing that I would do is
> replace the benchmark example [by] a NuTo related one. 😃

vhirtham:
Okay, then I ll just do that tomorrow. The CMake script for benchmarks can be moved as another task to the cmake cleanup issue.

joergfunger (commented):
Looks good, I'd prefer this compared to manually written benchmarks. Would be great to present a summary at the nuto-meeting.

vhirtham:
I rewrote the "Solvers" benchmark to use google benchmark and added it as "SolversGB". I also added a cmake script to add benchmarks. This links all the nuto libs and the benchmark lib to the executable. It also adds "-Wno-suggest-override" to the compile flags of the executable to avoid warnings from the header. There might be a better solution, but I didn't have enough time to find one so far. Feel free to fix it.

Apart from that, if the tests don't fail, we can merge the PR.

TTitscher:
The more I think about it, the more I like to have it as a truly optional dependency. A lot of work has been done already, but hear me out.
The _user_ (that only does `make install`) will neither run our tests nor our benchmarks. (They are not installed...). So this dependency is simply not required. And honestly, how many of _our_ developers ran the benchmarks? 
For me, this seems like the perfect case for a `cmake -DENABLE_BENCHMARK=TRUE ...`. If you want to run benchmarks, compile and install `google benchmark` yourself. They should provide a modern cmake target, so 
~~~cmake
if(ENABLE_BENCHMARK)
  # add some path ...
  find_package(benchmark)
  include_directories(application/benchmarks)
endif()
~~~
or something is sufficient.

vhirtham:
I think the option makes a lot of sense. Didn't dare to suggest it myself because there was so many effort to get rid of the other options :smile: I can set this ASAP, but I suggest to stick with the embedded version. There is no out of the box find_package version that we can use and as long as you don't turn on the benchmarks, you won't get any trouble by not pulling google benchmark.

TTitscher:
We can do it as you suggested. But there is a find_package. 
If you `sudo make install` the google benchmark lib, copy a random example into `custombenchmark.cpp` and have the following `CMakeLists.txt`, everything works fine
~~~cmake
cmake_minimum_required (VERSION 3.0)
project (stuff)
set(THREADS_PREFER_PTHREAD_FLAG ON)
find_package(Threads REQUIRED)
find_package(benchmark REQUIRED) # will create target 'benchmark'
add_executable(custombenchmark custombenchmark.cpp)
target_link_libraries(custombenchmark benchmark Threads::Threads)
~~~

Psirus:
So than there is no need for `ENABLE_BENCHMARK`. Just make `find_package(benchmark)` not required, and wrap the `add_subdirectory(benchmark)` in a `if(${benchmark_found})` (or whatever the variable is called).

vhirtham:
Okidoki. Will make the adjustments later and push it.

CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ include(NuToMacros)
 set_compiler_flags()
 get_git_head_revision(git_refspec commit_nr)
 check_for_dependencies()
+#setup_google_benchmark()
```
Psirus: Lets just remove this.

README.md
```diff
@@ -57,6 +57,17 @@ In the end, run the test suite to see if all went well
 make test
 ```
 
+If you want to use the benchmarks, checkout google benchmark from github, build and
```
Psirus: This should be part of the developer documentation in Doxygen, rather than in the README, imho.

Psirus (approved):

CMakeLists.txt
vhirtham: Ups, yeah, forgot it. As you can see its already disabled by a comment :stuck_out_tongue_closed_eyes:

README.md
vhirtham: Okay, will be removed with next commit.

Issue #130, opened 2017-11-01
PDE visualize


TTitscher:
I tried to implement a visualization prototype and wanted to discuss it, before we continue working on that. If you want to jump directly at the code, you can have a look a [_visualization group_](https://github.com/nutofem/nuto/blob/PDE_visualize/src/visualize/GroupAverage.h) and its [_usage_](https://github.com/nutofem/nuto/blob/PDE_visualize/test/visualize/GroupAverage.cpp).

# Prerequiste

Currently, we use the [VTK-XML file format](https://www.vtk.org/Wiki/VTK_XML_Formats) with the __UnstructuredGrid__ type. This mainly contains four field types
- `Points` - ONE list of 3d coordinates, each coordinate has a point index
- `Cells` - ONE list of point indices that connect to cells
- `PointData` - MANY lists of data for each point index (scalar data, vector data, ...)
- `CellData` - MANY lists of data for each cell index (scalar data, vector data, ...)

It is important that the data fields are consistent and available for all points and cells. 

> Visualizing a multiphysics mesh, where e.g. half the points have `displacement` point data and the other half has `temperature` point data is __not__ possible in the same file. Each data field must contain data for all points/cells.

This may not be the case for other file formats - who knows. But we currently work with VTK.

# Some classes

### `NuTo::Visualize::UnstructuredGrid`
- provides simple interface to write all the fields above
- exports to VTU (= vtk in xml format) in human readable and binary format
- reuse this class. Everything that is about to come is "just" a wrapper to this class.

### _Visualization groups_ 
- contain cells with the same shape/interpolation/integration/integrationPointData/dofs. Since this corresponds with the pde type, there is (most commonly) one visualization group per PDE.
- in the end, they produce one visualization output file. For 3 PDEs you will have 3 visualization groups and 3 output files.
- optional: Have separate functions for writing the _geometry_ (= fields `Points` and `Cells`) and _data_ (= fields `PointData` and `CellData`). The geometry is constant and needs to be calculated only once. (If the geometry changes, we call the `ExtractGeometry()` method more often, no problemo.)
- different types of visualization = different visualization groups: 
    - voronoi cells = one visualization cell for each integration point
    - averaged cells = one visualization cell for each cell, all integration point data is averaged
    - extrapolating cells = one visualization cell for each cell, all integration point data is extrapolated to points. No `CellData`, only `PointData`.

[...]

Issue #129, opened 2017-11-01
PDE access integration point data


TTitscher:
We need to access the data on each integration point for
- visualization files
- restart files
- other post processing
- tests (stress == expected stress?)
- ...

and there is currently no pattern for that.

### Possible data structure

- define `IpValue` as a single data entry at an integration point
~~~cpp
struct IpValue
{
    std::string name;
    Eigen::VectorXd data;
};
~~~
- one integration point may want to visualize several data entries. Collect them in 
~~~cpp
using IpValues = std::vector<IpValue>;

// e.g.
IpValues valuesAtASingleIntegrationPoint = {{"Stress", Eigen::Vector3d(...)}, {"Strain", Eigen::Vector3d(...)}};
~~~
- each cell has several integration points. Collect them in a vector
~~~cpp
virtual std::vector<IpValues> CellInterface::XYZ???(???) = 0;
~~~

### Possible interface

Some options where the integrand visualizes all its data:
- demand a `virtual IpValues Integrand::Interface::GetIpValues() = 0;` for all integrands
- add a `VisualizeAllOperation`

It may be nice to return only part of the integration point values
- add multiple visualize operations `VisualizeStrainOperation, VisualizeStressOperation`
- stick to one operation/method an provide options as arguments
    - `Integrand::GetIpValues({"Stress", ...});`
    - `Cell::Visualize(VisualizeOperation({"Strain", "Stress", "Damage"});`

Or something completely different.

Ideas?

Psirus:
With the danger of angering someone over my repeated insistence on this, I think these things would be much easier if you just have a single integrand that stores all the history. Restart files simply become `outstream << myIntegrand`, and visualization should just call one method multiple times, say `integrand.GetOutput("Stress", cellId, ip)` instead of defining yet another data structure `IpValue`. The same goes for tests.  

In terms of visualization, I would probably favour an interface with strings, so e.g.

~~~cpp
visualizer.Visualize(["Stress", "Strain"])

// in Visualize(outputList):
for cell in cells:
    VisualizeCellGeometry(cell)
    VisualizeCellData(integrand, cell.Id, outputList)

// in VisualizeCellData(integrand, cellId, outputList):
for output in outputList:
    data = integrand.GetOutput(output, cellId)
    WriteCellData(output, data)

// equivalent for IPData if desired
~~~

TTitscher:
[...]

PR #136, opened 2017-11-17
Pde visualize

potto1:
* Visualization for PDE
* AverageVisualizer (formerly GroupAverage) and VoronoiVisualizer for  tensor product cells
* Triangles, Tetrahedra, ... still need to be done


Psirus:
Is there some document that explains the design process? What is the problem you're trying to solve, what is the API, how are the classes connected, etc? Could you add an example of how this is supposed to be used, maybe as an extension of [PatchTest.cpp](https://github.com/nutofem/nuto/blob/PDE_visualize/test/mechanics/cell/PatchTest.cpp)?

potto1:
Actually there are two test files [https://github.com/nutofem/nuto/blob/PDE_visualize/test/visualize/VoronoiVisualizer.cpp](url) and [https://github.com/nutofem/nuto/blob/PDE_visualize/test/visualize/AverageVisualizer.cpp](url)
All the things are IMO far from to be done. Just opened the pull request to track the changes. The discussion is taking place in the issue #130 

I may be wrong thinking the pull request is the right instrument for this, IMO the branch is not mergable until the issue #129  is solved.

Visualizing the [PatchTest.cpp](https://github.com/nutofem/nuto/blob/PDE_visualize/test/mechanics/cell/PatchTest.cpp) is a good idea, but the function giving the IP values is not ready yet, IMO.

src/mechanics/cell/Cell.h
```diff
@@ -58,6 +58,35 @@ class Cell : public CellInterface
         return mId;
     }
 
+    Eigen::VectorXd Interpolate(Eigen::VectorXd naturalCoords) const override
+    {
+        const auto& coordinateElement = mElements.CoordinateElement();
+        return coordinateElement.GetNMatrix(naturalCoords) * coordinateElement.ExtractNodeValues();
```
TTitscher: There is the free function `Interpolate(ElementInterface, NaturalCoords)` for that.

src/mechanics/cell/Cell.h
```diff
@@ -58,6 +58,35 @@ class Cell : public CellInterface
...
+    }
+
+    Eigen::VectorXd Interpolate(Eigen::VectorXd naturalCoords, DofType dof) const override
+    {
+        const auto& element = mElements.DofElement(dof);
+        return element.GetNMatrix(naturalCoords) * element.ExtractNodeValues();
+    }
+
+    std::vector<Eigen::VectorXd>
```
TTitscher: Should we introduce a custom datatype for that? Maybe for now, just a file `IpData.h` containing `using IpData = Eigen::VectorXd`. 
I used that e.g. for the `DofVector`. First, it was just a typedef. When I needed more functionality, it became a class. But I only needed to change code inside DofVector file, no API changes.

Psirus: Personally, I'm not a big fan of that. At the locations where it is used, I'll know immediately what the type does and what it's member functions are when I read `Eigen::VectorXd`. With `IpData`, I'll have to remember that or look it up, putting undue strain on my small brain :)

src/mechanics/cell/Cell.h
```diff
@@ -58,6 +58,33 @@ class Cell : public CellInterface
...
+        return NuTo::Interpolate(mElements.CoordinateElement(), naturalCoords);
+    }
+
+    Eigen::VectorXd Interpolate(Eigen::VectorXd naturalCoords, DofType dof) const override
+    {
+        return NuTo::Interpolate(mElements.DofElement(dof), naturalCoords);
+    }
+
+    std::vector<Eigen::VectorXd>
+    Eval(std::function<Eigen::VectorXd(const CellData& cellData, const CellIpData& cellIpData)> f) const override
```
TTitscher: What do you think about a typedef like
~~~
using EvalFunction = std::function<Eigen::VectorXd(const CellData& cellData, const CellIpData& cellIpData)>;
~~~
similar to `VectorFunction, ScalarFunction,...`

src/mechanics/cell/CellInterface.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "mechanics/cell/IpValues.h"
```
TTitscher: I guess this is a leftover - together with the actual file `mechanics/cell/IpValues.h`

src/visualize/Point.cpp
```diff
@@ -3,16 +3,22 @@
...
+Point::Point(Eigen::Vector3d coordinates)
     : mCoordinates(coordinates)
 {
-    mData.resize(numData);
+}
+
+Point::Point(Eigen::VectorXd coordinates)
+    : mCoordinates(Eigen::Vector3d::Zero())
+{
+    const int dim = coordinates.rows();
```
TTitscher: Optional: Using `EigenCompanion::To3D()` here would save some lines. I guess, I wrote it actually, but forgot about it...

src/visualize/Point.h
```diff
@@ -13,8 +13,11 @@ class Point
 public:
     //! @brief ... constructor
     //! @param coordinates ... point coordinates
-    //! @param numData ... number of different point data fields
-    Point(Eigen::Vector3d coordinates, int numData);
+    Point(Eigen::Vector3d coordinates);
```
TTitscher: Is this method covered/equivalent to `Point(Eigen::VectorXd coordinates);` and can be removed?

src/visualize/TensorProductVoronoiHandler.cpp
```diff
@@ -0,0 +1,146 @@
...
+                mVoronoiCellCorners.push_back({start1, start1 + 1, start1 + numCellsPerDirection + 2,
+                                               start1 + numCellsPerDirection + 1, start2, start2 + 1,
+                                               start2 + numCellsPerDirection + 2, start2 + numCellsPerDirection + 1});
+            }
+        }
+    }
+}
+
+template <int TDim>
+std::vector<int> TensorProductVoronoiHandler<TDim>::WriteGeometry(const CellInterface& cell, UnstructuredGrid* grid)
```
TTitscher: I claim that the methods `WriteGeometry, WriteDofValues, CellData` will be the same for any other `VoronoiHandler`. Following YAGNI, it may be fine for now. To avoid code duplication, we could have a `VoronoiHandler` base class with said methods and let classes like `TensorProduct..` derive from that. 
But I suggest a `VoronoiHandler` that works on a `VoronoiTriangulation`. This way, the creation of this geometry is separated from its visualization. 
Btw: `WriteGeometry, WriteDofValues, CellData` do not need the template argument `TDim`. For me, this is a strong indicator that we can/should split this class into a `TDim` dependent part (the actual voronoi cell generation) and a general part (the `VoronoiHandler`. Even if it is just so save compile time.
What do you think?

test/mechanics/cell/PatchTest.cpp
```diff
@@ -54,7 +58,7 @@ MeshFem QuadPatchTestMesh()
      *    3-----------------------2
      * /| | - _        e2       / | -->
      * /| |     -7------------6   | -->
-     * /| | e5  /     e4      |   | -->
+     * /| | e4  /     e3      |   | -->
```
TTitscher, line 63: Apparently, visualizing `CellId` really found a bug :)

test/mechanics/cell/PatchTest.cpp
```diff
@@ -188,6 +192,24 @@ BOOST_AUTO_TEST_CASE(PatchTestForce)
...
+        EngineeringStrain<2> strain = cellIpData.GetBMatrixStrain(displ) * cellData.GetNodeValues(displ);
+        return linearElasticLaw.Stress(strain, 0.0, 0, 0);
+    };
+    visualize.CellData(stress, "Stress");
+
+    visualize.CellData([](const CellData&, const CellIpData&) { return Eigen::Matrix<double, 1, 1>(7.0); }, "Seven");
+    visualize.CellData(
+            [](const CellData& cd, const CellIpData&) { return Eigen::Matrix<double, 1, 1>(cd.GetCellId()); },
+            "CellId");
+    visualize.WriteVTKFile("outputVoronoi.vtu");
```
TTitscher: 1) `WriteVtkFile` would match our naming convention. 
2) VTK or VTU? VTU is the XML version of VTK. The method name suggests that the output file should be of type `*.vtk`. Since we actually write VTU, paraview expects the wrong format. So I suggest renaming the method to `WriteVtuFile()`. And adding a check like `assert(filename.EndsWith(".vtu")`.

test/visualize/AverageVisualizer.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+struct UnstructuredGridCheck
+{
+public:
+    static void Average(std::string filename)
+    {
+        pt::ptree tree;
+        pt::read_xml(filename, tree);
+        int numOfPoints = tree.get<int>("VTKFile.UnstructuredGrid.Piece.<xmlattr>.NumberOfPoints");
+        int numOfCells = tree.get<int>("VTKFile.UnstructuredGrid.Piece.<xmlattr>.NumberOfCells");
```
TTitscher, line 21: Optional (since I have no better idea): This tests the `QuadAverageHandler`, the `Visualizer`, the `UnstucturedGrid` and the `XMLWriter`. As I said, I have no better idea. But I think there is a rule that somehow states that the code should be designed in a (unit)testable way. Maybe we can come up with something in the future.

test/visualize/TestStructure.h
```diff
@@ -0,0 +1,70 @@
...
+    {
+        elements0.AddDofElement(d, NuTo::ElementFem({nd0, nd1, nd2, nd3}, interpolation));
+        elements1.AddDofElement(d, NuTo::ElementFem({nd1, nd4, nd5, nd2}, interpolation));
+    }
+
+    NuTo::Group<NuTo::CellInterface> Cells()
+    {
+        Eigen::VectorXd bla = Eigen::Vector3d::Constant(1.0);
+        fakeit::When(ConstOverloadedMethod(cell0, Interpolate, Eigen::VectorXd(Eigen::VectorXd)))
+                .AlwaysReturn(bla);
```
TTitscher, line 61: Renaming to `blub` would make the intention of the code clearer. 
Btw: Nice example of the mocking. I think we have never used the `ConstOverloadedMethod` before.

TTitscher (approved):
I think that further splitting of the `Handler` and their underlying geometry would make the code less prone to code duplication and make it easier to test. Like:
~~~cpp
struct VoronoiCells;
//! Cell handler that visualizes arbitrary `VoronoiCells`
class VoronoiHandler;
~~~

We can then have
~~~cpp
Visualizer<VoronoiHandler> visu(cells, TensorProductVornoiCells<2>(3));
~~~
which also reads quite nicely. Same goes for `AverageHandler` We can then test the creation of voronoi cells independent from their visualization. 

But, I like to visualize stuff asap. So this change is optional can be done as a refactoring step, as soon as more VisualizeHandlers are implemented. Or, if you agree with this idea, I can add that now.

Psirus:
While I agree that further splitting makes it easier to test, I'm not quite sure that the added complexity is worth the gain in testability. So yeah, this to me is an additional refactoring once the code duplication becomes more apparent. But I won't stop you if you want to do it now.

src/mechanics/cell/Cell.h
Psirus: I'm indifferent, but I'll change it for consistency.

PR #137, opened 2017-11-18
Compiler Warnings in Travis

Psirus:
This will add two additional jobs to our Travis build that fail when there are any compiler warnings. These have been missed previously, because one usually doesn't test on multiple compilers on ones own machine. Additionally, the flag `-Wextra` (see #21) has been added. This yields quite a few warnings that should be fixed. 

*Note:*
- These "Compiler warnings" jobs are allowed to fail; even if there is a warning, you'll still get the green tick.
- They fail on the first warning, even if there are more, to reduce Travis time. The corresponding "Normal build" will have all the warnings.
- This is only applied to the PDE branch, and I think it should stay that way for now. Fixing all the warnings in master is not worth the effort imho.


joergfunger (commented):
I think trying to remove the warnings in the PDE-branch is a good way to clean up.

.travis.yml
```diff
@@ -8,34 +8,41 @@ notifications:
...
-    - sudo apt-get install -y lcov curl libglu1-mesa-dev
+    - sudo apt-get -qq install -y g++-6 clang libeigen3-dev libgcc-6-dev libiomp-dev swig3.0 libboost-filesystem1.55-dev libboost-test1.55-dev libboost-mpi1.55-dev libopenblas-dev libmetis-dev libmumps-seq-dev libann-dev libarpack2-dev lcov curl libglu1-mesa-dev
 before_script:
     - mkdir build
     - cd build
-    - BUILD_OPTIONS="-DCMAKE_BUILD_TYPE=$BUILD_TYPE -DENABLE_OPENMP=$OPENMP -DCMAKE_CXX_FLAGS=$COVERAGE -DENABLE_MPI=TRUE"
-    - if [[ "$CXX" == "g++" ]]; then CXX=g++-6 cmake $BUILD_OPTIONS ..; fi
-    - if [[ "$CXX" == "clang++" ]]; then CXX=clang++ cmake $BUILD_OPTIONS ..; fi
+    - BUILD_OPTIONS="-DCMAKE_BUILD_TYPE=$BUILD_TYPE -DENABLE_OPENMP=$OPENMP -DENABLE_MPI=TRUE"
+    - if [[ "$CXX" == "g++" ]]; then CXX=g++-6 cmake $BUILD_OPTIONS -DCMAKE_CXX_FLAGS="$COVERAGE $WERROR" ..; fi
```
TTitscher, line 48: Is it advantageous to avoid all the optoins and `if`s here and put this up in the "stage" part (line 14 ff)?

Will this "stages" approach work with docker images?

Psirus: Afaik, each job could have a separate script for each [phase](https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle), so you could make the cmake line a separate `before_script` of each job and avoid this if. 

The stages are just a way of organizing the jobs. What you do in each job is not affected (I think).

TTitscher:
I cleaned some warnings. The remaining warnings only occur in release mode. Some parameters are only required in `assert`. How do we deal with that? 
(That only occurs in _old_ NuTo stuff...)
__Edit__ I realized we only do `Werror` in debug builds. Problem solved...

TTitscher (approved):

Psirus:
Hm, I don't see that as "solved". Rather, should we then not test both versions? And passing parameters just for asserts seems kinda weird.

PR #143, opened 2017-12-05
Pde cleanup mechanics

TTitscher:
- remove the _old_ NuTo code from the mechanics module
- remove corresponding unit tests
- deactivate integrationtests and examples in cmake
    - so they are easily accessible if someone wants to rewrite them using PDE (we can remove those as well...)

- TODO after there are no longer name clashes:
    - [x] rename `constraintsPde --> constraints` (files + namespace)
    - [x] rename `Groups::Group --> Group` (namespace)


Psirus (approved):
-54k Lines. 'Nuff said. :) 👍

TTitscher:
Yip. For me, this feels somehow strange, removing all that. On the other hand, I see no reason to keep it around, causing unnecessary compile targets and naming clashes.

src/mechanics/constitutive/inputoutput/EquivalentStrain.cpp
```diff
@@ -1,214 +0,0 @@
-/*
- * EquivalentStrain.cpp
```
joergfunger, line 2: Shouldn't we keep thoose helper functions?

src/mechanics/constitutive/inputoutput/EquivalentStrain.h
```diff
@@ -1,46 +0,0 @@
-#pragma once
```
joergfunger, line 1: This one should be kept as well (not in the core, but in the integrand lib)

src/mechanics/constitutive/laws/Creep.cpp
```diff
@@ -1,385 +0,0 @@
-#include "mechanics/constitutive/laws/Creep.h"
```
joergfunger, line 1: I would also move this to the integrand lib.

src/mechanics/constitutive/laws/EngineeringStressHelper.cpp
```diff
@@ -1,70 +0,0 @@
-#include "mechanics/constitutive/laws/EngineeringStressHelper.h"
```
joergfunger, line 1: Same goes here (IntegrandLib?)

joergfunger (commented):
Should we already at this point remove all the constitutive models we had? I kind of feel sorry about removing them instead of moving them to the new layout. This is now a complete rewrite, instead of a gradual transition (where one test file say for Mises plasticity is removed from the old nuto and at the same time directly added to the new version).

src/mechanics/constitutive/inputoutput/EquivalentStrain.cpp
TTitscher: now here https://github.com/nutofem/nuto/blob/PDE_remove_old/src/mechanics/constitutive/EngineeringStrainInvariants.h

TTitscher:
I thought that a gradual rewrite is still a rewrite. You can, for example, compare [new](https://github.com/nutofem/nuto/blob/PDE_remove_old/src/mechanics/constitutive/LinearElastic.h) vs [old](https://github.com/nutofem/nuto/blob/master/src/mechanics/constitutive/laws/LinearElasticEngineeringStress.cpp). Apart from the physics, they have nothing in common...

Psirus:
This "gradual" rewrite, where all the tests of old and new version pass is just too difficult. Travis takes ages, you end up with `NodeSimple` vs `NodeDof`, `Constraints` vs `ConstraintsPDE` and so on. This makes it much clearer what is new, we can already use the names we want, and compilation and running all the tests is much, much faster. In the end, we just compare `master` with `pde` to find what is still missing once we're "done".

joergfunger:
I perfectly agree, I just feel that this comparison at the end will result in loosing some functionality that we already have, but nobody is using it currently (say Mises plasticity or a cohesive formulation for the truss coupling with solids). But probably this is the price to pay.

Psirus:
Whether we loose functionality or not depends on how rigorously we are with "everything that worked before should work now", not on whether the corresponding file still exists in the new branch.

TTitscher:
I added the list _Implement before `merge master`_ to [project PDE](https://github.com/nutofem/nuto/projects/1) to keep track of the things that once worked and should also work in PDE.

PR #144, opened 2017-12-07
Pde ip data

vhirtham:
Since we now have a working example, we can discuss the details now. 

# **1. Interface changes**

I tried to minimize interface changes as good as I can.

**cell**
- Gets a uniqueID during construction and therefore inherits the all the functions

**cellData**
- Constructor now needs the cells ID
- Getter for cell ID 

**cellIPData**
- Constructor now needs the IPs number
- Getter for IP number

**MechanicsInterface**
- Tangent and Stress function have the cellID and IPnum as additional parameter

**Questions**
 - How to handle the cell ID? It was suggested to let it be in the users responsibility and pass the ID by the constructor. The inheritance from UniqueID will be removed.

# **2. Standard history data**

An approach how the standard history data can be implemented is currently in the HistoryData.cpp testfile:

~~~cpp
template <typename T>
class HistoryData
{
    std::map<int, std::vector<T>> mHisData;

public:
    const T& GetIPHistoryData(int cellNum, int ipNum) const
    {
        auto ipVecIt = mHisData.find(cellNum);
        if (ipVecIt == mHisData.end())
            throw Exception(__PRETTY_FUNCTION__, "Cell data not found! - Did you initialize the history data?");
        return (*ipVecIt).second[ipNum];
    }

    T& GetIPHistoryData(int cellNum, int ipNum)
    {
        auto ipVecIt = mHisData.find(cellNum);
        if (ipVecIt == mHisData.end())
            throw Exception(__PRETTY_FUNCTION__, "Cell data not found! - Did you initialize the history data?");
        return (*ipVecIt).second[ipNum];
    }

    void InitializeHistoryData(const Groups::Group<CellInterface>& cells, DofType dof)
    {
        for (CellInterface& cell : cells)
        {
            cell.Apply([this](const NuTo::CellData& cellData, const NuTo::CellIpData& cellIpData) {
                if (mHisData.find(cellData.GetCellId()) == mHisData.end())
                    mHisData.emplace(cellData.GetCellId(), std::vector<T>(cellData.GetNumIntegrationPoints()));
            });
        }
    }
};
~~~

This class contains the history data for each IP and provides a const and non-const member function to access the history data. It also provides an initialize function to resize the container to its necessary size. A law with history data should simply be derived from that class. However, here are some issues:

* The container is not optimal. We discussed that before. Best would be a boost::multi_array or an std::vector.  But there are some conditions for that: 
    - Cell num range from [ 0 ... numCells]
    - Same number of IPs for each cell.
* What happens if the same law is shared by multiple cellGroups? Currently this would cause some trouble, because you would need an own history data for each cellgroup. But what if the cell groups share some elements (not sure if that makes sense). 

Remark: An alternative way to have varying number of IPs in the cells was suggested by joergfunger. If we use a lookup table like a sparse matrix the number of IPs per cell is not relevant anymore.

# **3. Example law**

Here is the example law (creep) from the test which uses history data. I replaced the irrelevant code snippets by triple dots: 

~~~cpp
struct CreepHistoryData
{
    EngineeringStrainPDE<1> prevStrain;
    EngineeringStressPDE<1> prevStress;
    VectorXd gamma = Eigen::VectorXd::Zero(2);

    CreepHistoryData()
    {
        prevStrain[0] = 0.;
        prevStress[0] = 0.;
    }
};

class CreepLaw : public Laws::MechanicsInterface<1>, public HistoryData<CreepHistoryData>
{
    ....

    EngineeringStressPDE<1> Stress(EngineeringStrainPDE<1> strain, double delta_t, int cellNum,
                                   int ipNum) const override
    {
        const auto& hisData = GetIPHistoryData(cellNum, ipNum);
      ... 
    }

    ....

    void UpdateHistoryData(const NuTo::CellData& cellData, const NuTo::CellIpData& cellIpData, DofType dofType,
                           double delta_t)
    {
        // Get history data
        auto& hisData = GetIPHistoryData(cellData.GetCellId(), cellIpData.GetIPNum());

        // Calculate necessary values for update
        NuTo::BMatrixStrain B = cellIpData.GetBMatrixStrain(dofType);
        NuTo::NodeValues u = cellData.GetNodeValues(dofType);
        EngineeringStrainPDE<1> deltaCreep{DeltaCreep(hisData, delta_t)};
        NuTo::EngineeringStrainPDE<1> strain = B * u;
        NuTo::EngineeringStressPDE<1> stress = Stress(strain, delta_t, cellData.GetCellId(), cellIpData.GetIPNum());
        MechanicsTangent E = Tangent(strain, delta_t, cellData.GetCellId(), cellIpData.GetIPNum());
        NuTo::EngineeringStrainPDE<1> deltaStrain = strain - hisData.prevStrain;

        // The actual update
        hisData.prevStrain = strain;
        hisData.prevStress = stress;
        for (unsigned int i = 0; i < mE_KC.rows(); ++i)
            hisData.gamma[i] = Lambda(delta_t, i) * E / mE_KC[i] * (deltaStrain - deltaCreep) +
                               Beta(delta_t, i) * hisData.gamma[i];
    }

    ....
};
~~~

First we define a structure which contains all the history data that we need at each IP. Then our new law derives from the previously shown history data class:

~~~cpp
class CreepLaw : public Laws::MechanicsInterface<1>, public HistoryData<CreepHistoryData>
~~~

The history data can than be accessed by the getter in each function of your law/integrand:

~~~cpp
auto& hisData = GetIPHistoryData(cellData.GetCellId(), cellIpData.GetIPNum());
~~~

Remark: Instead of using this history data class, you can also write your own implementation into your integrand. 

If we now have a look at the update of the history data, it gets a little bit ugly because of a problem joergfunger mentioned. Since we don't store the results of the timestep, we have to calculate them again. Additionally our MomentumBalance integrand has currently no update history data routine. So what I did is looping over each cell and call the cells Apply function with the update routine of our law:

~~~cpp
auto MomentumUpdateHistoryDataF =
    std::bind(&CreepLaw::UpdateHistoryData, std::ref(creepLaw), _1, _2, displ, std::ref(delta_t));
...
for (auto& cell : momentumBalanceCells)
            cell.Apply(MomentumUpdateHistoryDataF);
~~~

Now there are some things that should be discussed:
- do we want to store the "pre update history data" somewhere, so that we don't have to calculate all the values again during the update. There are other benefits than just a speed benefit mentioned by joergfunger. maybe he can write them down himself, since I don't know much about it.
- Should we provide a function that loops over all cells and performs the update or is that the user responsibility. I actually don't know were we would place this function.
- For the special case of the momentum balance integrand: How do we want to handle history data? We can either let the law contain the history data or the momentum balance integrand. I think the law makes more sense, since I see a lot of problems, when putting it at the integrand. However, we probably need an UpdateHistoryData function at the integrand that calls the same function of the law. So the `Laws::MechanicsInterface` would need that virtual function which by default does nothing.

# **4. Misc**
Should the user do the initialization himself or do we want to add some kind of automation? Currently it is the responsibility of the user:
~~~cpp
creepLaw.InitializeHistoryData(momentumBalanceCells, displ);
~~~

# **Implementation list**
- [x] Remove unique ID from cell and add an additional  parameter to the constructor
- [x] Test correct cell ID passing to functions  (Cell.cpp)

# **Disclosure**

I hope I didn't forget anything. If I did, please remind me or just add it. I will add more items to the checklist as soon as we get some results in the discussion.


src/mechanics/cell/CellData.h
```diff
@@ -14,11 +14,23 @@ namespace NuTo
 class CellData
 {
 public:
-    CellData(const ElementCollection& elements)
+    CellData(const ElementCollection& elements, int numIntegrationPoints, int cellID)
```
TTitscher: Why `numIntegrationPoints`? Maybe because you chose to initialize the history data in your test via the cell interface. This is not the right way, IMO. 
Even if we choose to allow a varying number of integration points, this should not be here. We _know_ that in our main file and can initialize our history data there.

src/mechanics/cell/CellIpData.h
```diff
@@ -12,11 +12,11 @@ namespace NuTo
 class CellIpData
 {
 public:
-    CellIpData(const ElementCollection& elements, NuTo::Jacobian jacobian,
-               NaturalCoords ipCoords)
+    CellIpData(const ElementCollection& elements, NuTo::Jacobian jacobian, NaturalCoords ipCoords, int ipNum)
         : mElements(elements)
         , mJacobian(std::move(jacobian))
         , mIPCoords(std::move(ipCoords))
+        , mIPNum(ipNum)
```
TTitscher: 1) Naming. We decided to write things like `Ip or Pde or Iga` instead of `IP, PDE, IGA` - same goes for `CellID`
2) On the first glance, I thought _Why do we need the number of integration points here?_ Maybe rename that to `IpId`?

src/mechanics/constitutive/laws/MechanicsInterface.h
```diff
@@ -11,8 +11,9 @@ template <int TDim>
 struct MechanicsInterface
 {
     using MechanicsTangent = Eigen::Matrix<double, Voigt::Dim(TDim), Voigt::Dim(TDim)>;
-    virtual EngineeringStressPDE<TDim> Stress(EngineeringStrainPDE<TDim> strain, double deltaT) const = 0;
-    virtual MechanicsTangent Tangent(EngineeringStrainPDE<TDim>, double deltaT) const = 0;
+    virtual EngineeringStressPDE<TDim> Stress(EngineeringStrainPDE<TDim> strain, double deltaT, int cellNum,
```
TTitscher: Naming consistency: `cellNum = cellData.GetCellId()`?

test/mechanics/integrands/HistoryData.cpp
```diff
@@ -0,0 +1,397 @@
...
+class HistoryData
+{
+    std::map<int, std::vector<T>> mHisData;
+
+public:
+    //! @brief Returns the history data for a specific integration point
+    //! @param cellNum: Number of the cell containing the integration point
+    //! @param ipNum: Number of the integration point
+    //! @return History data of the integration point
+    const T& GetIPHistoryData(int cellNum, int ipNum) const
```
TTitscher: what about `T& operator()(...)`?

test/mechanics/integrands/HistoryData.cpp
```diff
@@ -0,0 +1,397 @@
...
+        auto ipVecIt = mHisData.find(cellNum);
+        if (ipVecIt == mHisData.end())
+            throw Exception(__PRETTY_FUNCTION__, "Cell data not found! - Did you initialize the history data?");
+        return (*ipVecIt).second[ipNum];
+    }
+
+    //! @brief Initializes the history data
+    //! @param cells: Cellgroup that is related to the integrand
+    //! @param dof: Dof type
+    void InitializeHistoryData(const Groups::Group<CellInterface>& cells, DofType dof)
```
TTitscher: I do not like the coupling to `Group`, `Cell` and `DofType` here. This is supposed to be some kind of a 2D container of `T`s. Why do we need those? Sure, for initialization. You can provide the sizes in a `map<int, int>` or something. To what value is everything initialized?

test/mechanics/integrands/HistoryData.cpp
```diff
@@ -0,0 +1,397 @@
...
+using namespace NuTo::Groups;
+
+
+constexpr double SpecimenLength = 10.;
+constexpr unsigned int numElements = 4;
+
+//! @brief Standard history data management object.
+//! @tparam T: Data type of the history data
+template <typename T>
+class HistoryData
```
TTitscher: IMO this class provides no benefit compared to a `map<int, vector<T>>` written directly at the law. It just provides access and a strange initialization. The main difference may be that it throws a `NuTo::Exception` instead of a `std::out_of_range`.

test/mechanics/integrands/HistoryData.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+    CreepHistoryData()
+    {
+        prevStrain[0] = 0.;
+        prevStress[0] = 0.;
+    }
+};
+
+//! @brief Creep law using the exponential algorithm
+class CreepLaw : public Laws::MechanicsInterface<1>, public HistoryData<CreepHistoryData>
```
TTitscher: I would prefer composition over inheritance. But this is no `Core` code...

test/mechanics/integrands/HistoryData.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+// %%% Mesh generation %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+MeshFem Mesh1D()
+{
+    MeshFem mesh;
+    const InterpolationSimple& interpolation = mesh.CreateInterpolation(InterpolationTrussLinear(1));
+    NodeSimple* nr = nullptr;
+    constexpr unsigned int numNodes = numElements + 1;
+    constexpr double elementLength = SpecimenLength / numElements;
```
TTitscher: I would be nice if you could move this to `UnitMeshFem::CreateTrusses(int numElements)` for everyone to access.

TTitscher (changes requested):
__Edit:__ Oh man, I completely forgot: Thanks for tackling this issue! Big step towards PDE! :ghost:

0. Github provides a nice interface to view code and changes to it, so there is IMO no need to put that much code in the issue text. 
1. It is hard to review, if basic design decisions are not made. Why not further discuss that in an issue instead of a pull request? Anyways, my opinion:
   - `NuTo::UniqueId` --> custom id in ctor (already on the todo list)
   - ~All cells of an integrand have the same number of integration points~. But it is only relevant for the history data container and ...
   - ... the history data container is user code. No discussion on design/initialization needed right now.
2. You provided a test of a creep law including a custom time integration. Sure, if something is wrong with the history data, this test should fail. But why? Wrong cell id? Wrong integraiton point id? Uninitialized data? Why not evaluate the cell with id = 7 with a function like
~~~cpp
void CheckCellId(celldata, cellipdata)
{
    BOOST_CHECK_EQUAL(celldata.Id(), 7);
    BOOST_CHECK_EQUAL(cellipdata.Id(), mIpCounter++);
}
~~~
or something? It is clear what is supposed to happen with those ids. Your test seems like an example of how to use PDE with history data. Btw: Why not perfect plasticity - something everyone knows?


I would like to see:
- remove `numIp` from `CellIpData`
- a test - possibly in `Cell.cpp` - that checks the correct passing of ids to the functions
- renaming of `HistoryData.cpp` to something that could include `creep` and `example` :)

vhirtham:
> Github provides a nice interface to view code and changes to it, so there is IMO no need to put that much code in the issue text.

Don't know this feature yet


> All cells of an integrand have the same number of integration points.

And that's assured somewhere in the code?

> But it is only relevant for the history data container and  the history data container is user code. No discussion on design/initialization needed right now.

I disagree. If you want to write your custom history data, you still have to initialize it somehow and it would be nice if there is some easy way to do this instead of hacking yourself through the code. So in my opinion, a simple way to initialize your (custom) history data is part of the topic

    

> You provided a test of a creep law including a custom time integration
...
Your test seems like an example of how to use PDE with history data. Btw: Why not perfect plasticity - something everyone knows?

I told you before, that this is just an intermediate solution to have at least some running code to discuss implementation details and quickly see the effects of changes on a full calculation. Yes it will be moved to the examples, yes there will be a "real" unit test instead of an implementation test and creep was the history data related law I am most familiar with. 

> would like to see:
    1. remove numIp from CellIpData
    2. a test - possibly in Cell.cpp - that checks the correct passing of ids to the functions
    3. renaming of HistoryData.cpp to something that could include creep and example :)

1. alternative to get the Ipnum to the law?
2. I can add this to the todo
3. See answer above

src/mechanics/cell/CellData.h
vhirtham: I ll try an alternative in the next push

src/mechanics/cell/CellIpData.h
vhirtham: 1. my apologies
2. probably better

src/mechanics/constitutive/laws/MechanicsInterface.h
vhirtham: will change that

src/mechanics/cell/Cell.h
```diff
@@ -35,14 +36,14 @@ class Cell : public CellInterface
...
-        CellData cellData(mElements);
+        CellData cellData(mElements, mIntegrationType.GetNumIntegrationPoints(), Id());
         for (int iIP = 0; iIP < mIntegrationType.GetNumIntegrationPoints(); ++iIP)
         {
             auto ipCoords = mIntegrationType.GetLocalIntegrationPointCoordinates(iIP);
             Jacobian jacobian(mElements.CoordinateElement().ExtractNodeValues(),
                               mElements.CoordinateElement().GetDerivativeShapeFunctions(ipCoords),
                               mElements.CoordinateElement().GetDofDimension());
-            CellIpData cellipData(mElements, jacobian, ipCoords);
+            CellIpData cellipData(mElements, jacobian, ipCoords, iIP);
```
joergfunger, line 46: IMO, the cell should not store an id, but the start of history data entry in the integrand container, thus HistoryData[currentIP_in_Integrand] = HistoryData[cellStart+ip_in_cell]. Thus, this can be directly stored in the CellIpData (no need to put it in CellData).

test/mechanics/integrands/HistoryData.cpp
vhirtham: the whole point is to discuss a default "can use but don't need to" class for a standard user, so he does not have to care about history data management (suggested by joergfunger). I know that it is not super, duper nice yet. It is a base for discussion. We can also discuss, if we even wan't to provide such a class.

src/mechanics/cell/Cell.h
```diff
@@ -60,15 +61,15 @@ class Cell : public CellInterface
...
+        CellData cellData(mElements, mIntegrationType.GetNumIntegrationPoints(), Id());
         for (int iIP = 0; iIP < mIntegrationType.GetNumIntegrationPoints(); ++iIP)
         {
             auto ipCoords = mIntegrationType.GetLocalIntegrationPointCoordinates(iIP);
             auto ipWeight = mIntegrationType.GetIntegrationPointWeight(iIP);
             Jacobian jacobian(mElements.CoordinateElement().ExtractNodeValues(),
                               mElements.CoordinateElement().GetDerivativeShapeFunctions(ipCoords),
                               mElements.CoordinateElement().GetDofDimension());
-            CellIpData cellipData(mElements, jacobian, ipCoords);
+            CellIpData cellipData(mElements, jacobian, ipCoords, iIP);
```
joergfunger, line 77: See above

test/mechanics/integrands/HistoryData.cpp
vhirtham: Doesn't work so well with inheritance. What would you expect, what the  function call `
mylaw(celldata, cellipdata);`will return? If we don't use inheritage it is an option.

src/mechanics/cell/CellData.h
joergfunger: I agree that we do not want the number of integration points being stored at the CellData. However, I would not allow the user to initialize the history data. I would rather have an additional class that comprises the cellgroup and the corresponding integrand. This container could the also initialize the history data. I would also suggest to move this handling at least for the standard case outside of the integrand with a template class that has a vector of history data, and the reallocation, access etc. is generally performed in this class.

src/mechanics/cell/CellIpData.h
joergfunger: See above, I would add here the direct entry in the vector of history data, not the relative one related to a single cell.

test/mechanics/integrands/HistoryData.cpp
vhirtham: We can also just use a single int for the size, if we can decide on a basic data type like a std::vector. The conditions for that are mentioned already. 
With the map<int,int> and the current underlying container the user needs to loop over all cells to get their ids and number of IPs.

TTitscher:
I just want this feature done asap :) And I think you put too much effort into not so important stuff. 

IMO we _only_ need the cellid/ipid. It may be helpful to have an example implementation of an integrand with history data, (personally, I think not), but in the end, everyone is going to write his own one. (Or will use some of the future `IntegrandLib`)

And a custom history data container is no magic. I would pick a `std::vector` and initialize it with 
~~~cpp
std::vector<MyData> mData(numCells * integrationType.GetNumIps(), MyData{0,0,0})
~~~
For me, that is simple enough.

And even if I have a varying number of integration points, I can use
~~~cpp
// some integrand
std::vector<std::vector<MyData>> mData;
// main
myIntegrand.mData.resize(numCells);
for (auto& cell : cells)
   myIntegrand.mData[cell.Id()].resize(cell.IntegrationType().GetNumIps(), MyData{0,0,0});
~~~

So maybe there is not even the need to have the same integration type for all cells. Point is, in my main file, I know what I do and I chose my history data accordingly. 

---

> alternative to get the Ipnum to the law?

I mean remove the `numIp`, not the `ipId`.

test/mechanics/integrands/HistoryData.cpp
vhirtham: Can do that.

test/mechanics/integrands/HistoryData.cpp
joergfunger: I disagree as already stated before, I would put the history data in a different container that does not have to be accessed by the user (if he wants to, he is free to do so, but not forced to).

test/mechanics/integrands/HistoryData.cpp
joergfunger: No need to provide these two int, one (the sum or cellNum*numIp+ipNum) should be fine. Then we would also not need a map, but a single vector is fine.

src/mechanics/constitutive/laws/MechanicsInterface.h
joergfunger: provide already the entry in the vector  cellNum*numIpPerCell+ipNum

vhirtham:
> I just want this feature done asap :) And I think you put too much effort into not so important stuff.

Well, writing this file was a big help for me to understand the whole stuff and how everything works together.


> IMO we only need the cellid/ipid. It may be helpful to have an example implementation of an integrand with history data, (personally, I think not), but in the end, everyone is going to write his own one. (Or will use some of the future IntegrandLib)

feel free to delete it once we are finished with this stuff

> And a custom history data container is no magic. 

No it's not. Still would be nice to have a default solution to import instead of always writing custom stuff. I think 90%+ of the cases this would be sufficient.


> I mean remove the numIp, not the ipId.

still not clear. Could whether mean rename` IPnum` to `ipId` or remove `numIntegrationPoints` from **!`cellData`** (you wrote cell**Ip**Data) - nevermind. I just do both

test/mechanics/integrands/HistoryData.cpp
vhirtham: Member is also okay, but than you have to to write a pass through for the Initialize function or a getter to the history data. I think inheritance is the easier way. YMMV

test/mechanics/integrands/HistoryData.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+public:
+    using typename Laws::MechanicsInterface<1>::MechanicsTangent;
+
+
+    //! @brief Ctor
+    //! @param E: Youngs modulus
+    //! @param E_KC: Kelvin chain stiffness
+    //! @param T_KC: Kelvin chain retardation time
+    CreepLaw(double E, VectorXd E_KC, VectorXd T_KC)
```
joergfunger: For a first pull request an a discussion of the idea, I guess that a creep law with some time integration is not the optimal example. A 1D-plasticity model is very simple and easier to understand.

test/mechanics/integrands/HistoryData.cpp
```diff
@@ -0,0 +1,397 @@
...
+    const Vector2d E_KC(60000., 120000.);
+    const Vector2d D_KC(1., 2.);
+    CreepLaw creepLaw(E, E_KC, D_KC);
+
+
+    // Create integrand %%%%%%%%%%%%%%%%%%%%%%%%%
+    Integrands::MomentumBalance<1> momentumBalance(displ, creepLaw);
+    double delta_t = 0.1;
+    auto MomentumGradientF =
+            std::bind(&Integrands::MomentumBalance<1>::Gradient, momentumBalance, _1, _2, std::ref(delta_t));
```
joergfunger: Why is here the reference to delta t? This is the kind of function call that follows the patter
Set(t), Gradient() instead of Gradient(t). When evaluating the gradient, I would like to see the time (or delta time) directly be added as a parameter (i.e. via a lambda)

test/mechanics/integrands/HistoryData.cpp
```diff
@@ -0,0 +1,397 @@
...
+        }
+        if (numIter >= maxIter)
+        {
+            std::cout << residual.lpNorm<Infinity>() << std::endl;
+            std::cout << time << std::endl;
+            throw Exception(__PRETTY_FUNCTION__, "No convergence");
+        }
+        // Update history data %%%%%%%%%%%%%%%%%%
+        for (auto& cell : momentumBalanceCells)
+            cell.Apply(MomentumUpdateHistoryDataF);
```
joergfunger: Similar here, delta_t should be an input to the function.

joergfunger (changes requested):

src/mechanics/cell/Cell.h
TTitscher: I vote for keeping the ids separate. This allows data at the integrand that is available once per _cell_, not once per _ip_. I can think of the cell area in a regularized local damage model. The integrand would have two data members:
- `mHistoryData` - `numCells x numIpPerCell` (access via cellId _and_ ipId)
- `mCellArea` - `numCells` (access _only_ via cellId)

__Edit__: And it allows to have different integration types for the cells. I am not a fan of that, but technically, it is easily possible, if the user wishes to do so.

src/mechanics/cell/CellData.h
vhirtham: Okay, this cellGroup-Integrand container might do the trick with the initialization, but since it is more like an own topic, should we for now stick to the manual initialization and postpone the automated initialization if we have this kind of class? I mean, I have a clear picture how it would work for a standard history data container. However, I am not sure if we would add constraints to the custom versions and I think we don't want to add constraints because of default solutions.

src/mechanics/cell/Cell.h
vhirtham: I also think, providing a single number is kind of an unwanted constraint. The user now has to use a data type which takes a single number as key.

test/mechanics/integrands/HistoryData.cpp
vhirtham: As I already told TTitscher: for me it was more than just writing a test. It was about understanding PDE and I am much more familiar to creep than to plasticity. In the OP, I removed all the creep related stuff and only left the things that are relevant to history data. It is also my intention to move this  to the examples later. I am not as familiar with the PDE implementations as you guys. So I wasn't able to write a minimal test. But as soon as I am, I ll do that.

test/mechanics/integrands/HistoryData.cpp
vhirtham: Well the answer is simple: 
Without the ref, every change in delta_t (dynamic timestepping) would not be passed to `MomentumGradientF`. So I would have to call std::bind again with the updated value.
The whole line
~~~cpp
 auto MomentumGradientF = std::bind(&Integrands::MomentumBalance<1>::Gradient, momentumBalance, _1, _2, std::ref(delta_t));
~~~

can also be written as

~~~cpp
auto MomentumGradientF = [&momentumBalance, &delta_t](cellData cd, cellIPData cipd){
    return momentumBalance.Gradient(cd,cipd,delta_t);
};
~~~

When we call `assembler.BuildVector` it expects a function which only has cellData and CellIPData as parameters, but no delta_t. So we need a lambda function or a std::bind to squeeze the gradient function of the law (which needs a delta_t) through the interface. As I said, if you remove the ref, you have to rebind the function every time delta_t changes. It is like removing the & in front of delta_t in the lambda capture.

Apart from that, I think this is not important to the history data itself. Maybe it should be discussed in another issue, if this stuff is going to be part of any core time integration scheme. In an example or test it shouldn't matter

test/mechanics/integrands/HistoryData.cpp
vhirtham: See above. Our generalized cell is not interested in a delta_t. We can change that if that's what everybody wants, but this topic is not related to history data.

src/mechanics/cell/CellData.h
vhirtham: Removed the numIntegration points. The value is (temporarily) passed directly to Initialize history data

test/mechanics/integrands/HistoryData.cpp
TTitscher: I thought about that as well. 
This is like a combination of [currying](https://en.wikipedia.org/wiki/Currying) and [partial application](https://en.wikipedia.org/wiki/Partial_application). Your gradient function takes 4 arguments. You want to transform that into a function, that takes `delta_t` and returns a function that takes `celldata, cellipdata` and partially applies the integrand. I solved that with a lambda that returns another lambda. It is a bit messy but can even be solved with templates and put somewhere in NuTo. 
I think I even put a snippet of that into our git.bam.de (no access atm...)

@vhirtham I think you are right that this may be a different topic. Ehr. But now I already typed it... :)

src/mechanics/cell/Cell.h
joergfunger: IMO should the integrand not be aware of the underlying cell structure, the area e.g. is no history data, but rather cell data (that could be recalculated each time from the CellData). The user just provides an interface for his integrand with a certain number of history data - that's all the standard integrand writing user should deal with.

vhirtham: Well if we want to hide the ID from the user (he can't set an ID himself) this discussion might be relevant. If the user can set the cell ID himself (via ctor), he can decide which approach he favors ("ID" or start in container) by picking his IDs accordingly. Since he his also able to use his own history data approach he can do whatever he likes.

src/mechanics/cell/CellIpData.h
vhirtham: Just another remark:

My git blame tells me, that the line above the marked line is from some TTitscher. Please remind him of point 1. `mIPCoords` --> `mIpCoords` :stuck_out_tongue:

TTitscher: Hmm, who should remind him? Ill make an issue :smile_cat:

src/mechanics/cell/Cell.h
TTitscher: @joergfunger What is the benefit of this single id? The only one I see is not storing an `int` in the `CellData` object. But this is no problem.
Everything else, like passing the two ids to a law, is somehow part of the integrand. @vhirtham chose for his law to get two ids. We can also design a default momentum balance integrand that calculates a single continuous id and passes that to the `MechanicsInterface`. 
But I see no reason in limiting _all_ integrands to that.

test/mechanics/integrands/HistoryData.cpp
```diff
@@ -42,9 +43,9 @@ class HistoryDataContiguousMemory
     //! @param cellNum: Number of the cell containing the integration point
     //! @param ipNum: Number of the integration point
     //! @return History data of the integration point
-    const T& GetIPHistoryData(const unsigned int cellNum, const unsigned int ipNum) const
+    const T& GetIpHistoryData(const unsigned int cellNum, const unsigned int ipNum) const
     {
-        assert(cellNum < mHistoryData.size() / mIpsPerCell);
+        assert(cellNum < mHistoryData.size() / mIpsPerCell && "Have you initialized the history data?");
```
Psirus: Shouldn't these be exceptions?

src/mechanics/cell/Cell.h
```diff
@@ -78,5 +84,6 @@ class Cell : public CellInterface, UniqueId<Cell>
 private:
     const ElementCollection& mElements;
     const IntegrationTypeBase& mIntegrationType;
+    const int mId;
```
TTitscher, line 86: I would generally not use `const` here. This prevents some compiler generated trivial copy/move stuff [How would you manually write a copy assignment?]. If you keep this data private, no one should be able to modify it from outside. The other data members remove the trivial assignment methods anyways, so it does not matter much here, but I find it interesting. Some more on this topic [here](http://www.drdobbs.com/the-problem-with-const-data-members/184403306).

src/mechanics/mesh/UnitMeshFem.h
```diff
@@ -6,6 +6,11 @@ namespace NuTo
 namespace UnitMeshFem
 {
 
+//! @brief creates a 1 dimensional mesh from (0) -- (1) with numX divisions
+//! @param numX number of divisions in x direction
+//! @return created mesh
+MeshFem CreateTrusses(int numX);
```
TTitscher: I would really like to see a simple test every new method. Have a look at the [existing ones](https://github.com/nutofem/nuto/blob/PDE_reviewed/test/mechanics/mesh/UnitMeshFem.cpp). Nothing fancy. You could create a mesh out of 2 elements and check if `numNodes == 3` and if there are nodes at `0, 0.5 and 1`. (no exception of `mesh.NodeAtCoordinate(0.5)`. IMO that is enough.

src/mechanics/cell/Cell.h
vhirtham: You might be right, that const is not necessary if the data is not exposed. On the other hand, do we want to copy assign something with an ID? Afterwards you have two objects with the same ID. Do we want that?

src/mechanics/mesh/UnitMeshFem.h
vhirtham: Was already on my list, just give me some time xD

test/mechanics/integrands/HistoryData.cpp
vhirtham: I thought about it, wasn't sure. What does anybody else think about it?

src/mechanics/cell/Cell.h
TTitscher: You are totally right. Maybe we should even explicitly delete copy/move ctor/assign. I just wanted to write it here, because this _make it as `const` as possible_ is not always the best thing to do.

src/mechanics/mesh/UnitMeshFem.h
TTitscher: Ideally, you write the test first :stuck_out_tongue:

test/mechanics/integrands/HistoryData.cpp
TTitscher: Can we delay everything regarding these containers or the integrands and move the discussion to an `IntegrandLib` issue? (--> #146)

test/mechanics/mesh/UnitMeshFem.cpp
```diff
@@ -28,6 +28,36 @@ void Check2DMesh(NuTo::MeshFem& mesh)
     BOOST_CHECK_NO_THROW(transformedMesh.NodeAtCoordinate(Eigen::Vector2d(43., 9.)));
 }
 
+BOOST_AUTO_TEST_CASE(MeshTrusses)
+{
+    auto mesh = NuTo::UnitMeshFem::CreateTrusses(13);
+    BOOST_CHECK_EQUAL(mesh.Elements.Size(), 13);
+    BOOST_CHECK_EQUAL(mesh.Nodes.Size(), 14);
+
+    bool nodesHaveExpectedValue = true;
```
TTitscher: Why not check the condition directly without this `bool`?
~~~cpp
for(...) BOOST_CHECK(...);
~~~

test/mechanics/mesh/UnitMeshFem.cpp
```diff
@@ -28,6 +28,36 @@ void Check2DMesh(NuTo::MeshFem& mesh)
...
+    for (unsigned int i = 0; i < mesh.Nodes.Size(); ++i)
```
TTitscher: Your test code is _very_ similar to the actual implementation. So if you have a bug in the implementation, chances are that you have the same bug in the test. This may not be a problem in this simple case, but generally I prefer testing against hardcoded values.

test/mechanics/mesh/UnitMeshFem.cpp
```diff
@@ -28,6 +28,36 @@ void Check2DMesh(NuTo::MeshFem& mesh)
...
+    {
+        if (std::abs(mesh.Nodes[i].GetValues()[0] - static_cast<double>(i) / mesh.Elements.Size()) > 1e-6)
```
TTitscher: You rely on `operator[]` and that the nodes are somehow in the same order as you created them. I would not require that... 
.

test/mechanics/mesh/UnitMeshFem.cpp
```diff
@@ -28,6 +28,36 @@ void Check2DMesh(NuTo::MeshFem& mesh)
...
+        {
+            nodesHaveExpectedValue = false;
+            break;
+        }
+    }
+    BOOST_CHECK(nodesHaveExpectedValue);
+
+    bool correctNodeOrder = true;
+    for (auto& element : mesh.Elements)
```
TTitscher: I like that you check the elements as well. I forgot that in the other tests in this file. But what is the actual requirement? I'd say that it is not a specific ordering, but a positive determinant of the jacobian.

src/mechanics/cell/Cell.h
vhirtham: I think moving is fine and should be allowed. However, since there are also references stored in the class, assignments couldn't be done anyways without also changing them to pointers. So if nobody has any complains, I would delete the copy ctor, and both assignment operators for the cell.

test/mechanics/mesh/UnitMeshFem.cpp
vhirtham: Because if you have messed something up, you get as many errors as you have nodes (for the same reason). Maybe there is a smart boost-way to give only one error message for checking members of an array, but I don't know it.

test/mechanics/mesh/UnitMeshFem.cpp
vhirtham: If you are checking 2d or 3d elements, checking J is probably smarter. However, for a linear 1d element there are only 2 ways to store the nodes, one that gives you a negative J and one that gives you a positive J. Calculating J requires an interpolation, which we don't have at this point, right?

test/mechanics/mesh/UnitMeshFem.cpp
vhirtham: I could also check if the coordinate is dividable by `1 / mesh.Element.Size()` without a rest, if that is more to your liking. That does not require an order.

test/mechanics/mesh/UnitMeshFem.cpp
vhirtham: see below

test/mechanics/mesh/UnitMeshFem.cpp
TTitscher: With this bool variable, don't know how many nodes are wrong and you don't know which nodes are wrong. I think this would be useful information for a test...

TTitscher (approved):

vhirtham:
At this point I think we are somehow done. All you need for history data is the cellId and the ipId at your integrand, which is working now. All the other stuff like design of a default history data class, initializing and updating can be discussed later. 
The important parts to use history data are there. How the user uses them is currently 100% his responsibility. 

I will move the current test to the PDE examples and write an own unit test which basically just checks if the data is passed to the integrand correctly. Afterwards, I think we can merge, if there are no bigger issues left.

src/mechanics/cell/Cell.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "base/UniqueId.h"
```
Psirus: No longer needed, now that you supply the id via the constructor.

src/mechanics/mesh/UnitMeshFem.h
Psirus: What do you think about a rename? "Trusses" already suggests mechanics again, so I'd prefer something with "line" or "interval" instead.

Psirus (approved):

TTitscher (approved):
[optional] I like the new test. Maybe we can even simplify it more. By leaving out constraints and the assembler and just evaluate a single cell. I can do that, if you wish. I mean, I would really be happy to help.

vhirtham:
Do it, if you like. You have a better overview than me. Haven't looked to much into the details of the assembler. If it does nothing that is involved in the procedure of passing the required values, than it should be fine to skip this part.

@Psirus: Cant answer directly somehow. Regarding the renaming, I think lines is okay.

src/mechanics/cell/Cell.h
vhirtham: Good point. I ll remove it

vhirtham:
So what does everybody think? Are we ready to merge or is there something missing?

joergfunger (approved):

PR #149, opened 2017-12-14
Pde remove custom benchmark

TTitscher:
Transforms most of our benchmarks to google benchmark. 

- removes solver benchmark because we currently don't have solvers :disappointed: 
- hidden feature: remove our custom `install_cmake.sh`, since travis provides a sufficient version after an [update](https://docs.travis-ci.com/user/build-environment-updates/2017-12-12/)


applications/benchmark/BuildGradient.cpp
```diff
@@ -1,111 +1,15 @@
...
 #include "mechanics/elements/ElementFem.h"
 #include "mechanics/cell/Cell.h"
 #include "mechanics/interpolation/InterpolationQuadSerendipity.h"
 #include "mechanics/integrands/MomentumBalance.h"
-#include "mechanics/constitutive/laws/LinearElastic.h"
-#include "mechanics/sections/SectionPlane.h"
+#include "mechanics/constitutive/LinearElastic.h"
 
 namespace Benchmark
 {
```
vhirtham: I find it quiet confusing to use the namespace "benchmark". Whenever I read a line starting with `Benchmark::` I think of a library feature until I see that you introduced this namespace. Maybe you should use a fixture here anyways. Then you have everything included into your benchmark and also don't need to write down the namespace.

applications/benchmark/BuildGradient.cpp
```diff
@@ -373,17 +182,22 @@ BENCHMARK(BuildGradient, NuToPDE, runner)
...
+    auto Gradient = [&](const NuTo::CellData& cellData, const NuTo::CellIpData& cellIpData) {
+        return integrand.Gradient(cellData, cellIpData, 0);
+    };
+
+    for (auto _ : state)
+        cell.Integrate(Gradient);
 }
+BENCHMARK(NuToPde);
+
+BENCHMARK_MAIN();
```
vhirtham: `BENCHMARK_MAIN()` does not need a semicolon. My compiler gives me a pedantic warning here.

vhirtham (commented):
I think you should add some documentation to the benchmarks. Comments and/or Doxygen tags. It would be nice to know what exactly is benchmarked without the need to look into the code or try to guess it by the name. Especially if the code is a little bit longer it isn't necessarily obvious what exactly happens.

I ll continue tomorrow. It's getting late :wink:

applications/benchmark/BuildGradient.cpp
```diff
@@ -121,24 +124,23 @@ class HardCodeElement8N
     }
 
 private:
-    std::vector<Node<2>*> mNodes;
+    std::vector<FixNode<2>*> mNodes;
```
vhirtham, line 127: This has nothing to do with the pullrequest itself, but why not use a std::array here? The size is fixed and won't change. Not sure if there will be an additional performance gain (probably not) - by the way, nice speed up in general :wink:

applications/benchmark/MeshFemBenchmark.cpp
```diff
@@ -1,73 +1,52 @@
...
-    NuTo::MeshFem mesh = NuTo::UnitMeshFem::CreateTriangles(1000, 1000);
-    while (runner.KeepRunningTime(1))
+    auto f = [](Eigen::VectorXd oldCoords) -> Eigen::VectorXd {
+        return Eigen::Vector2d(oldCoords[0] * 12 + 1, oldCoords[1] * 4 - 6174);
+    };
+
+    NuTo::MeshFem mesh = NuTo::UnitMeshFem::CreateTriangles(n, n);
+    for (auto _ : state)
+    {
         NuTo::MeshFem newMesh = NuTo::UnitMeshFem::Transform(std::move(mesh), f);
```
vhirtham: The results of this benchmark are somehow weird. The first iterations are constant time (just a few nanoseconds) and the last one suddenly explodes. It seems like there are some compiler optimizations involved. The complexity calculation makes no sense in this case.

vhirtham (commented):
The shape function memorization benchmark is a little bit messy. One memorizer is located at the top, one at the bottom. Maybe you can shift the declarations a bit, so that it easier to read from top to bottom. Additionally the filename is missing a "r" -> Memo-R-ization

Apart from that and the other comments, I think everything is okay. :+1:

Psirus:
https://en.wikipedia.org/wiki/Memoization

applications/benchmark/MeshFemBenchmark.cpp
TTitscher: Feel free to investigate the results. The benchmark still has a RMS error of only 2%, which, for me, is totally fine. Internally, I may be something like `0.01 N^2` (from Group::Add()) `+ O(1)` (something else)

vhirtham: The thing is: Whats the point about having a BigO benchmark if the first results are optimized out and are therefore fast while the last ones can't be optimized any more and deliver the "true" results. We are now mixing results of optimized and non optimized code when calculating BigO.  Here are my results:

~~~
Transform/16           14 ns         14 ns   47988208

Transform/32           14 ns         14 ns   48290600

Transform/64           15 ns         15 ns   47282502

Transform/128          15 ns         15 ns   46240131

Transform/256          15 ns         15 ns   39576618

Transform/512          28 ns         28 ns   20294803

Transform/1024 2887815884 ns 2887282402 ns          1

Transform_BigO       0.00 N^3       0.00 N^3 
Transform_RMS           4 %          4 % 
~~~

Even though the RMS is just 4%, have a look at the version with n=1024. I find the jump from 28ns to 2887815884ns a little bit weird. Checked it on 2 Computers (gcc/clang release). Maybe we can talk about it on Monday.

TTitscher: Yeah, seems suspicious. On my current machine (with clang 3.8) it is:
~~~
Transform/16           21 ns         21 ns   33408514
Transform/32           21 ns         21 ns   33455663
Transform/64           21 ns         21 ns   33265761
Transform/128          21 ns         21 ns   32113723
Transform/256          23 ns         23 ns   29502755
Transform/512   527130628 ns  527109910 ns          1
Transform/1024 7392408471 ns 7391243517 ns          1
Transform_BigO       0.01 N^2       0.01 N^2 
Transform_RMS           2 %          2 % 
~~~

Lets investigate this apart from this PR.

vhirtham: Okay. I ll give you the freedom to merge :stuck_out_tongue:

vhirtham (approved):

joergfunger (approved):
Semicolons after BENCHMARK_MAIN() where on my machine required (otherwise I got a compilation error on gcc version 7.2.1). The speedup of the memoizer is significant, but it comes with many issues - so having the benchmark is nice, but I would rethink about it when moving this to the standard.

TTitscher:
For me, this is the danger of these _micro_ benchmarks. On my machine, for example, the gain is about 1 or 2 microseconds. In an implicit simulation, this will have no impact. In a explicit calculation, it _may_. But maybe there is another part of our tool, where we can save 10 microseconds or 100 or 1000. So we should always keep that in mind, when benchmarking single functions.

PR #150, opened 2017-12-19
Pde enhance global dof types

TTitscher:
Brief summary:
- `GlobalDofVector::operator()(dofType, globalDofNumber)` allows access via the _global_ dof number
- provide `ToEigen` and `FromEigen` for import export
    - I chose to put them in separate files, so we can easily add something like `ToTpetra` or `ToPetsc` without including unused code. 

I made an error when I included 3422738. I noticed that too late to conveniently fix it. Please ignore :)


src/mechanics/dofs/DofVectorConvertEigen.h
```diff
@@ -0,0 +1,53 @@
...
+    }
+    return combined;
+}
+
+//! @brief imports a values into a properly sized DofVector
+//! @param rDestination properly sized dof vector
+//! @param source eigen vector whose values are imported
+//! @param dofs dof types to import
+template <typename T>
+inline void FromEigen(DofVector<T>& rDesination, const Eigen::Matrix<T, Eigen::Dynamic, 1>& source,
```
Psirus: Without a reference, my impression of most cases where the output is passed as a reference to the function, the output parameter often comes last, or at least after the input, as in `FromEigen(source, dofs, destination)`. Also, did you not previously prefer pointers in such cases to make it obvious, as in `FromEigen(source, dofs, &destination)`?

Psirus (approved):

PR #152, opened 2017-12-19
Pde spectral elements

pmueller2:
Implemented spectral elements based on legendre polynomials (for use with Lobatto integration) with variable order. But since there is currently a limit on the number of nodes per element (maxNumNodes = 125) this is not so variable after all (order 4 for a brick as in old NuTo).

The routine to evaluate the shape functions is based on the barycentric interpolation formula to prevent round off and have a O(N) complexity. For the derivatives this is not (yet) the case.


src/mechanics/interpolation/InterpolationBrickLobatto.h
```diff
@@ -0,0 +1,76 @@
...
+
+    std::unique_ptr<InterpolationSimple> Clone() const override
+    {
+        return std::make_unique<InterpolationBrickLobatto>(*this);
+    }
+
+    ShapeFunctions GetShapeFunctions(const NaturalCoords& naturalIpCoords) const override
+    {
+        Eigen::VectorXd result(GetNumNodes());
+        const std::vector<double> shapes = ShapeFunctions3D::ShapeFunctionsBrickLagrange(naturalIpCoords, mNodes);
```
TTitscher: I would prefer defining your `ShapeFunctions3D` to return `Eigen` types directly.

src/mechanics/elements/SpectralShapeFunctions.h
```diff
@@ -0,0 +1,248 @@
...
+}
+
+
+
+namespace ShapeFunctions3D
+{
+//! @brief Lobatto nodes for brick element (tensor product)
+//! @param order Polynomial order in 1D
+//! @return local node coordinates
+std::vector<Eigen::Vector3d> NodeCoordinatesBrickLobatto(int order)
```
TTitscher: Is it possible to derive a `TensorProduct` abstraction? Ideally something like
~~~cpp
using NodeCoordinatesBrickLobatto = TensorProduct<3>(NodeCoordinatesTrussLobatto);
~~~
Similar for Quad and the other methods.

src/mechanics/elements/SpectralShapeFunctions.h
```diff
@@ -0,0 +1,248 @@
...
+//! @param nodes local node coordinates
+//! @return shapes evaluated at x. Fulfill interpolation condition fi(xj) = delta_ij
+std::vector<double> ShapeFunctionsBrickLagrange(const Eigen::Vector3d x, const std::vector<double>& nodes)
+{
+    std::vector<double> Nx = ShapeFunctions1D::ShapeFunctionsTrussLagrange(x(0), nodes);
+    std::vector<double> Ny = ShapeFunctions1D::ShapeFunctionsTrussLagrange(x(1), nodes);
+    std::vector<double> Nz = ShapeFunctions1D::ShapeFunctionsTrussLagrange(x(2), nodes);
+
+    std::vector<double> result;
+    for (size_t k=0; k<nodes.size(); k++)
```
TTitscher: Here, the loop order is `k -- j --i`, above it is `i -- j -- k`. Is that important?

pmueller2: Here (Shapes) the order is important (k-j-i), above (NodeCoordinates) it doesnt matter - but I think I will change it also to k-j-i for better reading.

src/mechanics/elements/SpectralShapeFunctions.h
pmueller2: This is probably a good idea. We talked about this when implementing the tensor product based integrationtypes and there such an abstraction would also be beneficial.

src/mechanics/interpolation/InterpolationQuadLobatto.h
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    int GetDofDimension() const override
+    {
+        return mDofDimension;
+    }
+
+private:
+    int mDofDimension;
+    std::vector<double> mNodes;
```
TTitscher: Why do you store the this vector here? 

The methods like `ShapeFunctions` are generally _slow_, because they do actual calculations instead of a lookup. This will be solved with this memoization technique. So we can at least make the objects as small as possible. So I would "trade" the storage vs some more calls to `NodeCoordinatesTrussLobatto`. (Same goes for Truss/Brick)

__Edit:__ Why make objects small? I think of storing the the interpolation as a _value_ at each element. This makes copying the element much easier. So the interpolation is not stored once, but once per element.

test/mechanics/interpolation/InterpolationBrickLobatto.cpp
```diff
@@ -0,0 +1,34 @@
...
+}
+
+BOOST_AUTO_TEST_CASE(InterpolationBrickLobatto)
+{
+    int order = 4;
+    NuTo::InterpolationBrickLobatto interpolation(1, order);
+
+    std::vector<Eigen::VectorXd> rPoints = GetTestPoints();
+
+    BOOST_TEST_MESSAGE("Checking copy move...");
```
TTitscher: You can try to avoid the code duplication. There is this `InterpolationTests.h`. Why not refactor this duplicate code (in Quad, Truss, Brick) into a `InterpolationLobattoTests.h`?

test/mechanics/interpolation/InterpolationQuadLobatto.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+std::vector<Eigen::VectorXd> GetTestPoints()
+{
+    constexpr double a = 0.577350269189626;
+    return {Eigen::Vector2d({-a, -a}), Eigen::Vector2d({-a, a}), Eigen::Vector2d({a, a}), Eigen::Vector2d({a, -a})};
+}
+
+BOOST_AUTO_TEST_CASE(InterpolationQuadLobatto)
+{
+    int order = 5;
```
TTitscher, line 13: Why _only_ `order = 5`? What happens for `order = 0`? What for `order = 1,2,3,4,6,7,8,9, ...`? How many orders make sense?

test/mechanics/interpolation/InterpolationTrussLobatto.cpp
```diff
@@ -0,0 +1,25 @@
+#include "BoostUnitTest.h"
+#include "InterpolationTests.h"
+#include "mechanics/interpolation/InterpolationTrussLobatto.h"
+
+std::vector<Eigen::VectorXd> GetTestPoints()
+{
+    return {Eigen::VectorXd::Ones(1) * (-1.), Eigen::VectorXd::Ones(1) * (0.), Eigen::VectorXd::Ones(1) * (1.)};
```
TTitscher: There is `Eigen::VectorXd::Constant(size, value)` that is more appropriate here. And `Eigen::VectorXd::Ones(1) * 0` looks like `Eigen::VectorXd::Zero(1)`. And `Eigen::VectorXd::Ones(1) * 1`, well.

TTitscher (changes requested):
I like to request some changes apart from my comments. Back in the days, we collected all shape functions in `ElementShapeFunctions`. First I thought that this is no longer needed and you should write the content of your shape function file directly into the interpolations. But since the code still undergoes severe changes, it may be a good idea to keep it separated. But, I would like you to remove the old lobatto code from `ElementShapeFunctions`. 

Another change concerns the tensor products. Can you think of a way to implement that in a generic way?
- `Quad = TensorProduct(Truss, Truss)`
- `Prism = TensorProduct(Triangle, Truss)`

or more like
- `Quad::NodeCoords = TensorProduct(Truss::NodeCoords, Truss::NodeCoords)`

And it can be applied for the integration types as well. It may be tricky for derivative shape functions though. If this is too complicated, this is not requested for the review. Then you could open an issue on that.

src/mechanics/elements/SpectralShapeFunctions.h
```diff
@@ -84,18 +90,18 @@ std::vector<double> ShapeFunctionsTrussLagrange(const double x, const std::vecto
...
 {
-    std::vector<double> result(nodes.size(), 0.);
-    std::vector<double> w = BarycentricWeights(nodes);
-    for (size_t j = 0; j < nodes.size(); j++)
+    Eigen::VectorXd result = Eigen::VectorXd::Zero(nodes.size());
+    Eigen::VectorXd w = BarycentricWeights(nodes);
+    for (int j = 0; j < nodes.size(); j++)
     {
-        for (size_t k = 0; k < nodes.size(); k++)
+        for (int k = 0; k < nodes.size(); k++)
```
TTitscher: One (hopefully) last request. Eigen::size() returns something like the size in memory. For a vector, this is equal to the number of its rows. Still, I find it _more_ correct, to use `.rows()` here and in the other places. This leaves no room for doubts.

TTitscher (approved):

PR #153, opened 2017-12-19
Pde remove sparse matrix

Psirus:
What it says on the tin. Continuing what @TTitscher started.


TTitscher (approved):
:+1: for `makeCompressed()`. I guess that was the reason for a similar bug I had in `PatchTest.cpp`.

Issue #142, opened 2017-12-05
PDE interpolation


pmueller2:
I think the interpolation should not know the DoF dimension. It is just not needed for the type of interpolation we use.

PR #154, opened 2017-12-20
Pde remove dof dimension

TTitscher:
Solves #142.

The only place where there the dof dimension is needed is the calculation of the `N` matrix. This dimension is now extracted from the first node of the element, namely the `NodeSimple::GetNumValues()`.


pmueller2 (approved):
Supi.

PR #155, opened 2017-12-21
Pde import from gmsh

pmueller2:
(Re)Implemented mesh import from Gmsh. Currently only ASCII msh files are supported. Element types implemented are: 
- LinearLine
- LinearTriangle
-  LinearQuad

The idea is to import all sorts of elements found in the mesh file into a `MeshFEM`. Subsets of this mesh (Groups of ElementCollection) can be accessed via their PhysicalGroupId or PhysicalGroupName.

**Design question**
1. How to test support for different element types?
We can either create a mesh for each element type or one mesh for each dimension and add an element of each type.


src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,375 @@
...
+    int dimension = 0;
+    std::string name;
+};
+
+
+struct GmshFileContent
+{
+    GmshHeader header;
+    int minNodeId = INT_MAX;
+    int maxNodeId = INT_MIN;
```
Psirus: Okay, as far as I can make out, `minNodeId` is used for an offset in a `std::vector`, so that if the IDs start at 5, that node is still at at `[0]` in the vector. But what to do you need `maxNodeId` for?

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,375 @@
...
+    int dimension = 1;
+    std::vector<GmshNode> nodes;
+    std::vector<GmshElement> elements;
+    std::vector<GmshPhysicalNames> physicalNames;
+};
+
+// Helper functions (cpp only)
+// %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+static GmshHeader ReadGmshHeader(std::ifstream& file)
```
Psirus: What does this `static` do?

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+{
+    std::string line;
+    getline(file, line);
```
Psirus: Sometimes you use `getline`, and sometimes `std::getline`. Is this intentional?

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+    {
+        throw NuTo::Exception(__PRETTY_FUNCTION__, "Gmsh version 2.0 or higher requiered. - File version is " +
+                                                           std::to_string(header.version));
+    }
+    getline(file, line); // $EndMeshFormat
+    return header;
+}
+
+
+void ReadNodesASCII(std::istream& file, GmshFileContent& fileContent)
```
Psirus: I'd prefer this to be `std::pair<std::vector<GmshNodes>, double> ReadNodesASCII(std::istream& file)`, and the call then `std::tie(fileContent.nodes, fileContent.dimension) = ReadNodesASCII(file);`. This also gives you the opportunity to include `pair` and `utility`, helping you in your quest to include all standard headers in a single file 👻 

 Similar arguments could be made for the other functions.

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+}
+
+
+// Member functions
+// %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+NuTo::MeshGmsh::MeshGmsh(const std::string& fileName)
+{
+    ReadGmshFile(fileName);
+}
```
Psirus, line 229: Why not simply put the contents of `ReadGmshFile` here?

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+void NuTo::MeshGmsh::CreateElements(const GmshFileContent& fileContent,
+                                    const std::vector<NuTo::NodeSimple*>& nodePtrVec)
+{
+    std::map<int, const InterpolationSimple*> interpolationPtrMap;
+
+    for (GmshElement gmshElement : fileContent.elements)
+    {
+        // Skip node elements
+        if (gmshElement.type == 15)
+            continue;
```
Psirus, line 272: I think this is fine, for now, but I generally, could see a "node element" being useful. What does everyone else think?

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+}
+
+void NuTo::MeshGmsh::CreateMesh(const GmshFileContent& fileContent)
+{
+    std::vector<NuTo::NodeSimple*> nodePtrVec = CreateNodes(fileContent);
+    CreateElements(fileContent, nodePtrVec);
+}
+
+
+void NuTo::MeshGmsh::ReadGmshFile(const std::string& fileName)
```
Psirus: Nitpick: too many blank lines in this function.

Psirus (commented):
👍

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+{
+    std::ifstream file;
+    file.open(fileName, std::ios::in);
+
+
+    if (not file.is_open())
+    {
+        std::cout << fileName << std::endl;
+        throw Exception(__PRETTY_FUNCTION__, "Error opening input file "
```
Psirus: Inconsistent use of `NuTo::Exception` vs `Exception`.

Psirus:
Oh, and maybe a small test with a 3D mesh, just to be sure.

TTitscher:
> How to test support for different element types?

Those test should track down errors as quickly as possible. Thus, `ImportFromGmsh.cpp::ImportQuadCubic fails` is nicer than `ImportFromGmsh.cpp::2D fails`. Or, what about things like a Quad in 3D?. So I vote for separate mesh files for each type. Hopefully in a convenient way to test like:
~~~cpp
BOOST_SOMETHING_TEST_CASE(QuadCubic)
{
   NuTo::Test::CheckGmshImport("quad.msh");
}
~~~

Follow-up question: What is tested? Import does not throw? Element has positive `Det(Jac)` on some points?

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+    bool is3d = false;
+    for (GmshNode& node : fileContent.nodes)
+    {
+        file >> node.id;
+        file >> node.coordinates[0];
+        file >> node.coordinates[1];
+        file >> node.coordinates[2];
+        fileContent.minNodeId = std::min(fileContent.minNodeId, node.id);
+        fileContent.maxNodeId = std::max(fileContent.maxNodeId, node.id);
+        is2d = (is2d || node.coordinates[1] != 0);
```
TTitscher: Is it safe here to compare floats without a floating point tolerance? Apparently, your 2D test has no problems with that...

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+    std::getline(file, line);
+    if (line[0] == '$')
+    {
+        // remove $
+        std::string sectionName = line.substr(1, line.size() - 1);
+        // capitalize
+        std::transform(sectionName.begin(), sectionName.end(), sectionName.begin(), ::toupper);
+
+        // Execute section read function
+        if (sectionName.compare("NODES") == 0)
```
TTitscher, line 185: This `if else` things could make sense, if the order of these sections were arbitrary. But IMO they are not. So I would probably move the check for `section name = "NODES"` in the `ReadNodesASCII` method. Then, this method may simply look like
~~~cpp
ReadNodesASCII(...);
ReadElementsASCII(...);
ReadPhysicalNamesASCII(...);
~~~
or could be avoided at all.

src/mechanics/mesh/MeshGmsh.h
```diff
@@ -0,0 +1,93 @@
...
+    explicit MeshGmsh(const std::string& fileName);
+
+    //! @brief Gets an element collection group associated to the physical name defined in gmsh
+    //! @param physicalName: Group name (defined in gmsh)
+    //! @return Desired element collection group
+    const NuTo::Group<ElementCollection>& GetPhysicalGroup(std::string physicalName) const;
+
+
+    //! @brief Gets an element collection group associated to the physical ID defined in gmsh
+    //! @param physicalGroupId: gmsh ID
```
TTitscher, line 52: How can I know the gmsh ID for my group? Say, I call my physical group "LeftBoundary", can I reason about its id? If I call the physical group "42", will it have the gmsh ID `42`? And what if I call it `42a`?

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+            ReadElementsASCII(file, fileContent);
+        else if (sectionName.compare("PHYSICALNAMES") == 0)
+            ReadPhysicalNamesASCII(file, fileContent);
+        else
+            throw NuTo::Exception(__PRETTY_FUNCTION__, "Unhandled gmsh section type: " + sectionName);
+    }
+}
+
+
+const NuTo::InterpolationSimple& CreateElementInterpolation(NuTo::MeshFem& mesh, int gmshType, int dimension)
```
TTitscher: I would prefer having this method as a general factory that has nothing to do with the mesh. The actual creation of the interpolation _inside_ the mesh requires checking for duplicates and that stuff. You handle that below in `CreateElements`. So I would call `mesh.CreateInterpolation` there and remove `mesh` (and `dimension`) from the signature.

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+
+const NuTo::Group<NuTo::ElementCollection>& NuTo::MeshGmsh::GetPhysicalGroup(int physicalGroupId) const
+{
+    auto physGroupIt = mPhysicalGroups.find(physicalGroupId);
+    if (physGroupIt == mPhysicalGroups.end())
+        throw Exception(__PRETTY_FUNCTION__, "Couldn't find group with ID " + std::to_string(physicalGroupId));
+    return physGroupIt->second;
+}
+
+std::vector<NuTo::NodeSimple*> NuTo::MeshGmsh::CreateNodes(const GmshFileContent& fileContent)
```
TTitscher: I once said that `map<int, T>` smells like `vector<T>`. In this case, however, I would prefer the `map` for simplicity. You do not need to handle the `MINMAX_ID` stuff, no need for continuous gmsh node ids, (no memory allocation erros that you needed to fix - that shows that even for the original authors, the code is not simple enough to avoid the errors :) )
If you are worried about performance, consider a hashed map, flat map, flat hashed map or whatever. (And ideally prove your concerns with a benchmark.)
I'll point out some places (outside this method) that will probably be more readable by simply referring to this comment.

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+    int id = 0;
+    int dimension = 0;
+    std::string name;
+};
+
+
+struct GmshFileContent
+{
+    GmshHeader header;
+    int minNodeId = INT_MAX;
```
TTitscher: see `map` comment

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+        auto interpolationIter = interpolationPtrMap.find(gmshElement.type);
+        if (interpolationIter == interpolationPtrMap.end())
+            interpolationIter = interpolationPtrMap
+                                        .emplace(gmshElement.type, &CreateElementInterpolation(mMesh, gmshElement.type,
+                                                                                               fileContent.dimension))
+                                        .first;
+        std::vector<NodeSimple*> elementNodes(gmshElement.nodes.size());
+        for (unsigned int i = 0; i < elementNodes.size(); ++i)
+        {
+            int nodeVectorPos = gmshElement.nodes[i] - fileContent.minNodeId;
```
TTitscher: see `map` comment

src/mechanics/mesh/MeshGmsh.h
```diff
@@ -0,0 +1,93 @@
...
+#include <map>
+
+class GmshFileContent;
+
+namespace NuTo
+{
+
+//! @brief Reads a gmsh msh-file and converts it to a MeshFEM. Additionally physical groups are created and can be
+//! accessed via ID or name
+class MeshGmsh
```
TTitscher, line 17: I am not sure how to handle the import from an API point of view. I would prefer a free function that somehow returns the functionality you provided. So 
~~~cpp
auto gmshMesh = NuTo::ImportFromGmsh(filename);
MeshFem& mesh = gmshMesh.Mesh();
auto leftBoundary = gmshMesh.Get("LeftBoundary");
~~~
or something in that fassion.
The problem with that may be that the `MeshFem` is currently not copyable... :(

src/mechanics/mesh/MeshGmsh.h
```diff
@@ -0,0 +1,93 @@
...
+{
+public:
+    //! @brief ctor
+    //! @remark deleted: The class is always constructed with its associated files name
+    MeshGmsh() = delete;
```
TTitscher, line 22: I would remove those two
~~~cpp
MeshGmsh() = delete;
~MeshGmsh() = delete;
~~~
because I find them pointless. But this may be personal preference.

src/mechanics/mesh/MeshGmsh.h
```diff
@@ -0,0 +1,93 @@
...
+    //! @return Desired element collection group
+    const NuTo::Group<ElementCollection>& GetPhysicalGroup(int physicalGroupId) const;
+
+private:
+    //! @brief Adds an element to its corresponding physical group. If the element is part of named group and the group
+    //! does not exist, then a named physical group is also created.
+    //! @param fileContent: Special structure (see MeshGmsh.cpp) that holds the read file content
+    //! @param element: Current element
+    //! @param physicalGroupId: Id of the elements physical group
+    void AddElementToPhysicalGroup(const GmshFileContent& fileContent, ElementCollection& element, int physicalGroupId);
```
TTitscher: 1) I would avoid using return arguments as much as possible. This may not be possible here due to not copyable objects.
2) Please provide the `r` prefix for return arguments. So `rElement`. Same goes for all the other signatures with nonconst references.

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+        is2d = (is2d || node.coordinates[1] != 0);
+        is3d = (is3d || node.coordinates[2] != 0);
+        std::getline(file, line);
+    }
+    if (is2d)
+        fileContent.dimension = 2;
+    if (is3d)
+        fileContent.dimension = 3;
+
+    std::getline(file, line);
```
TTitscher: This and the following lines are duplicated in the methods below. Please refactor :)

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+
+    fileContent.physicalNames.resize(numNames);
+    for (GmshPhysicalNames& name : fileContent.physicalNames)
+    {
+        file >> name.dimension;
+        file >> name.id;
+        file >> name.name;
+
+        // remove quotation marks from string
+        name.name.erase(std::remove(name.name.begin(), name.name.end(), '\"'), name.name.end());
```
TTitscher: `name.name name.name name.name name.name` :)

src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -0,0 +1,378 @@
...
+void ReadPhysicalNamesASCII(std::istream& file, GmshFileContent& fileContent)
+{
+    std::string line;
+    std::getline(file, line);
+    int numNames = std::stoi(line);
+
+    fileContent.physicalNames.resize(numNames);
+    for (GmshPhysicalNames& name : fileContent.physicalNames)
+    {
+        file >> name.dimension;
```
TTitscher: what is the `dimension` of a `name`? Its length? And what is its id? Some continuous number in the vector of all possible names? :stuck_out_tongue:

TTitscher (changes requested):
I really see the limitation of not having copyable elements in this PR. You are forced into a wierd API and have to use return arguments a lot. Since there are plans to introduce this feature soon, there might come big refactoring to this method. 
Until then, I like - apart from some little comments... and the `map` thing, that IMO would _really_ remove a lot of index confusion.)

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: For the mapping during element creation. Gmsh does not guarantee contiguous or ordered IDs. The created `NodeSimple` doesn't know about its gmsh ID. So you can use a map with the ID and a pointer to the created node or a vector with pointers and an easily reproducible mapping.
Let's say you have 3 Nodes with the IDs 1,4,6 and use a densely populated vector with size 3. How do you know that ID 1 is at `[0]`, 4 at `[1]` and 6 at `[2]`. So what we do is creating an oversized vector with the size maxID-minID. In this case you get some vector content like this:

~~~
[0] -> Pointer to node with ID = 1
[1] -> nullptr
[2] -> nullptr
[3] -> Pointer to node with ID = 4
[4] -> nullptr
[5] -> Pointer to node with ID = 6
~~~

Benefit of this approach is that generation and lookup are constant time, while the map gets slower with growing N. Drawback is a potentially higher memory consumption.

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: whoops that should be gone... I ll remove that

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: nope. I will add the missing namespaces

src/mechanics/mesh/MeshGmsh.cpp
Psirus: Looking at the code, you set this variable once (l. 105), but never read it. So it can be removed imho.

TTitscher: __Another edit:__ You commented on @Psirus remark. I mixed that up, sorry.

I am aware of the implementation. I just want it to be simpler. Some options:
- use a `map`
- if you are concerned about performance, use an `unordered_map`
- if you are, for some reason, not happy with that performance - keep in mind that it only called once... - use a `vector`. But then, why you introduce this `minNodeId`? To save half a megabyte of memory we use once? This just makes the code uglier IMO. __Edit__: The simplification would be, that `gmshNodeId == vectorId`, no calculations involved.

src/mechanics/mesh/MeshGmsh.cpp
TTitscher: Just out of curiosity: Why does that also work without the namespace? Is there a call to a similarly named method outside `std`?

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: Sorry to disappoint you, but pair and utility are already included by other STL headers. :trollface:  - I remove the unnecessary STL headers. We forgot that during clean up.

I also fulfill your other wish :stuck_out_tongue:

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: Maybe we will provide additional ctors with additional options in the future. For example if you want just a single physical group of the gmsh file. But we can remove the extra function call if you want.

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: Well yes, since you are reading directly from a file here. If there is a 0 in the file, than your double would also be 0. If the node in the file has a non zero value, even with minimal distance to 0, you probably messed something up in gmsh or shiftet your plane/line to this value (which is currently not allowed). ;)
We make an assumption/prescription about the meshes dimension here. 1d is always  [x,0,0] and 2d is always [x,y,0]. We did this to avoid a lot of checks. But that musn't be our final approach. What do you think, should we be more flexible for 1d and 2d mesh generation?

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: Not sure if I got you right, ...but seriously?

You want to introduce an own function with two parameters (file and physical name) for those four lines
~~~cpp
    std::getline(file, line);
    std::transform(line.begin(), line.end(), line.begin(), ::toupper);
    if (line.compare("$ENDPHYSICALNAMES") != 0)
        throw NuTo::Exception(__PRETTY_FUNCTION__, "$EndPhysicalNames not found!");
~~~

that are duplicated in 3 funtions that are never used outside the cpp file? I think it is a little bit too much, to be honest. While the compiler doesn't care and inlines it anyways, it is sometimes a big PITA to read a function if you have to jump from function to function in order to understand what is actually happening. If the majority wants that, I will do it, but if nobody else cares, I wont.

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: change it to whatever you like. The origin of the naming:
http://gmsh.info/doc/texinfo/gmsh.html#MSH-ASCII-file-format

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: see above :trollface:

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: Have not checked it myself, but @pmueller2 said, that the new gmsh standard allows multiple duplications of some sections like nodes.

Psirus:
Sorry, wrong branch 😬

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: from the gmsh documentation:

> Important note about efficiency. Node and element tags can be "sparse", i.e., do not have to constitute a continuous list of indexes starting at 1. However, using non-continuous tags will lead to performance degradation. For meshes, non-continuous indexing forces Gmsh to use a map instead of a vector to access nodes and elements. The performance hit is on speed. For post-processing datasets, which always use vectors to access data, the performance hit is on memory. A NodeData with two nodes, tagged 1 and 1000000, will allocate a (mostly empty) vector of 1000000 elements. 

The current solution is constant time during lookup and when it is filled. Every other solution is not. The code isn't that much more complicated and when the gmsh class is working, nobody would probably ever look into the code anymore. Apart from that, the solution is part of gmsh too and I think they chose it for a reason ;) 

> no memory allocation erros that you needed to fix - that shows that even for the original authors, the code is not simple enough to avoid the errors

If the code needs to be so simple, that we never make any errors during development, we should stop writing NuTo and focus on `"HelloWorld"` examples. :wink:  - A cookie for the guy who gets an simple hello world example to fail due to memory access violation :trollface:

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: see `map` answer

src/mechanics/mesh/MeshGmsh.cpp
TTitscher: Nope, seems fine as it is then.

src/mechanics/mesh/MeshGmsh.h
vhirtham: I don't know. I don't see the benefit. Now you need a constructor to pass all the internal variables of MeshGmsh to the created mesh since the function can't access the internal variables of the mesh it returns until it is a friend function.

And... yeah, I just saw that we forgot to write a getter for the mesh...

src/mechanics/mesh/MeshGmsh.cpp
TTitscher: Code duplication is not about how many arguments or how many lines you save. Its about making it more maintainable and readable. Imagine this likes would be
~~~cpp
ReadNextExpectedLine(file, "$EndPhysicalNames");
~~~
I would prefer this. Even if it only occurs once.

src/mechanics/mesh/MeshGmsh.cpp
TTitscher: [`unordered_map`](http://en.cppreference.com/w/cpp/container/unordered_map/at#Complexity) has also constant lookup, just saying. 
Its fine, if you do not want to make the change. So, would you mind if I remove the `minNodeId` (and we still use vector)?

src/mechanics/mesh/MeshGmsh.h
vhirtham: I would advice against deleting default dtors :stuck_out_tongue: - change it if you like, but I prefer to explicitly write down the big 5 + dtor so I don't have to guess what the compiler is doing. But as you said preferences.

src/mechanics/mesh/MeshGmsh.h
vhirtham: :no_mouth: ... By just remembering what you actually did in gmsh? ... or try looking into your file! :stuck_out_tongue_winking_eye:  The thing is, that a group does not need a name, but it needs an ID that the user can (but don't need to) specify. Each element has exactly one physical group. We don't want to force the user to always use names. So he can use the IDs. But what ever he does, he has to remember what he did in gmsh and he needs to know difference in cpp between writing `"42"` and `42`. I hope it isn't too much to ask for. :laughing:

src/mechanics/mesh/MeshGmsh.h
vhirtham: Will see which kind of `std::tie` and return value magic we can do to change that. If it is not possible I will add the r prefix.

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: Why is it so important to remove that? what happens if the first Id is 100000000000 and the second is 100000000001 (which is totally legal for gmsh)? we now have a vector of size 100000000002 instead of 2. 

And to my knowledge `unordered_map` does not have constant lookup. If two elements share the same hash, the "buckets" use linear search algos. But I might be wrong. In general it shouldn't be (that much) slower.

I think we are talking about a small detail, that nobody really cares about and that wouldn't be a performance issue. So if you really want it changed to an `unordered_map`, do it. All I am saying is: I don't see any reason to do it and therefore won't do it myself :wink:

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: both IDs are used in the line

~~~cpp
std::vector<NodeSimple*> nodePtrVec(fileContent.maxNodeId - fileContent.minNodeId + 1, nullptr);
~~~

Since the vector isn't filled with anything yet, we can't get the first members Id to get rid of minId :wink:

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: But for the readability it is not always beneficial to split the code in multiply functions for the reason I wrote above. At this point here, I think it is more about personal  preferences than any real benefit. YMMV

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: I am not sure. Until some point in time it didn't even compile without the std. Would be interesting which header introduces the global getline. We are definitely not `using namespace std`.

src/mechanics/mesh/MeshGmsh.cpp
vhirtham: Not sure what you mean? Is that a general design issue not related to gmsh?

TTitscher (approved):

PR #156, opened 2017-12-22
PDE CMake export dependencies

Psirus:
With the move to the imported targets, our dependencies where not properly exported by `sudo make install`. This tries to fix it.


scripts/cmake/NuToConfig.cmake
```diff
@@ -0,0 +1,7 @@
+include(CMakeFindDependencyMacro)
+set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
+find_dependency(MUMPS)
+find_dependency(ARPACK)
+find_dependency(ANN)
```
TTitscher, line 5: Just a reminder: What happens if ARPACK/ANN/MUMPS is not found in the original build. That could happen, if I chose a Trilinos solver and do not install MUMPS...

TTitscher (approved):

Issue #64, opened 2017-06-22
Travis improvements


Psirus:
Speed
-----

The Travis builds take quite a while, especially the coverage/debug build is now up to almost 28 minutes. Especially for pull requests, where each commit is build twice (once on the branch, once "as if" merged into master), the wait can become quite long. Also, I think that we should strive to be as economical as possible in our use of infrastructure that has been provided to us for free.

For example, just as locally, one can [use ccache with Travis](https://docs.travis-ci.com/user/caching).

Another option is to [use Docker images](https://docs.travis-ci.com/user/docker/). While I'm no expert, the idea is to bundle everything you need into one container. So instead of running `apt-get install very many packages` every build, you create a sort of image that has all the dependencies and use that instead. I'm not sure if this will be quicker, but maybe someone wants to try.

Seperate Unit/Integration Coverage
----------------------------------

With the use of [flags](https://docs.codecov.io/docs/flags), the code coverage can be seperated into results from unit tests and results from integration tests. Imho this would be really interesting to see. Also, one could seperate the coverage of `src/` from the coverage of the tests themselves. Take a look at the [manpage of lcov](http://ltp.sourceforge.net/coverage/lcov/lcov.1.php) on how to use the `--remove` and `--extract` options to achieve this. I suggest trying this locally first ([lcov quickstart guide](http://ltp.sourceforge.net/coverage/lcov.php), instead of many iterations on Travis.

phuschke:
**Status quo**
I managed to pull a docker image and use that image to run make + ctest. I tested it on this simple cpp project if you want to have a look:

[phuschke/GithubCppProjectTemplate](https://github.com/phuschke/GithubCppProjectTemplate)

Will build a new docker image for NuTo and update .travis.yml soon.

TTitscher:
I like! Any approximations for the speedup?

TTitscher:
I tried to go on a different road, following https://atchai.com/blog/faster-ci/

Travis is somehow able to cache our apt-get packages. This sounds like, what we want. But: We have to switch to their [APT Addon](https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-on-Container-Based-Infrastructure). The _only_ problem I currently have: `swig3.0`. This is included in ubuntu/trusty-backports, which is no problem for our current travis config. With this APT Addon thing, trusty-backports is not included. In addition, we cannot use `sudo` if we want to do that... Any ideas?

Any progress on Docker, @phuschke ? Can I help you somehow?

[...]

PR #158, opened 2018-01-04
Pde docker

TTitscher:
Basically solves #64 

- Travis provides ubuntu 14.04 which is outdated
- we need custom install scripts
- we cannot reproduce the travis runs

Docker container offer a nice solution. We now have a [Dockerfile](https://github.com/nutofem/nuto/blob/PDE_docker/Dockerfile) that defines something like an ubuntu virtual machine with all our dependencies. This image is uploaded to [dockerhub](https://hub.docker.com/r/nuto/nuto_docker/) and provides all the benefits of the docker service: You can run that image locally, on a server somewhere, and on travis. 

We can even create a docker container that already has nuto installed. For tutorials or whatever... another issue.
  
__edit:__
- I guess the coverage dropped because I set ENABLE_PYTHON=FALSE
- if you want do perform changes to the docker container, you have to adapt the `Dockerfile` and run `sudo bash /scripts/upload.docker.sh`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,14 +9,12 @@ notifications:
         on_success: change  # always|never|change
         on_failure: always
         on_start: never
-matrix:
-    include:
-        - compiler: clang
-          env: BUILD_TYPE=Release OPENMP=TRUE
-        - compiler: g++
-          env: BUILD_TYPE=Release OPENMP=FALSE
-          env: BUILD_TYPE=Release OPENMP=TRUE
-          env: BUILD_TYPE=Debug OPENMP=FALSE COVERAGE=--coverage
+
+env:
+    - BUILD_OPTIONS="-DCMAKE_CXX_COMPILER=clang++ -DCMAKE_BUILD_TYPE=Release -DENABLE_OPENMP=FALSE"
+    - BUILD_OPTIONS="-DCMAKE_CXX_COMPILER=g++ -DCMAKE_BUILD_TYPE=Release -DENABLE_OPENMP=FALSE"
+    - BUILD_OPTIONS="-DCMAKE_CXX_COMPILER=g++ -DCMAKE_BUILD_TYPE=Release -DENABLE_OPENMP=TRUE"
+    - BUILD_OPTIONS="-DCMAKE_CXX_COMPILER=g++ -DCMAKE_BUILD_TYPE=Debug -DENABLE_OPENMP=FALSE -DCMAKE_CXX_FLAGS=--coverage"
 
 before_install:
     - ci_env=`bash <(curl -s https://codecov.io/env)`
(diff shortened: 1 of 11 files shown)


scripts/upload_docker.sh
```diff
@@ -0,0 +1,3 @@
+sudo docker build -t nuto_docker .
+sudo docker tag nuto_docker nuto/nuto_docker:xenial
+sudo docker push nuto/nuto_docker:xenial
```
Psirus: As far as I can tell, none of this needs `sudo`.

.travis.yml
```diff
@@ -8,49 +9,30 @@ notifications:
...
-    - sudo apt-get -qq update
-    - sudo apt-get -qq install -y g++-6 clang libeigen3-dev libgcc-6-dev libiomp-dev swig3.0 libboost-filesystem1.55-dev libboost-test1.55-dev libboost-mpi1.55-dev libopenblas-dev libmetis-dev libmumps-seq-dev libann-dev libarpack2-dev lcov curl libglu1-mesa-dev
-before_script:
-    - mkdir build
-    - cd build
-    - BUILD_OPTIONS="-DCMAKE_BUILD_TYPE=$BUILD_TYPE -DENABLE_OPENMP=$OPENMP -DENABLE_MPI=TRUE"
-    - if [[ "$CXX" == "g++" ]]; then CXX=g++-6 cmake $BUILD_OPTIONS -DCMAKE_CXX_FLAGS="$COVERAGE $WERROR" ..; fi
-    - if [[ "$CXX" == "clang++" ]]; then CXX=clang++ cmake $BUILD_OPTIONS -DCMAKE_CXX_FLAGS="$COVERAGE $WERROR" ..; fi
+    - ci_env=`bash <(curl -s https://codecov.io/env)`
+    - echo $ci_env
```
Psirus: The echo is not needed.

Psirus (approved):
Looks good. 

One note: 
> if you want do perform changes to the docker container, you have to adapt the Dockerfile and run sudo bash /scripts/upload.docker.sh

I guess you need to authenticate to docker.io, because I can't upload any changes. (Which makes sense, because otherwise anyone could edit our container).

CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ option(ENABLE_CUSTOM_EXAMPLES
     "enable private examples in separate directory (applications/custom)" FALSE)
 option(ENABLE_OPENMP "enables OpenMP" FALSE)
 option(ENABLE_PARDISO "enables support for PARDISO solver" FALSE)
-option(ENABLE_PYTHON "create python wrapper for nuto" TRUE)
+option(ENABLE_PYTHON "create python wrapper for nuto" FALSE)
```
joergfunger, line 12: Why is this turned off?

joergfunger (approved):
Where exactly happens the upload to dockerhub?

CMakeLists.txt
TTitscher: The future python interface is not yet of our PDE development, so I want to save some time in travis and deactivate it. Alternatively, I could have passed `-DENABLE_PYTHON=FALSE` to travis. But since python is currently not working for /nuto/mechanics, I switched it off there...

TTitscher:
As for the image upload. You will have to create an account on https://hub.docker.com/. This account has to be added to our `nuto` account by one of its owners, currently me and phuschke. Then, you should be able do `docker push` to this repo. 
 So please send me an email with your dockerhub username.
  __Edit:__ Maybe, you'll need to run `docker login` first.

PR #160, opened 2018-01-11
Pde gmsh import addon

pmueller2:
Adds some more element types to gmsh import. Implemented some interpolations. New are:
Linear Tetrahedra, Bricks, Prisms
Quadratic Triangle, Quad (Serendipity)
"Linear" Pyramid. 
This last one is not really linear nor polynomial (not possible for pyramids) but conforms to the bilinear quad element (the pyramid base)


src/mechanics/elements/ElementShapeFunctions.cpp
```diff
@@ -1372,6 +1372,84 @@ Eigen::Matrix<double, 18, 3> DerivativeShapeFunctionsPrismOrder2(const Eigen::Ve
...
+{
+    const double x = rCoordinates[0];
+    const double y = rCoordinates[1];
+    double z = rCoordinates[2];
+
+    Eigen::Matrix<double, 5, 3> derivativeShapeFunctions;
+
+    if (std::abs(x * y) < 1.e-12)
+    {
+        z = 0.99;
```
Psirus: Que?

test/mechanics/interpolation/InterpolationPrismLinear.cpp
```diff
@@ -0,0 +1,16 @@
...
+{
+    constexpr double a = 0.577350269189626;
+    return {Eigen::Vector3d({-a, -a, -a}), Eigen::Vector3d({-a, a, -a}), Eigen::Vector3d({a, a, -a}),
+            Eigen::Vector3d({a, -a, -a}),  Eigen::Vector3d({-a, -a, a}), Eigen::Vector3d({-a, a, a}),
+            Eigen::Vector3d({a, a, a}),    Eigen::Vector3d({a, -a, a})};
+}
+
+BOOST_AUTO_TEST_CASE(InterpolationPrismLinear)
+{
+    NuTo::Test::RunTests<NuTo::InterpolationPrismLinear>(GetTestPoints());
```
Psirus, line 15: I know this is not your fault, just the first time I'm noticing this: `RunTests` as a name is really unhelpful. Can we rename it to `CheckInterpolation` or something?

test/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -18,3 +18,48 @@ BOOST_AUTO_TEST_CASE(NonContiguousNodeNumbering)
...
+
+BOOST_AUTO_TEST_CASE(PrismLinear)
+{
+    BOOST_CHECK_NO_THROW("prism1.msh");
+}
+
+BOOST_AUTO_TEST_CASE(PyramidLinear)
+{
+    BOOST_CHECK_NO_THROW("pyramid1.msh");
+}
```
Psirus, line 65: Shouldn't we add at least one case where we check more than `NO_THROW`?

Psirus (approved):
Nice! Just another thing that's not your fault, but I've noticed again:
1. Why are the shape functions separated into one extra file, instead of being defined right in the interpolation?
2. Should there be a namespace `Interpolation`? So e.g. `NuTo::Interpolation::PrismLinear`.

TTitscher:
@Psirus I think your points (extra file + namespace) make 100% sense. But I also think that this whole interpolation/integration needs kind of a redesign. See #34. It would be nice to have something like
~~~cpp
auto integration = SuitableIntegration(interpolationType, GAUSS, /*order =*/ 2);
auto coordInterpolation = Shape::Quad::Interpolation(SERENDIPITY, /*order =*/ 1);
auto dispInterpolation = SuitableInterpolation(coordInterpolation, LOBATTO, /*order =*/ 4);
// or whatever
~~~
Until we come up with a nice idea for that, I think minimizing all efforts that go into these classes is not a bad idea...

Psirus:
Fair enough.

PR #161, opened 2018-01-15
Pde damage law

TTitscher:
Adds an isotropic local damage model. Since we agreed on using a header-only integrand lib, the laws can be header-only as well and I chose a policy based design. Via template parameters, you can freely chose the
- damage law
- evolution equation
- underlying elastic law

See the documentation for details.


src/mechanics/constitutive/ModifiedMises.h
```diff
@@ -0,0 +1,253 @@
...
+    //! @param strain strain to evaluate
+    //! @return the value of the modified Mises strain norm
+    double Value(const NuTo::EngineeringStrain<TDim>& strain) const;
+
+    //! @param strain strain to evaluate
+    //! @return the derivative of the modified Mises strain norm with respect to the strain
+    Eigen::Matrix<double, Voigt::Dim(TDim), 1> Derivative(const NuTo::EngineeringStrain<TDim>& strain) const;
+
+private:
+    //! This method may be useful for other classes. However, it cannot be a member of
```
Psirus: Imho, this shouldn't be in Doxygen. I'd like to know what `Strain3D` does, not whether it may be useful for other classes. Let's make that either regular comments or `@remark`s.

Psirus (approved):

src/mechanics/constitutive/EngineeringStrain.h
```diff
@@ -5,9 +5,9 @@
 
 namespace NuTo
 {
```
joergfunger, line 7: Shouldn't this be moved into a separate integrand lib? Or the other way round, shouldn't we move the whole pde integration/cell features into another directory/lib and reorganize the mechanics subdirectory into a a subrepo?

src/mechanics/constitutive/ModifiedMises.h
```diff
@@ -0,0 +1,256 @@
...
+ * With the strain tensor invariant I1 and the deviatoric strain tensor invariant J2, it reads
+ * \f[/
+ * \varepsilon_\text{eq}(\boldsymbol \varepsilon) = \frac{k-1}{2k(1-2\nu)}I_1 +
+ * \frac{1}{2k}\sqrt{\left(\frac{k-1}{1-2\nu}I_1\right)^2 + \frac{2k}{(1+\nu)^2}J_2}
+ * \f]
+ *
+ * @tparam TDim dimension 1,2,3
+ */
+template <int TDim>
+class ModifiedMises
```
joergfunger: the naming should somehow include what is done (it is not a plasticity model with a modified Mises yield function, but just calculates the Mises - norm). Also, it should be somehow clear that is using the strains (and not the stresses) - or is there a possibility to reuse this when I'm converting the standard Mises-plasticity model to pde and have to define the yield function also as a function of J2?

src/mechanics/constitutive/LocalIsotropicDamage.h
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    void Update(EngineeringStrain<TDim> strain, double deltaT, int cellId, int ipId)
+    {
+        mKappas[Ip(cellId, ipId)] = Kappa(strain, deltaT, cellId, ipId);
+    }
+
+    size_t Ip(int cellId, int ipId) const
+    {
+        return cellId * mNumIpsPerCell + ipId;
```
joergfunger, line 121: Have we agreed on making separate integrands for cells of different dimensionality of the integrand (e.g. triangles and quads with a different number of integration points per cell)

test/mechanics/constitutive/LocalIsotropicDamage.cpp
```diff
@@ -0,0 +1,72 @@
...
+    const double kappa0 = ft / E;
+    const double beta = ft / gf;
+    const double alpha = 1;
+
+    // Define the law using policy based design principles, hopefully applied correctly
+    Laws::LinearElastic<1> linearElastic(E, nu);
+    Constitutive::DamageLawExponential dmg(kappa0, beta, alpha);
+    Constitutive::ModifiedMises<1> strainNorm(nu, fc / ft);
+    Laws::EvolutionImplicit<1, Constitutive::ModifiedMises<1>> history(strainNorm, /*numCells=*/1, /*numIps=*/1);
+    auto localDamageLaw = Laws::CreateLocalIsotropicDamage<1>(dmg, history, linearElastic);
```
joergfunger: I know that it is very convenient to write, but I think for test files that are often used as an example on how to use specific features, I personally would prefer not using the auto feature.

joergfunger (approved):
Nice general and flexible structure of the damage law.

src/mechanics/constitutive/LocalIsotropicDamage.h
TTitscher: `THistoryDataContainer` with `THistoryDataContainer::operator(cellId, ipId)` could be another template parameter for that.

test/mechanics/constitutive/LocalIsotropicDamage.cpp
TTitscher: I could do that. Without auto it would look like
~~~cpp
Laws::LocalIsotropicDamageLaw<1, Constitutive::DamageLawExponential, Laws::EvolutionImplicit<1, Constitutive::ModifiedMises<1>>, Laws::LinearElastic<1>> localDamageLaw(dmg, history, linearElastic);
~~~
or
~~~cpp
using DmgLaw = ...
using History = ...
using ElasticLaw = ...
Laws::LocalIsotropicDamage<1, DmgLaw, History, ElasticLaw> localDamageLaw(...);
~~~
Significantly longer. And not that easy to change since you have to type the types twice. I would prefer the second version, if we want to avoid `auto`.

src/mechanics/constitutive/ModifiedMises.h
TTitscher: The voight notation makes it hard to reuse `strain` methods for `stress`. We can try to reuse `mechanics/constitutive/EngineeringStrainInvariants.h`. They contain a method `ToTensor` that is different for stresses and strains. The invariants itself only depend on the tensor then. Does that make sense?

__Edit:__ I just saw that `ToTensor` is not actually used and the invariants are implemented directly in voigt notation. Maybe because of performance issues...

PR #162, opened 2018-01-15
Fix gmsh tests

TTitscher:
Fixes the existing import mesh tests that were not actually run.


pmueller2 (commented):
Upsi... Good that now also the build meshFEM is checked.

pmueller2 (approved):

PR #163, opened 2018-01-15
Pde binary gmsh format

TTitscher:
Add binary gmsh format support.


src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -154,49 +176,150 @@ std::vector<GmshElement> ReadElementsASCII(std::ifstream& rFile)
...
 {
     std::string line;
     std::getline(rFile, line);
-    if (line[0] == '$')
+
+    if (line.empty())
+        return;
+
+    if (line == "$PhysicalNames")
     {
```
pmueller2: By just checking equality (you removed the compare) trailing whitespaces in the Gmsh file will produce errors. Gmsh itself has no problem with that so NuTo also shouldn't (IMHO). This is an error that is I think not easy to find, so why not just use compare?

TTitscher: `string1.compare(string2)` is actually equal to `string1 == string2`. But I solved that trailing whitespace issue by comparing both strings only up to `string2.size()`.

pmueller2 (approved):
Ok, laser (or what the young people say).

PR #165, opened 2018-01-17
Change data type of DofMatrix/Vector to map

Psirus:
Previously, they didn't know the DofTypes they consisted of. This made
certain operations (e.g. outstream) very difficult and required passing
the doftype in situations where you would not expect it to be necessary.

Exporting the full vector/matrix and consequential interface changes
will follow shortly.


src/mechanics/dofs/DofMatrixContainer.h
```diff
@@ -1,55 +1,98 @@
...
-    static std::pair<int, int> CantorPairingFunctionReverse(int z)
-    {
-        int w = std::floor((std::sqrt(8 * z + 1) - 1) / 2);
-        int t = (w * w + w) / 2;
-        return std::make_pair(z - t, w - z + t);
-    }
+        bool operator()(const DofPair& a, const DofPair& b) const
+        {
+            // check for equiv(a.first, b.first)
+            if (!(a.first.GetName() < b.first.GetName()) && !(b.first.GetName() < a.first.GetName()))
```
TTitscher: `DofType` has a unique `.Id()` but not necessarily a unique `.GetName()`.

Psirus: Yes I also realized that while cycling home 😬

joergfunger (approved):
I'm not sure if we still need the dof containers (and thus user-defined operators), but this approach with the map is more intuitive than the previous version.

TTitscher (approved):

PR #166, opened 2018-01-19
Pde add unit mesh3 d

pmueller2:
Adds a 3D Brick unit mesh and a method to change the coordinate interpolation. This last method is almost a copy of `AddDofInterpolation`. The old coordinate nodes are replaced by new ones in the hope that nothing is referencing them. The old interpolation stays.


src/mechanics/mesh/MeshFemDofConvert.cpp
```diff
@@ -174,3 +174,45 @@ void NuTo::AddDofInterpolation(MeshFem* rMesh, DofType dofType, const Interpolat
...
+void NuTo::ChangeCoordinateInterpolation(MeshFem* rMesh, const InterpolationSimple& interpolation)
+{
+    // Setup subbox. These argument values are quite arbibrary and should maybe be chosen based on
+    // the dimensions of the mesh.
+    SubBoxes<NodePoint> subBoxes(SetupSubBoxDomain(*rMesh, /*numBoxesPerDirection=*/300, /*eps=*/1.e-10));
+
+    // Save old coordinate node refs to remove them later
+    auto oldCoordinateNodes = rMesh->NodesTotal();
+
+    for (size_t i = 0; i < rMesh->Elements.Size(); i++)
```
Psirus: Could this be a range-based for-loop?

src/mechanics/mesh/MeshFemDofConvert.cpp
```diff
@@ -174,3 +174,45 @@ void NuTo::AddDofInterpolation(MeshFem* rMesh, DofType dofType, const Interpolat
...
+
+        const auto& coordinateElement = elementCollection.CoordinateElement();
+        for (int iNode = 0; iNode < interpolation.GetNumNodes(); ++iNode)
+        {
+            Eigen::Vector3d coord = To3D(Interpolate(coordinateElement, interpolation.GetLocalCoords(iNode)));
+
+            NodePoint* nodePoint = subBoxes.FindAt(coord);
+            if (nodePoint)
+            {
+                nodesForTheNewlyCreatedElement.push_back(&nodePoint->mNode);
```
Psirus: I'd prefer a couple of parentheses (`&(nodePoint->mNode)`), because the operator precedence is not obvious to me.

src/mechanics/mesh/UnitMeshFem.h
```diff
@@ -23,6 +23,13 @@ MeshFem CreateTriangles(int numX, int numY);
 //! @return created mesh
 MeshFem CreateQuads(int numX, int numY);
```
Psirus, line 24: Admittedly outside of this PR, but instead of `CreateTriangles` and `CreateQuads`, i'd prefer `CreateUnitSquare` or `UnitSquareMesh` or something, with the type of element as a parameter.

Psirus (approved):

Issue #159, opened 2018-01-11
Docker image


Psirus:
We've briefly discussed that providing a Docker image with NuTo pre-installed would be nice to have. This is probably low priority, but just so that we don't forget I've created this issue :)

Things to consider:
- What else to install? (Jupyter?)
- How to generate? (Manually/Automatically?)
- ... ?

PR #167, opened 2018-01-19
Docker fix coverage

Psirus:
Since #158 has been merged, the coverage has been "messed up". 

- `src/base/Version.cpp` was not filed into the correct directory,
- the coverage of fakeit was wrongly included
- there was an extra `source` directory that one had to step through to get to the information

Compare [before #158](https://codecov.io/gh/nutofem/nuto/tree/a723474d773b763d68eb3fd0d7460d998a4c9768), [after #158](https://codecov.io/gh/nutofem/nuto/tree/5482c8c97576bfae8080eddddb38bf117226f397) and [now](https://codecov.io/gh/nutofem/nuto/tree/b707c6afd57abce391188875f532a01908a9e585).


Dockerfile
```diff
@@ -24,6 +24,7 @@ USER nuto
 
 # create the source directory /home/nuto/source
 RUN mkdir /home/nuto/source
+RUN mkdir /home/nuto/build
 
-# the build directory will be in /home/nuto 
-WORKDIR /home/nuto
+# the build directory will be in /home/nuto/source/build
```
TTitscher: Should be `home/nuto/build`. But I think the directory name itself indicates its purpose and we can delete this comment.

.travis.yml
```diff
@@ -16,16 +16,17 @@ matrix:
 
 before_install:
     - CI_ENV=`bash <(curl -s https://codecov.io/env)`
-    - travis_retry docker pull nuto/nuto_docker:xenial
-    - docker run $CI_ENV -itd --user nuto --name dock -v $(pwd):/home/nuto/source nuto/nuto_docker:xenial
+    - travis_retry docker pull nuto/nuto_docker:dev
```
TTitscher, line 19: The name is fine for now, I guess. Do we want to establish a naming convention?

TTitscher (approved):
Thanks for fixing the :duck:er image :eyes:. When I look at the travis build history, it looks like some very error-prone lines of code...

.travis.yml
Psirus: I thought that maybe a Docker image that has NuTo installed will follow in the future (see #159), and since this is one that has all the dependencies installed in case you want to start developing NuTo, not necessarily using it, we'd call this one `dev` and the other maybe `run` or `compute` or something.

PR #168, opened 2018-01-22
Fix slow matrix assembly with triplet lists

TTitscher:
- rename `SimpleAssember`
- add cpp file


src/mechanics/cell/SimpleAssembler.cpp
```diff
@@ -0,0 +1,133 @@
...
+                                        .push_back({globalDofNumberI - numIndependentDofsI, globalDofNumberJ,
+                                                    globalDofValue});
+                            }
+                            else
+                            {
+                                tripletsKK(dofI, dofJ)
+                                        .push_back({globalDofNumberI - numIndependentDofsI,
+                                                    globalDofNumberJ - numIndependentDofsJ, globalDofValue});
+                            }
+                        } // argh. any better ideas?
```
Psirus, line 92: If there were a class like `GlobalDofTripletList`, you could introduce an `operator()` to hide the if-else as you do in `GlobalDofVector::operator()`, but not sure if this is better...

Psirus (approved):

PR #169, opened 2018-01-23
Remove object libs from build system

Psirus:
These were intended to avoid multiple compilation of the same file while still allowing building the unit tests without the finished library. However, it makes the build system quite complicated and prohibits using the CMake targets properly. See #98.

For those coming from #98 and wondering which option of a), b) and c) this implements: d) Don't build things twice, but building unit tests without building the library first is not possible anymore.


TTitscher (changes requested):
You talk about those options, right?
>  a) Don't properly use the new features and just keep piling on the include directories for all the targets as we have been doing. Meh.
b) Now that make unit and make don't overlap anymore, is it acceptable to build things twice?
c) Expand the functionality of our CMake macros. There are a chain of macros that gets you from the list of sources to the finished NuTo library, to be found in scripts/NuToMacros.cmake (specifically create_nuto_module, source_to_objects, create_object_lib). I'm sure these could be augmented to pass the include directories through to the OBJECT libraries. But that would have to be written and maintained as well.

If I recall correctly, we introduced the split in unit/integration/examples with the following ideas in mind:
1) examples are like long integration tests, in terms of run time
2) integration tests require multiple components, thus `libnutomechanics`
3) unit test should also run if `libunitmechanics` fails to build to easily identify broken parts and fix them one after the other
4) compile no cpp twice

With this PR, what separates unit tests from integration tests? From the build systems point of view, they are like examples and integration tests but run even faster? That does not sound right to me. So...

A) if we like that idea, we should think about simplify those to `test` (= unit tests and integration tests) and `examples`. Or... 

B) we think again about option b) in the list above... (which means compiling things twice; and which would still be my favorite for now)

__edit:__ I think, when the rate of changes in the `core` decreases, a simple `make test` (option A) as everyone else is doing, is sufficient.

Psirus:
I don't lean strongly either way, with regards to "b) or d)". So I've implemented your favorite version. However, I find the difference between unit and integration tests is conceptual, not bound to the build system. A unit test should have very few dependencies, and should test the one class/function under consideration. Granted, this can be ensured when you're not "allowed" to link the rest of the library to it.

TTitscher (approved):

PR #170, opened 2018-01-24
Use timeout when pulling the docker image

Psirus:
Travis sometimes fails when Docker doesn't pull correctly for some reason. This should now be prevented by automatically trying again.


TTitscher (approved):
That is what I had in mind with this `travis_retry`. But my execution was flawed. Thanks!

PR #171, opened 2018-01-24
Pde multiple dof converts

TTitscher:
Make interpolation in AddDofInterpolation optional.

If you do _not_ provide a argument, each new dof element
will use the underlying coordinate interpolation resulting
in an isoparametric element. This will also work for mixed
meshes consisting of multiple interpolation types


src/math/EigenCompanion.h
```diff
@@ -8,36 +8,35 @@ namespace NuTo
...
-    static Eigen::VectorXd ToEigen(double d)
-    {
-        return Eigen::VectorXd::Constant(1, d);
-    }
+//! @brief transforms a single double number to an Eigen::VectorXd
+inline Eigen::VectorXd ToEigen(double d)
+{
+    return Eigen::VectorXd::Constant(1, d);
+}
 };
```
Psirus: Unnecessary semicolon, no?

Psirus (approved):
Nice! I'm now officially a fan of `boost::optional`. 🍇

PR #173, opened 2018-01-26
Add quadratic line element

TTitscher:
Gmsh node numbering does not always match our numbering, e.g. for
quadratic line elements. This changes in the gmsh import allows
addressing this.

### DIFF
--- a/src/mechanics/mesh/MeshGmsh.cpp
+++ b/src/mechanics/mesh/MeshGmsh.cpp
@@ -2,6 +2,7 @@
 
 #include "base/Exception.h"
 #include "mechanics/interpolation/InterpolationTrussLinear.h"
+#include "mechanics/interpolation/InterpolationTrussLobatto.h"
 #include "mechanics/interpolation/InterpolationTriangleLinear.h"
 #include "mechanics/interpolation/InterpolationTriangleQuadratic.h"
 #include "mechanics/interpolation/InterpolationQuadLinear.h"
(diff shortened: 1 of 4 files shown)


src/mechanics/mesh/MeshGmsh.cpp
```diff
@@ -361,6 +364,24 @@ const NuTo::InterpolationSimple& CreateElementInterpolation(NuTo::MeshFem& rMesh
...
+std::vector<NuTo::NodeSimple*> GetElementNodes(const std::unordered_map<int, NuTo::NodeSimple*>& nodePtrs,
+                                               const GmshElement& gmshElement)
+{
+    std::vector<NuTo::NodeSimple*> elementNodes(gmshElement.nodes.size());
+    for (unsigned int i = 0; i < elementNodes.size(); ++i)
+        elementNodes[i] = nodePtrs.at(gmshElement.nodes[i]);
+
+    switch (gmshElement.type)
+    {
+    case 8: /* TrussQuadratic */
```
Psirus, line 376: Should we introduce a `eGmshElementType` enum?

Psirus (approved):

pmueller2 (approved):
Ok. Can this be documented somehow? So that you immediately know if you look at this how Gmsh ordering is different?

Psirus:
I'd wait for other interpolations, maybe it happens more often, and it makes sense to introduce a mapping of `vector<int>` to `vector<int>` or something that handles it. So that I could say `GetGmshToNuToMapping(GmshElementType)` or similar.

TTitscher:
IMO we should handle the whole det(jacobian) thing in a dedicated PR.

PR #175, opened 2018-01-26
Pde jacobian adds

pmueller2:
Makes the Jacobian accessable from CellIPData and adds extended Jacobian for 2D in 3D or 1D in 2D situations. 

Example: The Jacobian for 2D in 3D is a rectangular matrix(3,2). If you add the surface normal as a 3rd column you get a usual 3x3 matrix again. Its inverse can be used in the transformation of gradients just as usual giving you the so called surface gradient (i.e. the gradient with its normal component missing).

Apart from that the Jacobian columns represent the covariant basis for vectors in tangent space of the surface element and the Rows of the Inverse their reciprocals and may thus be useful in laws dealing with surfaces.


joergfunger (changes requested):
What exactly do we need a 2D in 3D Jacobian for? IMO, the case with an embedded lower order element in higher dimensions should be dealt with separately by a transformation matrix that only does rotations (rotating e.g. 3d-nodal values (coordinates, dofs) into the 2D-system that describes the boundary), then we perform the standard 2D integration (with a pure 2D-Jacobian) returning a 2D-element matrix which is afterwards "blown up/rotated back" into the full 3D system.

src/mechanics/cell/Jacobian.h
```diff
@@ -21,53 +21,71 @@ class Jacobian
...
+                mDetJacobian = jacobian.determinant();
                 break;
+            }
             case 3:
-                std::tie(mInvJacobian, mDetJacobian) = CalculateFixedSize<3>(nodeValues, derivativeShapeFunctions);
+            {
+                Eigen::Matrix<double, 3, 3> jacobian = CalculateFixedSize<3,3>(nodeValues, derivativeShapeFunctions);
+                mJacobian = jacobian;
+                mInvJacobian = jacobian.inverse();
+                mDetJacobian = jacobian.determinant();
```
TTitscher, line 44: These 4 lines occur - with the only change being the dimension 1,2,3 - two times more. I would recommend a tiny template method for that.

src/mechanics/cell/Jacobian.h
```diff
@@ -21,53 +21,71 @@ class Jacobian
...
                 break;
             }
+            }
         }
-        else
+        else if (globalDimension == interpolationDimension + 1)
```
TTitscher, line 49: We have these `if`s that each contain another switch. So optionally, can you think of a suitable split into methods? Maybe even using template specialization?

src/mechanics/cell/Jacobian.h
```diff
@@ -21,53 +21,71 @@ class Jacobian
...
             {
-                const int numRows = derivativeShapeFunctions.rows();
-                NodeValues nodeCopy = nodeValues;
-                // see CalculateFixedSize(...) for the Eigen::Map functionality
-                auto nodeBlockCoordinates =
-                        Eigen::Map<Eigen::Matrix<double, 2, Eigen::Dynamic>>(nodeCopy.data(), 2, numRows);
-                mDetJacobian = (nodeBlockCoordinates * derivativeShapeFunctions).norm();
+                Eigen::Matrix<double, 2, 1> jacobian = CalculateFixedSize<2,1>(nodeValues, derivativeShapeFunctions);
+                Eigen::Matrix<double, 2,2> extendedJacobian;
+                extendedJacobian.col(0) = jacobian;
```
TTitscher: IMO you can assign the result of `CalculateFixedSize<2,1>` directly to `extendedJacobian.col(0)` without this temporary `jacobian` variable.

src/mechanics/cell/Jacobian.h
```diff
@@ -21,53 +21,71 @@ class Jacobian
...
+                Eigen::Vector2d normal( -jacobian(1,0), jacobian(0,0));
+                normal /= normal.norm();
```
TTitscher: There is something like `Eigen::normalize()` that might be appropriate here. Plus, I like to initialize the values as const as possible, so I think the following will work
~~~
const Eigen::Vector2d normal = Eigen::Vector2d(-jac(1,0), jac(0,0)).normalize()
~~~
And then we can think about avoiding the variable `normal` completely and assign the expression directly to ` extendedJacobian.col(1)`. But in this case, it might increase readablilty.

TTitscher (changes requested):
@joergfunger : I think that these lower dimensional jacobians, at least their determinant, are required. We currently use them for the neumann boundary conditions of mechanics problems. And they cannot be transformed in a lower dimensional space. Imagine a truss along the _x_ axis where you want to apply a normal pressure ( = force in _y_ direction). If you transform that into a 1D truss, there will be no _y_ components and you cannot solve that by the proposed transformation. So, again, I think that this is fine as it is.
The feature that you requested is useful as well, e.g. fibers in concrete. But this is another story IMO.

Back to the track: I added comments about small refactorings that you may or may not include. I request changes because I see no test or example actually using it. I mean, I read your explanation about surface gradients and that sounds ok. On the one hand, I cannot think about actual applications for that. On the other hand, it does not break existing code. So if you need that feature, I am fine with it. But then I would like you to provide a test/example.

joergfunger:
I'm not arguing that we don't need the normals to a local 2D in 3D coordinate system I only argue that this should be not included in the Jacobian, but instead be related to a new orthogonal coordinate system attached to this integration point. E.g. when I'm interested to apply not only a pressure, but something in the tangential direction, I would need normalized basis vectors (this becomes more difficult if we would consider none-cartesian coordinates). Similar, when calculating some derivatives (say strains) I won't be able to use this approach (since I do not need the derivative with respect to the natural coordinate system). In a similar way, when we are using e.g. a 1D truss in 3D. There are assumptions about the stress-strain relation (only s_xx is not vanishing). Using the above procedure, we would violate this. My suggestion in this case would be to rotate the nodal values as well as the coordinates in a local orthogonal coordinate system, calculate element matrices in this 1D system and then blow up the matrix (or the vector) to 3D. This same rotation matrix can also be used for the calculation of pressures, but it is (as opposed to the Jacobian above) an orthogonal matrix. Thus, my suggestion is to not add this to the Jacobian, but instead introduce another rotation matrix.

pmueller2:
Since it causes more questions than enlightenment and since elements of higher codimension are not in use and its application in finite strain seems also far away I will close this pull request. When the time comes it will reopen probably in a similar fashion.

PR #179, opened 2018-02-01
Pde visualize enhancements

TTitscher:


src/visualize/VoronoiHandler.cpp
```diff
@@ -40,3 +49,15 @@ void VoronoiHandler::CellData(int cellId, std::vector<Eigen::VectorXd> values, s
...
+                               std::vector<int> pointIds, std::string name, UnstructuredGrid* grid)
+{
+    grid->DefinePointData(name);
+    for (size_t iPoint = 0; iPoint < mGeometry.pointCoordinates.size(); ++iPoint)
+    {
+        auto coords = cell.Interpolate(mGeometry.pointCoordinates[iPoint]);
+        auto value = f(coords);
+        grid->SetPointData(pointIds[iPoint], name, value);
+    }
+}
```
pmueller2, line 63: This is the same as in AverageHandler. Mm. Maybe when we have a third Handler one could think about this again...

src/visualize/AverageHandler.cpp
```diff
@@ -22,9 +23,16 @@ void AverageHandler::WriteDofValues(const CellInterface& cell, const DofType dof
...
-        grid->SetPointData(pointIds[iCorner], dof.GetName(),
-                           cell.Interpolate(mGeometry.cornerCoordinates[iCorner], dof));
+    {
+        auto dofValues = cell.Interpolate(mGeometry.cornerCoordinates[iCorner], dof);
+
+        if (as3d)
+            dofValues = EigenCompanion::To3D(dofValues);
+
+        grid->SetPointData(pointIds[iCorner], dof.GetName(), dofValues);
+    }
```
pmueller2, line 35: Seems like this does not depend on the handler. Maybe move to `UnstructuredGrid`?

pmueller2 (approved):

PR #180, opened 2018-02-01
Pde fix vtk ascii

pmueller2:
When exporting an ascii vtu file with small number data like 5e-8 this was treated as zero due to usage of std::to_string. Now the outstream operator is used instead.


src/visualize/DataArray.cpp
```diff
@@ -72,7 +74,11 @@ void DataArray<TDataType>::WriteAscii(std::ostream& file) const
     file << ">\n";
 
     for (auto data : mData)
-        file << " " << std::to_string(data);
+        // https://stackoverflow.com/questions/14644716/how-to-output-a-character-as-an-integer-through-cout
+        if (typeid(TDataType) == typeid(uint8_t))
```
TTitscher: I think you can make that a compile time check instead of the run time check with `typeid`.

Could the following work? If so, I would prefer that.
~~~cpp
#include <type_traits>
...
if (std::is_same<TDataType, uint8_t>::value)
   ...
~~~

TTitscher (approved):

test/visualize/UnstructuredGrid.cpp
```diff
@@ -194,3 +194,22 @@ BOOST_AUTO_TEST_CASE(Export2ndOrderElements)
...
+
+    grid.AddCell(pIds, NuTo::eCellTypes::LINE);
+
+    // Point data
+    std::string name = "Data";
+    grid.DefinePointData(name);
+    for (int pId : pIds)
+        grid.SetPointData(pId, name, 5e-8);
+
+    grid.ExportVtuDataFile("ExportSmallNumbersAscii.vtu", false);
```
TTitscher: Where exactly is the test? :scream_cat:

You could use `boost::property_tree` for reading XML, example found [here](https://github.com/nutofem/nuto/blob/PDE_reviewed/test/visualize/AverageVisualizer.cpp#L19)

PR #181, opened 2018-02-02
Pde visualize add voronoi geometries

pmueller2:
Adds Voronoi geometries for lines, quads, bricks with partitions corresponding to lobatto or gauss integration points.
The functions to compute these points are moved from integration types to math/legendre.


src/math/Legendre.h
```diff
@@ -28,6 +29,16 @@ std::vector<double> LegendreRoots(int n);
 //! @return vector of the n-2 roots
 std::vector<double> LegendreDerivRoots(int n);
 
+//! @brief computes points and weights for Lobatto quadrature in 1D
+//! @param numIPs number of integration points
+//! @return pair of quadrature weights and points range [-1,1] including boundary points
+std::pair<std::vector<double>, std::vector<double>> ComputeWeightsAndPoints1DLobatto(int nIps);
```
Psirus: Can we move these to a different header (or rename this one)? I would not have guessed that if I want to call `NuTo::Math::Polynomial::ComputeWeightsAndPoints1DGauss(nIps);`, I'll need to include `Legendre.h`.

pmueller2: Do you have an idea? Since gaussian type quadrature parameters are related to orthogonal polynomials I thought this is where one should look.

Psirus: Well the first three functions in this file were called `Legendre`, `LegendreRoots` and `LegendreDerivRoots`, all things that could be useful outside of quadrature and clearly related to `Legendre.h`.  So maybe the new functions could live in a `math/Quadrature.h` or `math/GaussQuadrature.h`? :)

src/math/Legendre.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "base/Exception.h"
```
TTitscher: may not be needed anymore.

src/visualize/VoronoiGeometries.cpp
```diff
@@ -10,10 +14,35 @@ std::vector<double> Linspace(double s, double e, size_t num)
...
+    std::vector<double> pts = NuTo::Math::ComputeWeightsAndPoints1DGauss(num).second;
+    std::vector<double> result;
+    result.push_back(-1.);
+    for (size_t i = 0; i < pts.size() - 1; ++i)
+        result.push_back((pts[i] + pts[i + 1]) / 2);
+    result.push_back(1.);
+    return result;
+}
+
+std::vector<double> Lobattospaced(size_t num)
```
TTitscher: You could avoid code duplication by introducing a 
~~~
std::vector<double> MeanSpacing(std::vector<double>); // maybe better name
~~~
that gets either lobatto or gauss points.

TTitscher (approved):

PR #182, opened 2018-02-06
Solver interface

Psirus:
This adds an interface for all of the built-in Eigen solvers, two [SuiteSparse](http://faculty.cse.tamu.edu/davis/suitesparse.html) solvers and two versions of Mumps.

Also, there is a "convenience wrapper" for solving our `GlobalDofSparseMatrix`-`GlobalDofVector` system. Unfortunately, it has imho too many parameters:

dof
---
this was necessary, because the DofMatrix/Vector didn't know the dof types they contained; since #165 has been merged, implementing a `ExportEverything()` function would be possible for these types, making this parameter redundant (or at least optional)

numIndependentDofs
---------------------------
 as with the assembler, the number of independent dofs still needs to be passed to `BuildConstraintMatrix`…not sure why this is the case, and what can be done about it

time
----
Imaginary conversation between me and the solver:
> "Here solve this system Ax = b."
> "But I need a time as well."
> "WTF?!"

the boundary conditions need the time, and I'm not sure what to do here; make it a reference?


TTitscher:
Nice work, especially the external solvers! 
When I had a look at those I realized that they all implement the `Eigen` interface, with `factorize, compute, solve, analyzePattern, ...`. 
Can you elaborate why you chose to wrap those in this _factory-like_ function? This removes the possibility to customize the solver, e.g. modify the preconditioner, adjust the tolerance, re-use the solver, etc. (... and I do not like these constraints.)
An alternative could be to pass them via template argument, as in `NuTo::NewtonRaphson`. Another idea is to follow the _master branch_ approach and have some `SolverBase`.

Psirus:
The `solve()` function is just a convenience wrapper, it is not intended to replace `SolverBase`. As you've noticed, they all implement the [Eigen Sparse Solver Concept](http://eigen.tuxfamily.org/dox/group__TopicSparseSystems.html#TutorialSparseSolverConcept). There is nothing stopping you from using them right now [you even have a file showing you what to include, how to call them and what to link to :)]. Wrapping the wrappers seems pointless imho.

Additional benefits of the "factory" include trivial Python wrapper code and no `#ifdef HAVE_MUMPS` etc. in any header files.

Dockerfile
```diff
@@ -1,29 +1,19 @@
...
-# boost
-RUN apt-get install -y libboost-filesystem-dev libboost-system-dev libboost-mpi-dev libboost-test-dev
-
-# Nuto deps
-RUN apt-get install -y libeigen3-dev libiomp-dev python3-dev python3-numpy libopenblas-dev libmetis-dev libmumps-seq-dev libann-dev libarpack2-dev gmsh 
-
-# coverage and documentation
-RUN apt-get install -y lcov curl 
+# generic C++ stuff
+RUN apt-get update && apt-get install -y --no-install-recommends g++ clang cmake make libboost-filesystem-dev libboost-mpi-dev libboost-test-dev lcov curl sudo && rm -rf /var/lib/apt/lists/*
```
TTitscher, line 4: `rm -rf /var/lib/apt/lists/*`
que?

scripts/cmake/CheckForDependencies.cmake
```diff
@@ -32,22 +32,23 @@ macro(check_for_dependencies)
...
 
-    # find mumps solver
-    if(ENABLE_MUMPS)
-        find_package(MUMPS REQUIRED)
-        add_definitions("-DHAVE_MUMPS")
-        set(NuTo_SWIG_FLAGS "${NuTo_SWIG_FLAGS};-DHAVE_MUMPS")
+    find_package(MUMPS QUIET)
+    if(MUMPS_FOUND)
+        message(STATUS "MUMPS solver found.")
+    else()
```
TTitscher, line 43: This output pattern
~~~
if (found) --> sucessful
else --> not available
~~~
structure occurs several times. Do you think it is worth wrapping that in an `optional_find_package(MUMPS)`?

src/mechanics/CMakeLists.txt
```diff
@@ -10,6 +10,8 @@ set(MechanicsSources
     tools/GlobalFractureEnergyIntegrator.cpp
     mesh/MeshFem.cpp
     mesh/MeshGmsh.cpp
+    solver/Solve.cpp
+    solver/EigenSolver.cpp
```
TTitscher: I would expect solvers to be in `/math/`.

scripts/cmake/CheckForDependencies.cmake
```diff
@@ -58,22 +59,24 @@ macro(check_for_dependencies)
...
     find_package(benchmark QUIET)
     if(benchmark_FOUND)
         message(STATUS "Found google benchmark")
     else()
         warning("Google benchmark NOT found - Benchmarks won't be build!")
     endif()
+
+    find_package(SUITESPARSE QUIET)
+    if(SUITESPARSE_FOUND)
+        message(STATUS "Found SuiteSparse solver library")
```
TTitscher, line 78: Maybe change that message to: "Sweet, found SuiteSparse solver library" or "Found sweet SuiteSparse solver library"? :bowtie:

src/mechanics/solver/EigenSolver.h
```diff
@@ -0,0 +1,31 @@
...
+//! 
+//! Built-in iteratitive solvers:
+//! - `EigenConjugateGradient`
+//! - `EigenLeastSquaresConjugateGradient`
+//! - `EigenBiCGSTAB.
+//! 
+//! and [SuiteSparse](http://faculty.cse.tamu.edu/davis/suitesparse.html) solvers (need to be installed seperately):
+//! - `EigenUmfPackLU`
+//! - `EigenCholmodSupernodalLLT`
+Eigen::VectorXd EigenSolver(Eigen::SparseMatrix<double> A, Eigen::VectorXd b, std::string solver);
```
TTitscher: 1) I prefer the name `EigenSparseSolve` since `EigenSolver` 
   - a) sounds like a class name, not an action
   - b) sounds like involves eigenvalues 

2) Why not include `EigenMUMPS` here? Same inteface.

3) I'd prefer `/math/` for that

4) An `enum eSolver` (instead of `string`) would give easy autocomplete access to all available solvers, no?

src/mechanics/solver/MumpsSolver.cpp
```diff
@@ -0,0 +1,24 @@
...
+    return solver.solve(b);
+}
+
+Eigen::VectorXd NuTo::MumpsSolver(Eigen::SparseMatrix<double> A, Eigen::VectorXd b, std::string solver)
+{
+    if (solver == "MumpsLU")
+        return MumpsSolve<Eigen::MUMPSLU<Eigen::SparseMatrix<double>>>(A, b);
+    if (solver == "MumpsLDLT")
+        return MumpsSolve<Eigen::MUMPSLDLT<Eigen::SparseMatrix<double>, Eigen::Upper>>(A, b);
+    throw Exception("Unknown solver. Are you sure you spelled it correctly?");
```
TTitscher: _spelled it correctly?_ --> `enum`
On top, this check is case sensitive and following the actual solver name `Eigen::MUMPSLU`, I probably would have expected the string to be `"MUMPSLU"`...

src/mechanics/solver/Solve.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+Eigen::VectorXd Solve(Eigen::SparseMatrix<double> A, Eigen::VectorXd b, std::string solver)
+{
+    if (isPrefix(solver, "Eigen"))
+        return EigenSolver(A, b, solver);
+    throw Exception("Que?");
+}
+
+
+GlobalDofVector Solve(GlobalDofMatrixSparse K, GlobalDofVector f, Constraint::Constraints bcs, DofType dof,
```
TTitscher: The name `Solve` may be misleading, since it will only solve linear systems correctly. I do think that it makes sense to have a method like that, maybe rename `SolveOnce(..)` or `SolveLinearSystem(...)`.
Parts of the code in the method were previously called `ApplyCmat`, which I liked. Since this the only occurrence right now, there may be no need to refactor though...

src/mechanics/solver/Solve.cpp
```diff
@@ -0,0 +1,37 @@
...
+                      int numIndependentDofs, double time, std::string solver)
```
TTitscher: 1) `numIndependentDofs`. Well. It may sound misleading that the constraints do not know the number of dofs. The numbering is done by `NuTo::DofNumbering`. So this is seems to be a conceptual flaw. Maybe having `Constraints::BuildConstraintMatrix(...)` is in the wrong class. Or the dof numbering should belong directly to the `Constraints` class. But since they _do not have to_, we chose to keep both things separated. Maybe there should be another class that handles both. Especially, since building the constraint matrix without a numbering is pointless... I'll think about that.

2) `time` could default to `0.`

test/mechanics/solver/MumpsSolver.cpp
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    Eigen::SparseMatrix<double> A;
+    Eigen::Vector3d b;
+    Eigen::Vector3d expected_x;
+};
+
+auto mumpsSolverNames = {"MumpsLU", "MumpsLDLT" };
+
+BOOST_DATA_TEST_CASE(mumpsSolvers, bdata::make(mumpsSolverNames), solver)
```
TTitscher: 1) First time I see the data test case, nice.

2) This test is very similar to `EigenSolver.cpp`. So maybe unite both? Or have a `LinearTestSystem.h`?

TTitscher (changes requested):
Solvers! I like. The only change I actually request is moving the solver stuff to `math`. And maybe, introduce a `enum eSolver` that replaces the strings.

Dockerfile
Psirus: I googled "making Docker images smaller". This is what came up.

scripts/cmake/CheckForDependencies.cmake
Psirus: It would have three parameters, the name of the package, the successful message and the warning message if not found. Doesn't seems this useful then. Or what do you propose?

scripts/cmake/CheckForDependencies.cmake
Psirus: "Home, Suite(Sparse), Home."

src/mechanics/solver/EigenSolver.h
Psirus: 1. Kay.
2. Could do, or separate the SuiteSparse solvers out as well, so that this file doesn't become too big; what do you think?
3. Kay.
4. a) Enum is annoying in Python again.
    b) What happens if e.g. I don't have SuiteSparse installed? Throw or Enum value is not available?

src/mechanics/solver/Solve.cpp
Psirus: If I have `Solve` function that takes a matrix and a vector, I would expect it to solve a linear system, no? `SolveOnce` doesn't make sense, why would I want to solve the thing twice? `SolveLinearSystem` is okay with me, but I still think sort of redundant: "`SolveLinearSystem(LinearSystem sys)`".

src/mechanics/solver/Solve.cpp
Psirus: 1. This wasn't a complaint, just an observation. Maybe one could get `numIndependentDofs` from the matrix?
2. That doesn't solve the problem imho.

test/mechanics/solver/MumpsSolver.cpp
Psirus: 1. Yeah, I liked that as well.
2. Can do.

src/mechanics/solver/Solve.cpp
TTitscher: Jip, my bad. I messed that up in my head (solve vs. newton iterations...). `Solve` is fine, I guess.

src/mechanics/solver/Solve.cpp
TTitscher: 1) Yeah right. `f.J[dof].rows()` should do.

src/mechanics/solver/EigenSolver.h
TTitscher: 2) This wrapper allows easy swap of the solvers without changing the includes. Otherwise I just include them in my main file manually. So I vote for including all solvers that have the eigen sparse solver concept. 
3) [_thanks, markdown enumeration numbering._]
4) We try hard to not base or design on a possible python interface, right? :ghost: And even in python, I would be happy to have some sort of autocompletion instead of guessing the solver name (and its capitalization - well, that could be solved...)

Psirus:
Okay, Enum vs. String, instead of long back and forth, maybe edit this:

Pro Enum:
- Autocomplete
- Misspelling reported by the compiler

Pro String:
- no additional type
- easier in Python
- headers don't change with available/unavailable solvers
- reading in a solver as a command line parameter / json entry is easier

test/mechanics/cell/PatchTest.cpp
```diff
@@ -282,29 +284,11 @@ BOOST_AUTO_TEST_CASE(PatchTestDispl)
...
-    BOOST_TEST_MESSAGE("GradientK \n" << gradient.K);
-
-    //      have a look at DISS_UNGER, page 28 for all that CMat stuff.
-    Eigen::MatrixXd Kmod = kJJ - CMat.transpose() * kKJ - kJK * CMat + CMat.transpose() * kKK * CMat;
-    Eigen::VectorXd Rmod = gradient.J[displ] - CMat.transpose() * gradient.K[displ];
-    Eigen::VectorXd RmodConstrained = (kJK - CMat.transpose() * kKK) * (-constraints.GetRhs(displ, 0));
-
-    Eigen::VectorXd newDisplacementsJ = Kmod.ldlt().solve(Rmod + RmodConstrained);
-    Eigen::VectorXd newDisplacementsK = -CMat * newDisplacementsJ + constraints.GetRhs(displ, 0);
+    int numIndependentDofs = dofInfo.numIndependentDofs[displ];
```
joergfunger: If we add this numIndependentDofs to the dof-class (or better the total number of dofs for this type), then this could be calculated directly.

joergfunger (approved):

TTitscher (approved):

Issue #141, opened 2017-12-03
PDE nodal values


TTitscher:
# Problem

a) 
- no concept for time derivatives of nodal values
- having `NodeMerge`
    - nodal values are some internal state, _impure_ functions

b)
- elements contain node references
    - not trivially copyable elements

# Solution

a)
- introduce a `SolutionVector` object that stores the nodal values, including time derivatives, values of previous time steps or whatever - [ nice idea of @joergfunger ]
- only store dof numbers at the node
- (have dedicated coordinate nodes that have coords)
- nodes could look like

~~~cpp
struct CoordinateNode
{
    Eigen::VectorXd mCoords;
};

struct DofNode
{
    std::vector<int> mDofNumbers;
};
~~~

b)
- replace node _references_ with node _values_ at the elements
    - duplicate information, but:
    - trivially copyable
    - no _central_ node container (no need for operations on all nodes)

### Interface changes

Mainly pass the `SolutionVector` to the cell/assembler. 

~~~cpp
int main()
{
    // ...
    SolutionVector v(numDofs);
    auto gradient = assembler.BuildVector(gradientFunction, v);
}

Vector Cell::Integrate(Function f, SolutionVector v)
{
    for (auto ip : mIps)
    {
        // ...
        CellData cellData(element, v);
        f(cellData, cellIpData); // same interface
    }
}
~~~

### Possible implementation

~~~cpp
NodalValues CellData::GetNodalValues()
{
    return Memoized(mElement.ExtractNodalValues(mSolutionVector));
}

NodalValues ElementInterface::ExtractNodalValues(SolutionVector v)
{
    NodalValues nodalValues
    for (auto node : mNodes)
        for (int dof : node.Dofs())
           nodalValues.push_back(v[dof]);
    return nodalValues;
}
~~~

# Discussion

- Sounds good?
- Naming? `SolutionVector`? `SolutionVectors`? `DofValues`? `DofValueVectors`? `...`?
- Will this work with IGA straight forward as well?
- It will introduce some more dependencies: The element alone can no longer calculate stuff, it always needs a solution vector passed to it. This may make tests a bit more annoying.

vhirtham:
Can you explain the problem a little bit more in detail for people who didn't join the discussion that took place before and who don't know every detail about the PDE  concept (yet)? Would be beneficial to get more people involved.

- Why is it a problem that nodes have an internal state and why is it beneficial to have another object / class that is needed?
[...]

PR #183, opened 2018-02-07
Pde quasistatic problem

TTitscher:
This PR includes a couple of new convenience classes that are all used to solving a quasistatic problem - 1d uniaxial truss with a local damage model. Those include:

- `CellStorage` - creates `NuTo::Cell` from FEM elements and stores them
- `Merger` - merges a dof vector to nodes, this class should be deleted once #141 is solved
- `EquationSystem` - could evolve into a common interface for nonlinear problems
- `QuasistaticSystem` - wraps the `EquationSystem` to a `NewtonRaphson` solvable problem including linear constraints and a trial state

Known problems:
- no support for time dependent integrands


applications/integrationtests/mechanics/QuasistaticProblem1D.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    auto cellGroup = cells.AddCells(mesh.ElementsTotal(), integrationType);
+    auto law = ConstitutiveLaw(cellGroup.Size(), integrationType.GetNumIntegrationPoints());
+
+    // impose imperfection:
+    const int imperfectionCell = numElements / 2;
+    const int imperfectionIp = law.mEvolution.Ip(imperfectionCell, 0);
+    law.mEvolution.mKappas[imperfectionIp] = initialKappa;
+
+    Integrands::MomentumBalance<1> momenumBalance(disp, law);
```
Psirus: typo

src/mechanics/tools/Merger.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+#include "mechanics/dofs/GlobalDofVector.h"
+#include "mechanics/mesh/MeshFem.h"
+
+namespace NuTo
+{
+//! Performs our good old "NodeMerge" and should be replaced by any solution from issue #141 PDE nodal values
+class Merger
```
Psirus: Not sure if it is worth it, considering it is getting replaced, but I would prefer something more descriptive, e.g. `NodalValueMerger`.

src/mechanics/tools/QuasistaticProblem.h
```diff
@@ -0,0 +1,69 @@
...
+
+    //! sets the global time required for evaluating the constraint right hand side
+    //! @param globalTime global time
+    void SetGlobalTime(double globalTime);
+
+    //! builds the trial system where its residual contains forces equivialent to the applied constraints from time step
+    //! t_n to t_n+1
+    //! @param x independent dof values corresponding to globalTime
+    //! @param globalTime t_n
+    //! @param timeStep t_n+1 - t_n
```
Psirus: t_n+1 - t_n == 1 👻

Psirus (approved):

src/mechanics/tools/EquationSystem.h
```diff
@@ -0,0 +1,42 @@
...
+#include "mechanics/dofs/GlobalDofVector.h"
+#include "mechanics/dofs/GlobalDofMatrixSparse.h"
+
+#include "mechanics/tools/NodalValueMerger.h"
+
+namespace NuTo
+{
+
+//! Equation system that contains Gradient(u) and its derivative with respect to u
+class EquationSystem
```
joergfunger: IMO, this is a method specific for static problems that can be solved with a "static" time integration scheme, thus I would call this "StaticProblem". I would not try to make this more general, but then also have another "equation system" that is named "NewmarkProblem".

src/mechanics/tools/EquationSystem.h
```diff
@@ -0,0 +1,42 @@
...
+    EquationSystem(MeshFem* rMesh);
+
+    void SetDofInfo(DofInfo dofInfo);
+
+    void AddGradientFunction(Group<CellInterface> group, CellInterface::VectorFunction f);
+    void AddHessian0Function(Group<CellInterface> group, CellInterface::MatrixFunction f);
+    void AddUpdateFunction(Group<CellInterface> group, CellInterface::VoidFunction f);
+
+    GlobalDofVector Gradient(GlobalDofVector dofValues, std::vector<DofType> dofs);
+    GlobalDofMatrixSparse Hessian0(GlobalDofVector dofValues, std::vector<DofType> dofs);
```
joergfunger: For static problems, there is only a single hessian, thus I would remove the zero. And also add t.

src/mechanics/tools/EquationSystem.h
```diff
@@ -0,0 +1,42 @@
...
+public:
+    EquationSystem(MeshFem* rMesh);
+
+    void SetDofInfo(DofInfo dofInfo);
+
+    void AddGradientFunction(Group<CellInterface> group, CellInterface::VectorFunction f);
+    void AddHessian0Function(Group<CellInterface> group, CellInterface::MatrixFunction f);
+    void AddUpdateFunction(Group<CellInterface> group, CellInterface::VoidFunction f);
+
+    GlobalDofVector Gradient(GlobalDofVector dofValues, std::vector<DofType> dofs);
```
joergfunger: I would add the new time t as a parameter here, the time of the previous update is stored as a member variable, thus dt can be directly computed.

applications/integrationtests/mechanics/QuasistaticProblem1D.cpp
```diff
@@ -0,0 +1,224 @@
...
+        double timeStep = 0.01;
+        double maxTimeStep = 0.1;
+
+        int iStep = 0;
+        while (globalTime < tEnd && timeStep > 1.e-6)
+        {
+            std::cout << "Step " << iStep << " at t = " << globalTime << std::endl;
+            try
+            {
+                auto trialSystem = mProblem.TrialSystem(x, globalTime, timeStep);
```
joergfunger: Conceptually, I would prefer something like 
~~~cpp
StaticProblem myproblem(FunctionSpace myFunctionSpace); 
//this FunctionSpace essentially describes the dofs to be solved for and the 
//  corresponding dimensions of the solution vector (only including the dofs th ebe solved for)
//it also stored the current state (for static this is just d, for Newmark type 
//  problems this would be d,v,a)
//This class also provides methods for setting the initial conditions

StaticSolver.DoStep(myProblem, t, boundaryConditions);
~~~

The calculation of the trial state is then done in the staticSolver DoStep method. Other parameters (such as the solver, line search etc.) are passed either as template or as parameter to the StaticSolver.

joergfunger (changes requested):
Good starting point, though I'm still confused with the general procedure on the main level. I think the interfaces are difficult to understand and we expose to many details to the "end user".

src/mechanics/tools/EquationSystem.h
TTitscher: I find a renaming useful. But I would have a single `TimeDependentProblem` that I can plug into various other classes. Like `QuasistaticSomething` that simply irgnores `Hessian1/2`.

src/mechanics/tools/EquationSystem.h
TTitscher: I realized that our mechanics integrands need a `dt`. I can imagine some time dependent neumann boundary integrand that uses `t`. So we have to be very careful when binding these functions.

applications/integrationtests/mechanics/QuasistaticProblem1D.cpp
TTitscher: I thought about that and liked it until I wanted to implement it. If we chose to do `DoStep`, we have a lot of options to pass that are more suitable as template parameters. E.g. the solver or the solution scheme. So the default case would be to use a Newton-Raphson algorithm in this `DoStep`. Now it is not trivial do specify the solver. Another option is a different iteration scheme. What if I want to use a fix-point iteration or L-BGFS instead of Newton-Raphson.

Maybe providing `DoStep` as an additional method (and keeping the others as well) could be an option.

src/mechanics/tools/EquationSystem.h
joergfunger: Renaming to TimeDependentProblem is fine. However, if the only difference is Hessian1/2, you could also define a static problem and then just derive from this another Newmark problem that has Hessian 1/2. Thus, we would avoid to allow the latter methods being called for a static problem. For our explicit methods, we would additionally need a method that returns a diagonal hessian.

src/mechanics/tools/EquationSystem.h
joergfunger: The Static problem should store the last update state (history data is already stored, but the corresponding dof_state as well as the time should be a member variable). Thus, the Hessian and well as the gradient function and update function should IMO include t. dt is then automatically computed in the problem and passed to the integrands. I also still think that the interface for the gradienthessian functions for the integrands that we use in a Newmark scheme should be fixed (i.e. gradient is a function of d,v,a,t,dt and cell/ipdata). In this static case, it would be d,t,dt,cell/ip data). Why is the dof type (dofs) a parameter? Wouldn't it make sense to store this as a member variable that is initialized when creating the problem?

src/mechanics/tools/QuasistaticProblem.h
```diff
@@ -0,0 +1,78 @@
...
+
+namespace NuTo
+{
+class QuasistaticProblem
+{
+public:
+    //! Ctor
+    //! @param equations system of equations including Gradient(), Hessian0() and UpdateHistory()
+    //! @param dof dof type
+    QuasistaticProblem(EquationSystem& equations, DofType dof);
```
joergfunger: In this special situation, where we essentially do not need any time integration scheme, this works well. However, for generalization purposes, I think QuasiStaticProblem should also provide a Hessian and Gradient method that is then called in the DoStep method (where itself the newton iteration is used). However, this can certainly also done later when trying to generalize this. Futhermore, I don't fully understand the need for a distinction between EquationSystem (apparently used to call the Newton iteration) and the QuasiStaticProblem (which additionally includes the constraints). But this is probably due to me thinking of throwing out the constraints from the problem definition.

src/mechanics/tools/QuasistaticSolver.h
```diff
@@ -0,0 +1,86 @@
...
+    //! @param equations system of equations including Gradient(), Hessian0() and UpdateHistory()
+    //! @param dof dof type
+    QuasistaticSolver(TimeDependentProblem& equations, DofType dof);
+
+    //! @param constraints linear constraints
+    void SetConstraints(Constraint::Constraints constraints);
+
+    //! sets the global time required for evaluating the constraint right hand side
+    //! @param globalTime global time
+    void SetGlobalTime(double globalTime);
```
joergfunger, line 22: Shouldn't the methods that change the state be private (Update, SetTime)? They should IMO only be called by the DoStep method. Similarly, I'm not yet sure how to interpret the setting of mX in this context (not yet implemented due to the assumption of zero initial conditions). On the one hand, this also should be private, but this would not allow to set initial conditions.

joergfunger: I also think it would be a good idea to split the classes (not required now in this pull request) between the solver and the problem. So the DoStep method of the solver should be called with a specific problem.

joergfunger (approved):
Next step to a working newNuto.

src/mechanics/tools/QuasistaticSolver.h
TTitscher: I kept the methods to manually implement my custom `DoStep` that, for example, does not use an newton raphson algorithm to find the equilibrium. Or one that allows further configuration of the solver. But you are right, calling `SetGlobalTime()` in between `DoStep` will mess things up and has to be avoided. I'll think about that.

src/mechanics/tools/QuasistaticProblem.h
TTitscher: (You commented on an outdated file.)
But I agree with your point of having Hessian and Gradient as methods. They currently do exist and are called `Residual` and `Derivative`. The problem here is that the newton algorithm does not care about time. So the `Residual` function is called by the newton raphson algorithm with `x` as the only argument. This method calls `Gradient(x+constraints, t)` from the TimeDependentProblem and the time `t` and the constraints have to be provided somehow. This makes everything a bit messy.

applications/integrationtests/mechanics/QuasistaticProblem1D.cpp
```diff
@@ -0,0 +1,203 @@
...
+    using Law = Laws::LocalIsotropicDamage<1, DamageLaw, Evolution, ElasticLaw>;
+
+    MeshFem mMesh;
+
+    DofType mDof;
+    Law mLaw;
+    Integrands::MomentumBalance<1> mMomentumBalance;
+
+    TimeDependentProblem mEquations;
+    QuasistaticSolver mProblem;
```
joergfunger, line 136: This naming here is confusing the class TimeDependent problem is mEquations, and the Solver is called problem

src/math/NewtonRaphson.h
```diff
@@ -27,6 +28,33 @@ struct DoubleSolver
     }
 };
 
+//! @brief Wrapper for Eigen::SparseMatrix
```
joergfunger, line 31: Why do we need this "wrapping"?

src/mechanics/tools/CellStorage.cpp
```diff
@@ -0,0 +1,16 @@
+#include "mechanics/tools/CellStorage.h"
+#include "mechanics/cell/Cell.h"
+
+using namespace NuTo;
+
+Group<CellInterface> CellStorage::AddCells(Group<ElementCollectionFem> elements,
+                                           const IntegrationTypeBase& integrationType, int cellStartId)
```
joergfunger, line 7: how do we ensure that this id is then unique?

src/mechanics/tools/QuasistaticSolver.cpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+void QuasistaticSolver::SetGlobalTime(double globalTime)
+{
+    mTimeStep = globalTime - mGlobalTime;
+    mGlobalTime = globalTime;
+}
+
+std::pair<Eigen::SparseMatrix<double>, Eigen::VectorXd>
+QuasistaticSolver::TrialSystem(const Eigen::VectorXd& x, double globalTime, double timeStep)
```
joergfunger, line 31: Why do you need global time and timeStep, can't you compute the timestep from the (new) global time and the time stored during the last update? I also would not write a setTime function, only change the time when an update is made.

src/mechanics/tools/TimeDependentProblem.cpp
```diff
@@ -0,0 +1,86 @@
+#include "mechanics/tools/TimeDependentProblem.h"
+#include "mechanics/dofs/DofNumbering.h"
+
+using namespace NuTo;
+
+TimeDependentProblem::TimeDependentProblem(MeshFem* rMesh)
```
joergfunger, line 6: Why does Time dependent problem does only work for FEM? IMO, there should not be a difference between IGA and FEM when it comes to the solution procedure.

src/math/NewtonRaphson.h
TTitscher: Our newton raphson algortihm requests a solver with the interface `Vector Solve(Matrix, Vector)` which is not provided by `Eigen` out of the box. Thus, it is wrapped here. But this is _very_ similar to the [`EigenSparseSolver`](https://github.com/nutofem/nuto/blob/PDE_reviewed/src/math/EigenSparseSolve.h#L38) and could probably be removed.

src/mechanics/tools/TimeDependentProblem.cpp
TTitscher: We need some way to access nodal values for `NodeMerge`. Passing the mesh was the easiest. And currently, there is no `MeshIga` or `MeshInterface`. Plus, there are plans to get rid of `NodeMerge` and this is would not be required anymore.

applications/integrationtests/mechanics/QuasistaticProblem1D.cpp
TTitscher: That was misses during renaming, I guess. However, I am not happy with names `TimeDependentProblem` and `QuasistaticSolver` either. :cry: 
I expect a huge redesign when we tackle #141, where we can rethink those names.

src/mechanics/tools/CellStorage.cpp
TTitscher: It is the users task to match the cell numbering with his history data numbering. If we want to make that a library task, we have to impose some common interface on all history data classes (to size/resize them). This may be too much of a restriction.

PR #184, opened 2018-02-07
Add dof interpolation addon

pmueller2:
Tries to address dof interpolation for meshes with different element types. Now you can provide a group of elements (ElementCollections to be exact, like those you get from Gmsh import as physical groups) for which the new DoF element layer is created.


src/mechanics/mesh/MeshFem.cpp
```diff
@@ -99,7 +99,18 @@ Group<NodeSimple> MeshFem::NodesTotal(DofType d)
 {
     Group<NodeSimple> group;
     for (auto& element : this->Elements)
-        for (int iNode = 0; iNode < element.DofElement(d).Interpolation().GetNumNodes(); ++iNode)
+    {
+        const ElementFem* dofElm;
+        try
+        {
+            dofElm = &(element.DofElement(d));
```
TTitscher: Making try/catch blocks part of common logic is bad practice IMO. I suggest adding a `ElementColleciton::Has(DofType)`. That will also be much clearer and there is no need to have this `dofElm` pointer.
~~~cpp
if(not element.Has(d))
   continue;
...
~~~

Another option is to avoid this code here and throw an exception as before. As far as I can tell, you copied the code to `MeshFemDofConvert::NodePointsTotal` anyways. :yum: 
Then, my comment above would apply there.

src/mechanics/mesh/MeshFemDofConvert.cpp
```diff
@@ -138,6 +138,74 @@ SubBoxes<NodePoint>::Domain SetupSubBoxDomain(const NuTo::MeshFem& mesh, int num
...
+                subBoxes.Add(NodePoint(coord, node));
+                nodesForTheNewlyCreatedElement.push_back(&node);
+            }
+        }
+        elementCollection.AddDofElement(dofType, ElementFem(nodesForTheNewlyCreatedElement, interpolation));
+    }
+}
+
+
 void NuTo::AddDofInterpolation(NuTo::MeshFem* rMesh, DofType dofType,
```
TTitscher: A lot of code duplication here. I suggest making the existing implementation a special case of your new one. So
~~~
void OldAddDof(mesh, dof, interpolation)
{
    auto elements = mesh.ElementsTotal();
   NewAddDof(mesh, dof, elements, interpolation);
}
~~~

src/mechanics/mesh/MeshFem.cpp
```diff
@@ -99,7 +99,18 @@ Group<NodeSimple> MeshFem::NodesTotal(DofType d)
 {
     Group<NodeSimple> group;
     for (auto& element : this->Elements)
-        for (int iNode = 0; iNode < element.DofElement(d).Interpolation().GetNumNodes(); ++iNode)
+    {
+        const ElementFem* dofElm;
+        try
+        {
```
pmueller2: I did not want to enclose the whole loop in a try catch block
because I am only looking for the out of range exception thrown 
by the DofElement getter.

src/mechanics/mesh/MeshFemDofConvert.cpp
```diff
@@ -138,6 +138,74 @@ SubBoxes<NodePoint>::Domain SetupSubBoxDomain(const NuTo::MeshFem& mesh, int num
     return d;
 }
 
+//! @brief returns all Dof Nodes with their coordinates (NodePoints).
+std::vector<NodePoint> NodePointsTotal(NuTo::MeshFem* rMesh, NuTo::DofType dofType)
+{
```
pmueller2: Needed because nodes do not have coordinates.

TTitscher (changes requested):
Nice idea to fill the subbox with all existing nodes! If you get rid of the code duplication, I happily approve!

src/mechanics/mesh/MeshFem.cpp
pmueller2: The underlying container is a DoFContainer and this one uses a map to store its data. The public available accessors all use map.at() to get the data and I thought this implies that catching exceptions is the way to go. If I have to use find (Has(d) ) to know if the element is there anyway then it might makes sense to access the items also using find - but this is not implemented. Why?

TTitscher: Catching exceptions should be an exception... Why is `ElementCollection::Has(d)` not implemented? For the same reason as everything else: YAGNI. And now that we needed, we could add it. 
[`bool DofContainer::Has(d)` exists]

pmueller2: Öhm ok. I meant Find, but understood.

src/mechanics/mesh/MeshFem.cpp
```diff
@@ -100,17 +100,11 @@ Group<NodeSimple> MeshFem::NodesTotal(DofType d)
     Group<NodeSimple> group;
     for (auto& element : this->Elements)
     {
-        const ElementFem* dofElm;
-        try
+        if (element.Has(d))
```
TTitscher: I personally would reverse the logic. Like return/continue as early as possible. This also reduces the level of indentation. This may be taste however.

TTitscher (approved):

PR #189, opened 2018-02-12
Pde fix inline

TTitscher:


pmueller2:
Ok, Why do you think this is a bug? All these functions are tied to engineering strain and will not occur somewhere else.

TTitscher:
Well, a better label would have been "bugfix". The missing inlines are the bug and cause multiple definitions.

pmueller2 (approved):

PR #190, opened 2018-02-13
Pde visualize helper

TTitscher:
`Visualize::PostProcess`
- stores arbitrary `Visualizer` by name
- stores arbitrary visualize functions (dofs, cell data functions, point functions) for each stored `Visualizer`
- visualizes everything at once, including a `pvd` file


src/visualize/PostProcess.h
```diff
@@ -0,0 +1,90 @@
...
+        Visualizer mVisualizer;
+        std::vector<DofType> mDofs;
+        std::vector<std::pair<std::function<Eigen::VectorXd(const CellData&, const CellIpData&)>, std::string>>
+                mCellFunctions;
+        std::vector<std::pair<std::function<Eigen::VectorXd(Eigen::VectorXd)>, std::string>> mPointFunctions;
+    };
+
+    std::map<std::string, VisualizationInfo> mVisualize;
+
+    std::string mResultDir = "not set";
```
Psirus: Why not simply `std::string mResultDir;` and the check `if (mResultDir == "not set")` becomes `if (mResultDir.empty())`?

Psirus (approved):

src/visualize/PostProcess.h
TTitscher: Ouf, yeah. I thought it would make sense to allow writing files in the current directory, so `mResultDirectory = ""`. As it is implemented right now, `ResultDirectory(dir)` deletes the directory and creates a new one. So that would not make sense. I'll fix.

PR #191, opened 2018-02-14
Pde lumped mass

pmueller2:
Adds mass lumping. Procedure works as follows: 
- compute local (full) matrix
- take only the diagonal entries
- scale all entires with a factor so that the sum of all matrix entries is the same as for the full matrix
- return the diagonal

To be used for the mass matrix in explicit time integration schemes.


src/mechanics/cell/SimpleAssembler.h
```diff
@@ -20,6 +20,17 @@ class SimpleAssembler
...
                                       CellInterface::MatrixFunction f) const;
 
+    //! @brief Assembles a diagonally lumped matrix from local matrices calculated by f
+    //! @param cells group of cells to be used for assembly
+    //! @param dofTypes vector of dofTypes
+    //! @return a global dof vector that represents a collection of diagonal matrices
+    //! @remark HRZ lumping is used here, the assumptions made are:
+    //! - shape functions sum to 1, only then the total mass for a cell can be calculated by
+    //!   summing over all entries of the local mass matrix
+    //! - the total mass of a cell is the same for all components of the considered dof
```
TTitscher, line 30: Do you think it is required to somehow check these assumptions?

TTitscher (approved):
:+1:

src/mechanics/cell/SimpleAssembler.h
pmueller2: I don't want to and leave it to the user and the doc.
1) Partition of unity (first assumption) is also essential elsewhere and this might not be the ideal place to test for it
2) Currently vector valued (multicomponent) dofs use the same interpolation for each component. If this changes one has to rethink the algorithms at several places.
-> so these checks might be included later

joergfunger (approved):

PR #195, opened 2018-02-21
Pde jacobian adds

pmueller2:
Adds a getter for the Jacobian and computes its inverse also for embedded 2D in 3D / 1D in 2D elements.
In what way exactly the Jacobian is used is left to the user.
Access to the Jacobian is needed e.g. for boundary conditions that need normal and tangential directions.


Psirus:
https://git-scm.com/book/en/v2/Git-Branching-Rebasing

test/mechanics/cell/Jacobian.cpp
```diff
@@ -85,6 +85,28 @@ BOOST_AUTO_TEST_CASE(Jacobian2Din3D)
...
+    Eigen::MatrixXd B = NuTo::ShapeFunctions2D::DerivativeShapeFunctionsTriangleOrder1(Eigen::Vector2d::Zero());
+    Eigen::VectorXd coordinates = Eigen::VectorXd(9);
+    coordinates << p1, p2, p3;
+
+    NuTo::Jacobian jacobian(coordinates, B, 3);
+
+    Eigen::Vector3d basisVec1 = jacobian.Get().col(0);
+    Eigen::Vector3d basisVec2 = jacobian.Get().col(1);
+
+    Eigen::Vector3d normal = basisVec1.cross(basisVec2).normalized();
```
TTitscher: Is seems to be equivalent to 
~~~cpp
Eigen::Vector3d normal = jacobian.Get().col(2);
~~~
right?

TTitscher (approved):
Why not. :+1:

test/mechanics/cell/Jacobian.cpp
pmueller2: That would be right if the returned jacobian had the magic 3rd column. Instead I chose to remove this feature because there were some reservations and now the jacobian is 2x3 and its inverse 3x2 and the (re)calculation of the normal has to be done elsewhere, e.g. in the boundaryloadfunction if needed.

TTitscher: I am thinking about making this normal calculation more accessible.
- Maybe we could move the normal code to a free function - that only accepts _incomplete_ jacobians and throws otherwise
- Do you think the `jacobian.Get()` can be used for something else than calculating the normal? If that is unlikely, we could replace `.Get()` with `.Normal()`. Or simply add `.Normal()` to `CellIpData`?

pmueller2: Usage of the Jacobian:
1) If you need a basis in tangential direction at some point you can use the two columns of the jacobian rather than computing one from the normal. In a conforming mesh a adjacent element will have the same basis vectors at a shared face.
2) If you want to get local coordinates from global ones in a higher order element you can use Newton iterations and the Jacobian will be the required derivative.

Issue #197, opened 2018-02-21
Unite CellData and CellIpData


TTitscher:
Upon implementing some integrands/laws/visualize functions, I realized that `CellData` and `CellIpData` are a bit cumbersome to use. 

### Problem

You almost never need access to the vector of nodal values in the integrand. You actually want your dofs or their derivatives interpolated an certain integration points.

### Solution

Put `CellData` into `CellIpData` and add interpolation methods. So this
~~~cpp
BMatrixStrain B = cellIpData.GetBMatrixStrain(mDofType);
NodeValues u = cellData.GetNodeValues(mDofType);
NuTo::EngineeringStrain<TDim> strain = B * u;

NMatrix N = cellIpData.GetNMatrix(mDofType);
Eigen::VectorXd dofAtIp = N * u;
~~~

will become this
~~~cpp
NuTo::EngineeringStrain<TDim> strain = cellIpData.InterpolateStrain(mDofType);
Eigen::VectorXd dofAtIp = cellIpData.Interpolate(mDofType);
~~~

This will also reduce the argument list of the cell interface to only one argument. We could also think about putting `CellData::GetCellId()` and `CellIpData::GetIpId()` together in a `std::pair<int, int>` because they are mostly passed together. 

### Discussion

Rename the class `CellIpData` to something else?
How to name its methods?
- `Interpolate(dof), InterpolateGradient(dof), InterpolateStrain(dof)`? or simply
- `operator()(dof), Gradient(dof), Strain(dof)`?

Do you like the `std::pair<int, int>` idea?

Psirus:
Hm. In general, I like the idea, and I agree that most of the time, you probably just want the interpolated value. However, it clashes with another change I would like to propose^^

Personally, I don't like the fact that there is a `GetBMatrixStrain` and a `GetBMatrixGradient` at the `CellIpData`. I would rather just have a `GetDerivativeShapeFunctions` (or `GetDNMatrix`) and a couple of free functions, that do the transformation, .i.e
~~~cpp
auto dN = cellIpData.GetDerivativeShapeFunctions(mDofType);
auto B = StrainOperator(dN);
NodeValues u = cellData.GetNodeValues(mDofType);
NuTo::EngineeringStrain<TDim> strain = B * u;

auto B2 = GradientOperator(dN);
auto B3 = DivergenceOperator(dN);
// ...
~~~

The only way I see to combine both would be to pass an "operator" to the cell,
~~~cpp
[...]

Issue #124, opened 2017-10-24
PDE renaming


TTitscher:
There are some issues regarding names. 

___Feel free to edit this post to add new entries___

---

### Strange names caused by name clashes with _old NuTo_:
- [ ] `NodeSimple` --> I suggest `NodeDynamic` - so we can later add a `NodeFix`.
- [ ] `InterpolationSimple` --> I suggest moving this to `namespace InterpolationFem` and call it `class NuTo::InterpolationFem::Interface`.

### Strange names caused by inability to find better ones:
- [ ] `DofCalcContainer` (suggest calculations on dofs rather than its elements)
- [ ] `ElementCollectionXYZ` (contains a coodinate element and multiple dof elements)
    - `ElementLayers`
    - `ElementStack`
- [ ] `namespace Nabla` for differential operators. But IMO `DifferentialOperators` is too long. Maybe [`del`](https://en.wikipedia.org/wiki/Del)? Rather unknown, I whould say...

### I don't know the reason, it is just strange:
- [ ] ...

### inconsistencies 
- singular or plural namespaces and directories?
    - [ ] `mechanics/cell` or `mechanics/timeIntegration` vs `mechanics/constraints` or `mechanics/elmeents`

---

This is a rather simple issue, since it mostly involves renaming. But you get to know our code a bit better and get used to this whole `git`, `branch`, `pull request`, `travis` thing. So I would like to see some contributors other than _the usual suspects_. Don't hesitate to ask, if you need help!

Psirus:
I'd prefer the `AdjectiveNoun` variant, as in `DynamicNode` and `StaticNode`, because I think it reads more naturally.

vhirtham:
Not sure which order to prefer. Let's say, I have no idea about all the existing Node classes and want to check my options. `NounAdjective`  ordering automatically groups them for me and I won't miss one.

Another candidate for renaming:
`ConstraintCompanion` --- What does a companion do? The name does not specify its purpose, I think.
Maybe something like `ConstraintAssistant` or `ConstrainingAssistant` or `ConstraintBuilder` or `ConstraintConstructor` or `ConstraintAssembler` or `<Your ideas>` 

For `ElementCollectionXYZ` maybe something like `ElementSetXYZ` or `SpatialElementsSet`?

TTitscher:
[...]

PR #199, opened 2018-02-22
Pde unite celldata

TTitscher:
Implements the ideas of #197 

`CellData` is removed from `CellInterface` and passed to `CellIpData`. This means:
- less boilerplate when writing custom `CellInterface` functions 
- convenient access to interpolation methods like
~~~cpp
ScalarDofType dofT("Temperature");
double T = cellIpData.Value(dofT);
auto gradT = cellIpData.Apply(dofT, Nabla::Gradient());

// you can specify which differential operator is used, currently 
// Nabla::Gradient() --> default
// Nabla::Strain()
auto strain = cellIpData.Apply(dofDisplacements, Nabla::Strain());
~~~
- You can define custom B matrices by implementing [`Nabla::Interface`](https://github.com/nutofem/nuto/blob/PDE_unite_celldata/src/mechanics/cell/DifferentialOperators.h#L11)

As this interface change requires a lot of repetitive adaptions, the PR is quite large. You may want to have a look at early commits separately.


Psirus:
Code looks fine, but I don't like the name `Gradient`. Doing e.g. `cellIpData.Gradient(magneticField, B::Curl)` is weird. So I'd prefer something else. Maybe `ApplyOperator` or at least `Derivative`. Also, the namespace might be named `Operators`, so for example `Operators::Curl`.

src/mechanics/cell/GradientOperators.h
```diff
@@ -0,0 +1,93 @@
+#pragma once
```
pmueller2: Maybe change the filename to Differentialoperators?

src/mechanics/constitutive/LinearElastic.h
```diff
@@ -37,12 +37,12 @@ class LinearElastic : public MechanicsInterface<TDim>
         mC = CalculateC(mE, mNu, planeState);
     }
 
+    static MechanicsTangent CalculateC(double E, double Nu, ePlaneState planeState = ePlaneState::PLANE_STRESS);
+
```
pmueller2, line 41: Hey thanks! Good that it is accessible from the outside now.

pmueller2 (approved):
Ok. Maybe rename the GradientOperators file, because there are also operators that are not gradients, e.g. divergence and the name Gradient is somehow overused in NuTo already. Apart from this: Yup.

src/mechanics/constitutive/LinearElastic.h
TTitscher: :kiss:

src/mechanics/cell/DifferentialOperators.h
```diff
@@ -0,0 +1,93 @@
+#pragma once
+
+#include <Eigen/Core>
+#include "base/Exception.h"
+
+namespace NuTo
+{
+
+namespace Nabla
```
joergfunger, line 9: IMO, nabla is a specific operator, this naming of the namespace might lead to confusion

src/mechanics/integrands/MomentumBalance.h
```diff
@@ -23,26 +23,25 @@ class MomentumBalance
...
     {
-        BMatrixStrain B = cellIpData.GetBMatrixStrain(mDofType);
-        NodeValues u = cellData.GetNodeValues(mDofType);
         DofVector<double> gradient;
 
-        NuTo::EngineeringStrain<TDim> strain = B * u;
-        gradient[mDofType] = B.transpose() * mLaw.Stress(strain, deltaT, cellData.GetCellId(), cellIpData.GetIpId());
+        BMatrixStrain B = cellIpData.B(mDofType, Nabla::Strain());
+        gradient[mDofType] =
+                B.transpose() * mLaw.Stress(cellIpData.Apply(mDofType, Nabla::Strain()), deltaT, cellIpData.Ids());
```
joergfunger, line 32: This approach really requires to memoize all operators (in this case B). What is the advantage compared to the "old" version that (at least from my point of view) easier to understand?

joergfunger (commented):
Simplifies the interface, though I would rename to nable namespace to something else (e.g. as in the test files where you called it differentialOperators)

src/mechanics/cell/DifferentialOperators.h
TTitscher: Added to #124.

src/mechanics/integrands/MomentumBalance.h
TTitscher: I do not fully understand the question. 
Do you question the memoization itself? Then I'd say that memoizing `B` and `N` is the whole point of `CellIpData`. 
Or do you question the naming from `GetBMatrixStrain(dof)` to `B(dof, Nabla::Strain())`. Then I'd say that it
- provides a name for different type of `B`s to be used in the `Apply` function
- removes the `B` responsibility from `CellIpData`
- allows to provide custom `B` implementations from outside libnuto

Or do you question the naming of `Apply`? (But then, there was no equivalent in the "old" version...)

Issue #17, opened 2016-11-25
ARPACK wrapper 


TTitscher:
When activating ENABLE_ARPACK, FindArpack.cmake searches for a file "Arpack/Arpack.h". The `class EigenSolverArpack` includes it. This file was previously created by the [ARPACK.cmake.patch](https://sourceforge.net/p/nuto/wiki/Installation/) as a wrapper to the ARPACK fortran code. The CMake file also compiles a libArpack.a.

I naively included the "Arpack.h" (created by this patch) and linked the libarpack.a from the debian repository. Did not work. Help appreciated. :)

TTitscher:
OK. Turns out I am an idiot and forgot how to properly link. 
~~~cpp
/* main.cpp */
#include "Arpack.h"
int main()
{
   return dnaupd_ (0,0,0,0,0, 0,0,0,0,0, 0,0,0,0,0, 0);
}
~~~
and
~~~
g++ main.cpp -L/usr/lib -larpack
~~~
compiles just fine. Trying to include that in NuTo now.

TTitscher:
Fixed by https://github.com/nutofem/nuto/commit/97102c8aff8dfc183b0f2b49ac3d45f8ba69eece

TTitscher:
There are still issues with the memory. Calculating the eigenvalues/eigenvectors in each iteration of a time integration scheme causes the memory to fill rapidly. 

Maybe switch to
https://eigen.tuxfamily.org/dox-devel/unsupported/group__ArpackSupport__Module.html?

Priority low, I guess...

TTitscher:
We can remove our own implementation and use [`Eigen-unsupported`](https://eigen.tuxfamily.org/dox/unsupported/ArpackSelfAdjointEigenSolver_8h_source.html)

Psirus:
Fixed by #202.

PR #202, opened 2018-02-23
Pde arpack

TTitscher:
Just a brief example that shows the basic usage of `<unsupported/Eigen/ArpackSupport>`. I think, this solves #17.

### DIFF
--- a/applications/integrationtests/math/Arpack.cpp
+++ b/applications/integrationtests/math/Arpack.cpp
@@ -59,13 +59,11 @@ BOOST_AUTO_TEST_CASE(ArpackUsage)
     Eigen::ArpackGeneralizedSelfAdjointEigenSolver<Eigen::SparseMatrix<double>> arpack;
     Eigen::SparseMatrix<double> A = GetA();
 
-    // find smallest EV: - smallest magnitude ("SM") or smallest amplitude ("SA")
+    // find smallest EV: - smallest magnitude ("SM")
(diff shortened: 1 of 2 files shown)


applications/integrationtests/math/Arpack.cpp
```diff
@@ -0,0 +1,88 @@
...
+    return Eigen::SelfAdjointEigenSolver<Eigen::MatrixXd>(A).eigenvalues();
+}
+
+BOOST_AUTO_TEST_CASE(ArpackUsage)
+{
+    Eigen::ArpackGeneralizedSelfAdjointEigenSolver<Eigen::SparseMatrix<double>> arpack;
+    Eigen::SparseMatrix<double> A = GetA();
+
+    // find smallest EV: - smallest magnitude ("SM") or smallest amplitude ("SA")
+    BOOST_CHECK_CLOSE(arpack.compute(A, 1, "SM", Eigen::EigenvaluesOnly).eigenvalues()[0], EVs(A).minCoeff(), 1.e-10);
```
pmueller2: SA / LA means smallest/largest algebraic eigenvalue (I think meant is the algebraic multiplicity, that is: the number of equal values)

pmueller2 (changes requested):
I like this, especially because I need it! Only fix the documentation SA/LA (and if you want add a test).

pmueller2 (approved):

PR #204, opened 2018-02-26
Update Dockerfile to fix coverage problem

Psirus:
Ever since #182, there no longer was any coverage information. That's because the codecov script is pulled via https, but curl was missing this certificates package. Anyway, should work now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ matrix:
 
 before_install:
     - CI_ENV=`bash <(curl -s https://codecov.io/env)`
-    - travis_retry timeout 120 docker pull nuto/nuto_docker:smaller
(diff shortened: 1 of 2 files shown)


TTitscher (approved):
Nice! Thanks! Just for the records:
- How did you find out that `ca-certificates` is missing? When I looked at e.g. the travis output, I did not even see an error message...
- How can we preserve this knowledge? Something like a wiki entry `Fix coverage checklist`?

Psirus:
- The flag `-s` in the invocation of curl stands for silent. Not sure if that is required, we only copied that from codecov. Now, when everything works fine, it only suppresses a progress bar or something. But when something fails, it also suppresses the error message. I don't remember it anymore, but googling it quickly led to "Do `sudo apt install ca-certificates`!", which I obeyed, and found that it was good.
- Dunno. Feel free to. Next time it's probably something different.^^

PR #206, opened 2018-02-27
Pde minor fixes

TTitscher:


applications/benchmark/ShapeFunctionMemoization.cpp
```diff
@@ -109,8 +109,8 @@ struct Lobatto
     }
 };
 
-auto testFunction = NuTo::ShapeFunctions3D::DerivativeShapeFunctionsBrickSpectralOrder4;
-using result = Eigen::Matrix<double, 125, 3>;
+auto testFunction = NuTo::ShapeFunctions3D::DerivativeShapeFunctionsTetrahedronOrder2;
+using result = Eigen::MatrixXd;
```
Psirus: Now, `result` can just be replaced by `Eigen::MatrixXd` at the appropriate places. This just causes confusion imho.

src/mechanics/tools/AdaptiveSolve.cpp
```diff
@@ -54,6 +57,16 @@ void AdaptiveSolve::Solve(double tEnd)
                                 "Time step smaller than prescribed minimal time step " + std::to_string(dtMin) + ".");
             continue; // without updating the global time
         }
+        catch (Exception& e)
+        {
+            std::cout << e.what() << std::endl;
+            return;
+        }
+        catch (std::exception& e)
```
Psirus: Imho only one of these `catch`es is needed. Since `NuTo::Exception` inherits from `std::exception`, both would be caught by the lower one and are handled the same way anyway. BTW, why are you catching these exceptions anyway?

Psirus (approved):

src/mechanics/tools/AdaptiveSolve.cpp
TTitscher: Even the documentation of the method says that it only catches `NoConvergence`. So I forgot to update that. Fixed now.

PR #207, opened 2018-02-27
Pde dynamic momentum balance

pmueller2:
Adds a DynamicMomentumBalance integrand derived from the static one that additionally has a density and computes a mass matrix.


applications/integrationtests/mechanics/VibrationalModes1D.cpp
```diff
@@ -0,0 +1,99 @@
...
+    Eigen::VectorXd Solve()
+    {
+        DofInfo dofInfo = DofNumbering::Build(mMesh.NodesTotal(mDof), mDof, mConstraints);
+        Eigen::SparseMatrix<double> cmat = mConstraints.BuildConstraintMatrix(mDof, dofInfo.numIndependentDofs[mDof]);
+        SimpleAssembler asmbl = SimpleAssembler(dofInfo);
+
+        using namespace std::placeholders;
+
+        auto stiffnessF = std::bind(&Integrands::DynamicMomentumBalance<1>::Hessian0, &mPDE, _1, 0.);
+        auto massF = std::bind(&Integrands::DynamicMomentumBalance<1>::Hessian2, &mPDE, _1);
```
Psirus: Why not just use lambdas here?
~~~cpp
auto stiffnessF = [&](const auto& cipd) { return mPDE.Hessian0(cipd, 0.0); };
auto massF = [&](const auto& cipd) { return mPDE.Hessian2(cipd); };
~~~

Psirus (approved):

applications/integrationtests/mechanics/VibrationalModes1D.cpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        AddDofInterpolation(&mMesh, mDof, mMesh.CreateInterpolation(InterpolationTrussLobatto(order)));
+
+        mCellGroup = mCells.AddCells(mMesh.ElementsTotal(), mIntegrationType);
+    }
+
+    Eigen::VectorXd Solve()
+    {
+        DofInfo dofInfo = DofNumbering::Build(mMesh.NodesTotal(mDof), mDof, mConstraints);
+        Eigen::SparseMatrix<double> cmat = mConstraints.BuildConstraintMatrix(mDof, dofInfo.numIndependentDofs[mDof]);
```
TTitscher: cmat is not needed, right?

src/mechanics/integrands/DynamicMomentumBalance.h
```diff
@@ -0,0 +1,34 @@
...
+    DynamicMomentumBalance(DofType dofType, const Laws::MechanicsInterface<TDim>& law, double rho)
+        : MomentumBalance<TDim>(dofType, law)
+        , mRho(rho)
+    {
+    }
+
+    DofMatrix<double> Hessian2(const CellIpData& cellIpData)
+    {
+
+        NMatrix N = cellIpData.N(MomentumBalance<TDim>::mDofType);
```
TTitscher: Just for readability/shortness, I would probably add a line like `auto dof = MomentumBalance<TDim>::mDof`. Or, maybe better, `this->mDof`. No need to change that though.

TTitscher (approved):
Neat example and simple use of a mass matrix, I like.

In my experience, a 1D test for something that has a template `TDim` may not catch all errors. E.g. I sometimes forget a `.transpose()` on some values. If this value is `1x1` in 1D, it will compile and run fine in 1D. Thus, I try to add at least something in 2D or 3D, maybe just a check if `Something<3>::Hessian2` runs without exceptions. But this is just a remark, no need add that in this PR.

src/mechanics/integrands/DynamicMomentumBalance.h
Psirus: Or `using parent = MomentumBalance<TDim>;` + `parent::mDofType`.

PR #208, opened 2018-02-27
Simplify NuTo::Exception

Psirus:
The only useful feature is the (caller, message) syntax. The rest is
just not necessary imho.


TTitscher (approved):
Nice strip down. My knowledge on this virtual destruction thing is a bit sketchy. So as long as `NuTo::Exception` has no  virtual methods, we do not need a virtual destructor, right?
__edit:__ (Even if we derive from it.)

PR #209, opened 2018-02-28
Pde know your dofs

TTitscher:
I had a bug in my custom examples regarding multiple dof types in combination with integrands that do not have all of them. This causes invalid reads during assemby:
> Gradient damage has two dof types, d and eeq. NeumannBc has only d. We
previously try to access NeumannBc::Gradient[eeq]. This is now solved by
asking the gradient what his dof types are and assembe an intersection
of those and the required ones.

So I added `std::vector<DofTypes> DofMatrix/Vector::.DofTypes()` and fixed the assembler. 
Now things got a bit out of control. Some reasoning:
- To test that, I added a proper integration test. With a complex geometry, displacement and force BCs, visualization, adaptive solve, ...
- That required a mesh. (1100 LOC in the diff...)
- And it required voronoi geometries for high order triangles. As I had them available in my custom stuff, I added them as well.
- Vornoi is now based on the integration type and I needed to link `IntegrationTypeBase.cpp` that basically only contained `Info`... --> gone now.
- During a routinely check in valgrind, I realized
   - leaks in visualize and my custom subclasses of LocalIsotropicDamage and GradientDamage (missing virtual dtor)
   - memory access violations in `Legendre` (access `vector[i - 2]` for `i = 1`)


applications/integrationtests/mechanics/GradientDamage.cpp
```diff
@@ -137,11 +141,12 @@ BOOST_AUTO_TEST_CASE(Integrand)
...
 {
-    /**
-     * This just evaluates the gradient and hessian once
-     * to make sure all the dimensions in the matrix calculations
-     * match.
-     */
+    auto binary = boost::unit_test::framework::master_test_suite().argv[0];
+    boost::filesystem::path binaryPath = std::string(binary);
+    binaryPath.remove_filename();
+    std::string meshFile = binaryPath.string() + "/meshes/Holes.msh";
```
Psirus: Why create this `binary` variable if you don't use it? Just do
~~~cpp
boost::filesystem::path binaryPath =
    boost::unit_test::framework::master_test_suite().argv[0];
~~~

applications/integrationtests/mechanics/GradientDamage.cpp
```diff
@@ -157,23 +162,63 @@ BOOST_AUTO_TEST_CASE(Integrand2D)
...
+
+
+    using namespace NuTo::Visualize;
+    PostProcess visu("./GDMout");
+    visu.DefineVisualizer("GDM", cells, VoronoiHandler(VoronoiGeometryTriangle(integration)));
+    visu.Add("GDM", d);
+    visu.Add("GDM", eeq);
+    visu.Add("GDM",
+             [&](const NuTo::CellIpData& data) {
+                 return EigenCompanion::ToEigen(gdm.mDamageLaw.Damage(gdm.Kappa(data)));
```
Psirus: Maybe we should overload the `Add`/`Visualizer::CellData` functions for `double`s, so that you don't need the `ToEigen` as a user.

applications/integrationtests/mechanics/GradientDamage.cpp
```diff
@@ -157,23 +162,63 @@ BOOST_AUTO_TEST_CASE(Integrand2D)
...
+             "Damage");
+
+    auto doStep = [&](double t) { return problem.DoStep(t, "MumpsLU"); };
+    auto postProcess = [&](double t) { return visu.Plot(t, true); };
 
-    BOOST_CHECK_NO_THROW(gdm.Gradient(cipd));
-    BOOST_CHECK_NO_THROW(gdm.Hessian0(cipd));
+    NuTo::AdaptiveSolve adaptive(doStep, postProcess);
+    adaptive.dt = 0.01;
+    adaptive.Solve(0.01); // Only one step for this test. Increase to 1., if you want to see magic happen.
```
Psirus, line 217: Not as magical as I was lead to believe. 5/7.

src/visualize/VoronoiGeometries.cpp
```diff
@@ -147,20 +147,130 @@ VoronoiGeometry NuTo::Visualize::VoronoiGeometryBrick(int num, Spacing s)
...
+ * This will ensure that the voronoi diagramm has an edge on the hull.
+ *
+ * boost::polygon::voronoi now creates _cells_. We only evaluate cells that are associated with the origninal points and
+ * transform them to a NuTo::Visualize::VoronoiGeometry. [AddUnique] will make sure that we add points at the same
+ * coordinate only once.
+ */
+NuTo::Visualize::VoronoiGeometry TransformedVoronoi(const NuTo::IntegrationTypeBase& integrationType,
+                                                    std::vector<Eigen::Vector2d> hullpoints, Eigen::Matrix2d T)
+{
+    T *= 1.e6; // scaling with a big F to deal with the fact that boost::polygon::voronoi handles ints
```
Psirus, line 235: Assuming that our points are always within [-2, 2], you can even multiply by 1e9.

src/visualize/VoronoiGeometries.cpp
```diff
@@ -147,20 +147,130 @@ VoronoiGeometry NuTo::Visualize::VoronoiGeometryBrick(int num, Spacing s)
...
+ *
+ *         x
+ *
+ * This will ensure that the voronoi diagramm has an edge on the hull.
+ *
+ * boost::polygon::voronoi now creates _cells_. We only evaluate cells that are associated with the origninal points and
+ * transform them to a NuTo::Visualize::VoronoiGeometry. [AddUnique] will make sure that we add points at the same
+ * coordinate only once.
+ */
+NuTo::Visualize::VoronoiGeometry TransformedVoronoi(const NuTo::IntegrationTypeBase& integrationType,
```
Psirus, line 232: Personally, I would split that into two functions. One `boost::polygon::voronoi_diagram<double> CreateVoronoiDiagram(...)` with the `TransformedHull` stuff etc, and one `ConvertToVisualizeCells` that takes a boost voronoi diagram and returns a `VoronoiGeometry`.

Psirus (approved):

applications/integrationtests/mechanics/GradientDamage.cpp
TTitscher: Totally right. I mindlessly copied that from somewhere...

applications/integrationtests/mechanics/GradientDamage.cpp
TTitscher: Good point. Do you think it is enough to do this overload in `PostProcess`? I somehow suspect that this is the class people will use most of the time.

src/visualize/VoronoiGeometries.cpp
TTitscher: If someone wants to use this algorithm for quads (which is possible btw but has problems for LOBATTO where points lay _on_ the hull), the reflected points are within [-3, 3]. 
~~~
T *= std::numeric_limits<int>::max() / 3.
~~~
?
Or I'll leave it as a magic number.

src/math/Legendre.cpp
```diff
@@ -7,14 +7,15 @@
...
-    auto Factorial = [=](int i) -> int { return std::tgamma(i + 1); };
+    auto Factorial = [](int i) -> int { return std::tgamma(i + 1); };
 
     std::vector<double> vals(n + 1, 0.);
     if (n >= k)
         vals[k] = Factorial(2 * k) / (std::pow(2, k) * Factorial(k));
     for (int i = k + 1; i <= n; i++)
     {
-        vals[i] = ((2. * i - 1.) * x * vals[i - 1] - (i - 1. + k) * vals[i - 2]) / (i - k);
+        double vals_i_minus_2 = i >= 2 ? vals[i - 2] : 0;
```
pmueller2, line 17: Ok, I probably messed that up when changing the algorithm from recursive to this loop. Sad that this bug didn't  show up in the test.

src/visualize/VoronoiGeometries.cpp
TTitscher: That sounds good. However, `ConvertToVisualizeCells` also involves the back transformation. But I could do the transformation afterwards...

~~~
auto voronoi = CreateVoronoiDiagramm(hull, integration, T);
auto geometry = ConvertToVisualizeCells(voronoi);
TransformBack(&geometry, T);
~~~
?

src/mechanics/cell/SimpleAssembler.cpp
```diff
@@ -22,6 +22,19 @@ void SimpleAssembler::ThrowOnZeroDofNumbering(std::vector<DofType> dofTypes) con
                                     ". Please do so by SimpleAssembler::calling SetDofInfo(...).");
 }
```
pmueller2, line 24: While you are at it, maybe add some documentation?

src/mechanics/dofs/DofMatrixContainer.h
```diff
@@ -80,6 +81,15 @@ class DofMatrixContainer
         return out;
     }
 
+    std::vector<DofType> DofTypes() const
+    {
+        std::vector<DofType> dofTypes;
+        for (const auto& data : mData)
+            if (data.first.first.Id() == data.first.second.Id())
+                dofTypes.push_back(data.first.first);
```
pmueller2: Returns only dofs that occur in pairs? Is this right?

src/visualize/VoronoiGeometries.cpp
```diff
@@ -147,20 +147,130 @@ VoronoiGeometry NuTo::Visualize::VoronoiGeometryBrick(int num, Spacing s)
...
+    {
+        Eigen::Vector2d ip = T * Eigen::Vector2d(integrationType.GetLocalIntegrationPointCoordinates(i).x(),
+                                                 integrationType.GetLocalIntegrationPointCoordinates(i).y());
+        points.push_back({ip.x(), ip.y()});
+    }
+    return points;
+}
+
+std::vector<VoronoiPoint> ReflectPointsOnHull(std::vector<VoronoiPoint> points, std::vector<Eigen::Vector2d> hull)
+{
```
pmueller2, line 190: I think this is a bit over the top - but since it works... It would probably be enough to add 3 points far enough outside the triangle to have the same effect.

pmueller2 (approved):

src/visualize/VoronoiGeometries.cpp
TTitscher: Previously, I calculated an intersection of the voronoi diagram with the hull. I removed that.
Now we modify points (by this reflection) in a way the resulting voronoi diagram _must have_ edges on the hull.

src/mechanics/dofs/DofMatrixContainer.h
TTitscher: Well, maybe I should add  a comment on that on the code. 
~~~
//! @remark std::vector allows duplicates and we do not want duplicates. So we 
//! extract the dofs types from the diagonal entries, where they only appear once.
~~~
Does that make sense?

src/math/Legendre.cpp
TTitscher: I thought about changing the access from `[i]` to `.at(i)` which does bounds checking. Then you would get an `std::out_of_bounds` or something. This is a _tiny_ bit slower, which should not matter at all.

PR #211, opened 2018-03-02
Fix documentation and add warning

### LINKED ISSUE #21: Activate more compiler warnings

Psirus:
- Activate `-Wdocumentation` warning; fixes #21 
- fix the resulting warnings
- remove `TimeControl` class, since its not used and had a lot of warnings :)


TTitscher (approved):
I assumed that `Wdocumentation` forces you to add documentation. But as it only complains about _wrong_ documentation, this is a nice addition!

PR #212, opened 2018-03-02
Reactivate warnings checks on Travis

Psirus:
Copied from last time:

This will add two additional jobs to our Travis build that fail when there are any compiler warnings. These have been missed previously, because one usually doesn't test on multiple compilers on ones own machine.

Note:
- These "Compiler warnings" jobs are allowed to fail; even if there is a warning, you'll still get the green tick.
- They fail on the first warning, even if there are more, to reduce Travis time. The corresponding "Normal build" will have all the warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,13 @@ matrix:
         - compiler: g++
           env: BUILD_TYPE=Release OPENMP=FALSE
         - compiler: g++
-          env: BUILD_TYPE=Release OPENMP=TRUE
-        - compiler: g++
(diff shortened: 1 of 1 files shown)


TTitscher (approved):
Quick idea: The `allow_failures` builds do something like static analysis. So for optimizing the travis times, we could avoid _running_ the tests for those, right?

Psirus:
Yeah, you're right. Currently, that takes about 2 seconds, but you're right, this will probably become relevant in the future.

PR #213, opened 2018-03-02
Separate unit and integration test coverage report

### LINKED ISSUE #64: Travis improvements

Psirus:
Separate reports for unit tests and integration tests. For example, look at [the report of `Cell.h`](https://codecov.io/gh/nutofem/nuto/src/cb6f145f0b869afd43053d3ae0d5bd213379ea1c/src/mechanics/cell/Cell.h), and you can toggle on the upper right which methods got tested by which test suite. Also, the small change in the gradient damage test is to fix a GCC warning. (If you're wondering why that caused a warning, [look here](https://stackoverflow.com/questions/19105350/what-is-the-meaning-of-multi-line-comment-warnings-in-c)). Together with #212, this fixes #64 imho.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,11 +28,9 @@ script:
     # - The source code from travis is 'mounted' as /home/nuto/source via the -v option in the docker run command above
     - docker exec dock cmake -DCMAKE_CXX_COMPILER=$CXX -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DENABLE_OPENMP=$OPENMP -DCMAKE_CXX_FLAGS="$COVERAGE $WERROR" ../source
     - docker exec dock make -j2 unit
+    - if [[ "$COVERAGE" == "--coverage" ]]; then docker exec dock bash -c "cd ../source && scripts/upload_coverage.sh unittests"; fi
     - docker exec dock make -j2 integrationtests
     - docker exec dock ctest -R integration --output-on-failure
+    - if [[ "$COVERAGE" == "--coverage" ]]; then docker exec dock bash -c "cd ../source && scripts/upload_coverage.sh integrationtests"; fi
     - docker exec dock make -j2
     - docker exec dock bash ../source/scripts/check_install.sh
-
-after_success:
-    # will simply fail if --coverage is not set.
-    - docker exec dock bash -c "cd ../source && scripts/upload_coverage.sh"
--- a/applications/integrationtests/mechanics/GradientDamage.cpp
+++ b/applications/integrationtests/mechanics/GradientDamage.cpp
@@ -100,7 +100,7 @@ BOOST_AUTO_TEST_CASE(Integrand)
     //
     //  damage:
     //           _
-    //          / \
+    //          ' '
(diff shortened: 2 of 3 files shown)


applications/integrationtests/mechanics/GradientDamage.cpp
```diff
@@ -100,7 +100,7 @@ BOOST_AUTO_TEST_CASE(Integrand)
     //
     //  damage:
     //           _
-    //          / \
+    //          ' '
```
TTitscher, line 103: Don't you dare touching my... oh wait. Actually better.

.travis.yml
```diff
@@ -24,11 +24,9 @@ script:
     # - The source code from travis is 'mounted' as /home/nuto/source via the -v option in the docker run command above
     - docker exec dock cmake -DCMAKE_CXX_COMPILER=$CXX -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DENABLE_OPENMP=$OPENMP -DCMAKE_CXX_FLAGS=$COVERAGE ../source
     - docker exec dock make -j2 unit
+    - if [[ "$COVERAGE" == "--coverage" ]]; then docker exec dock bash -c "cd ../source && scripts/upload_coverage.sh unittests"; fi
```
TTitscher: Just curious: Is the order of commands important here? So could we, in theory, collect all the coverage stuff at the end? Would there be drawbacks?

TTitscher (approved):

.travis.yml
Psirus: Yes, the order matters. After the unit tests are run, the coverage is uploaded and the `.gcada` files are deleted. Then, the next set of tests is run, and the coverage uploaded again. So unfortunately, I don't see a way to move it to the end.

PR #216, opened 2018-03-06
Pde gradient damage improvements

TTitscher:
It might be annoying to review the GDM stuff. But this is what I am currently working on. And along the way, I find new features/enhancements/bugs where I conveniently use the GDM as a test. And, I think, the gradient damage model is nice to have.


applications/integrationtests/mechanics/StructuralGradientDamage.cpp
```diff
@@ -36,19 +35,19 @@ BOOST_AUTO_TEST_CASE(Integrand)
...
     double fc = 40;
     double gf = 0.021;
-    double c = 1.25;
+    double L = 40;
+    double c = 0.25;
 
     double k0 = ft / E;
     Laws::LinearElastic<1> elasticLaw(E, nu);
-    Constitutive::DamageLawExponential dmg(k0, ft / gf, 1.);
+    Constitutive::DamageLawExponential dmg(k0, ft / gf / 10, 1.);
```
Psirus: I know this 10 is correctly converted to double, but still, this makes me uneasy. `10.0`

applications/integrationtests/mechanics/StructuralGradientDamage.cpp
```diff
@@ -36,19 +35,19 @@ BOOST_AUTO_TEST_CASE(Integrand)
...
     Constitutive::ModifiedMisesStrainNorm<1> strainNorm(nu, fc / ft);
 
-    using Gdm = Integrands::GradientDamage<1, Constitutive::DamageLawExponential>;
+    using Gdm = Integrands::GradientDamage<1, Constitutive::DamageLawExponential, Integrands::DecreasingInteraction>;
```
Psirus: Could that live in a namespace or something? Seems quite specific, while `Integrands::DecreasingInteraction` sounds very generic.

src/mechanics/dofs/DofNumbering.h
```diff
@@ -12,7 +12,12 @@ namespace NuTo
 namespace DofNumbering
 {
 
+//! Builds the dof numbering for type `dof` for all the `dofNodes` and stores it at the `dofNodes`.
+//! @remark Numbering starts at 0 with all the dofs that are not constrained. Constraned dofs have the highest numbers
```
Psirus: typo: "Constraned" vs "Constrained"

src/mechanics/integrands/GradientDamage.h
```diff
@@ -4,17 +4,21 @@
 #include "mechanics/constitutive/LinearElastic.h"
 #include "mechanics/constitutive/ModifiedMisesStrainNorm.h"
 
+#include <iostream>
```
Psirus: Still necessary?

src/mechanics/tools/QuasistaticSolver.cpp
```diff
@@ -118,6 +123,21 @@ GlobalDofVector QuasistaticSolver::ToGlobalDofVector(const Eigen::VectorXd& x)
     return v;
 }
 
+void QuasistaticSolver::WriteTimeDofResidual(std::ostream& out, DofType dofType, std::vector<int> dofNumbers)
+{
+    mGlobalTime -= mTimeStep;
```
Psirus: Maybe you could add a short comment as to why you need to subtract a time step...

src/mechanics/tools/QuasistaticSolver.cpp
```diff
@@ -118,6 +123,21 @@ GlobalDofVector QuasistaticSolver::ToGlobalDofVector(const Eigen::VectorXd& x)
...
+    auto x = ToGlobalDofVector(ToEigen(mX, mDofs));
+    auto residual = mProblem.Gradient(x, {dofType}, mGlobalTime + mTimeStep, mTimeStep);
```
Psirus: ...and then add it back on, every time you use it in the function.

Psirus (approved):

src/visualize/VoronoiGeometries.cpp
```diff
@@ -266,6 +266,10 @@ NuTo::Visualize::VoronoiGeometry TransformedVoronoi(const NuTo::IntegrationTypeB
 
 VoronoiGeometry NuTo::Visualize::VoronoiGeometryTriangle(const NuTo::IntegrationTypeBase& integrationType)
 {
+    for (int i = 0; i < integrationType.GetNumIntegrationPoints(); ++i)
+        if (integrationType.GetLocalIntegrationPointCoordinates(i).sum() > 1.)
+            throw Exception(__PRETTY_FUNCTION__, "Only 2D triangle types are supported");
+
```
pmueller2: Ok, seems here would this shape thing be helpful, since tetrahedra would not throw and Quad, Line, Hexes with only one integration point at the center also not.

src/mechanics/tools/AdaptiveSolve.cpp
```diff
@@ -46,9 +46,9 @@ void AdaptiveSolve::Solve(double tEnd)
             if (t + dt > tEnd)
                 dt = tEnd - t;
         }
-        catch (NewtonRaphson::NoConvergence& e)
+        catch (...)
         {
```
pmueller2: If something goes wrong in the postprocessing - well: decrease the timestep. ?

src/mechanics/dofs/DofNumbering.h
```diff
@@ -12,7 +12,12 @@ namespace NuTo
...
 DofInfo Build(const Group<NodeSimple>& dofNodes, DofType dof, const Constraint::Constraints& constraints);
 
+//! Extracts the dof numbers from `dofNodes` for a given component (0 for scalar, x=0, y=1, z=2 for vector dofs)
+std::vector<int> Get(const Group<NodeSimple>& dofNodes, int component = 0);
+
```
pmueller2, line 21: I noticed that there are these X, Y, Z enums still in use instead of 0,1,2. Should we better stick to numbers?

pmueller2 (approved):

src/mechanics/dofs/DofNumbering.h
Psirus: These are different things imho. One is: "I want to constrain some node in x direction / find a node at x = 10.0"
~~~cpp
auto someNodes = mesh.NodesAtAxis(eDirection::X, 10.0);
Constraint::Component(nodesConstrainedInX, {eDirection::X}, 1.0));
// this would not be as self-explaining:
auto someNodes = mesh.NodesAtAxis(0, 10.0);
~~~
The other is "I want the 2nd component of this vector (regardless of what that represents)"
~~~cpp
auto dofNumbers = Get(someNodes, 2);
~~~

src/mechanics/tools/AdaptiveSolve.cpp
TTitscher: It is a bit tricky here. The solver implementation may throw _some_ exception. Maybe its cause is _"SolverType MUMPS not found"_ and we have to abort. Maybe it is some _"Factorization error"_ where a decreased time step may help. But you are totally right that catching everything is bad. I'll try to think of a solution. Maybe catch outside of NewtonRaphson...

src/mechanics/dofs/DofNumbering.h
TTitscher: Yip, this is the idea. On the call site, I call this method with
 `Get(..., ToComponentIndex(eDirection...))`.

PR #217, opened 2018-03-13
Shapes

Psirus:
- introduction of Shape classes
- factory functions for integration and interpolation types
- higher order integration types for triangles, quads and tetrahedra
- polymorphic handling of cells in `VoronoiHandler` (visualization)


src/mechanics/integrationtypes/IntegrationCompanion.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include "IntegrationType3D6NGauss2x3Ip.h"
+
+#include "IntegrationTypeTensorProduct.h"
+
+using namespace NuTo;
+
+namespace
+{
+
+std::unique_ptr<IntegrationTypeBase> G_Line(int order)
```
TTitscher: Why do you deviate from our naming convention?

src/mechanics/integrationtypes/IntegrationCompanion.cpp
```diff
@@ -0,0 +1,199 @@
...
+    case 11:
+    case 12:
+    case 13:
+    case 14:
+    case 15:
+    case 16:
+    case 17:
+    case 18:
+    case 19:
+    case 20:
```
TTitscher: This looks strange to me. Does ItnegrationTypeTriangle only cover order 7 and up? Why not use it for all?

src/mechanics/integrationtypes/IntegrationCompanion.cpp
```diff
@@ -0,0 +1,199 @@
...
```
TTitscher: Same as for Triangles. It seems like the point of `IntegrationType3D4NGauss1Ip` is also present in `IntegrationTypeTetrahedron`

src/mechanics/integrationtypes/IntegrationTypeTetrahedron.cpp
```diff
@@ -0,0 +1,3108 @@
...
+         Eigen::Vector4d(0.38888888888888901, 0.055555555555555525, 0.055555555555555525, 0.064317124103657003),
+         Eigen::Vector4d(0.055555555555555525, 0.055555555555555525, 0.27777777777777801, 0.064317124103657003),
+         Eigen::Vector4d(0.055555555555555525, 0.16666666666666652, 0.16666666666666652, 0.064317124103657003),
+         Eigen::Vector4d(0.055555555555555525, 0.27777777777777801, 0.055555555555555525, 0.064317124103657003),
+         Eigen::Vector4d(0.16666666666666652, 0.055555555555555525, 0.16666666666666652, 0.064317124103657003),
+         Eigen::Vector4d(0.16666666666666652, 0.16666666666666652, 0.055555555555555525, 0.064317124103657003),
+         Eigen::Vector4d(0.27777777777777801, 0.055555555555555525, 0.055555555555555525, 0.064317124103657003),
+         Eigen::Vector4d(0.055555555555555525, 0.055555555555555525, 0.16666666666666652, 0.064317124103657003),
+         Eigen::Vector4d(0.055555555555555525, 0.16666666666666652, 0.055555555555555525, 0.064317124103657003),
+         Eigen::Vector4d(0.16666666666666652, 0.055555555555555525, 0.055555555555555525
```
TTitscher: -0.7138836084956467, right? :astonished: 

I highly suspect that you neither calculated them by hand nor hacked them in from a book. Is there a formula we might use? Maybe calculate once upon construction and save for later?
Or at least provide the source?

src/mechanics/interpolation/InterpolationBrickLinear.h
```diff
@@ -31,5 +32,13 @@ class InterpolationBrickLinear : public InterpolationSimple
...
         return 8;
     }
+
+    const Shape& GetShape() const override
+    {
+        return mShape;
+    }
+
+private:
+    Hexahedron mShape;
```
TTitscher, line 42: This looks somehow problematic. We have to save that shape here to return it by reference for the polymorphism. Well. But it seems that there are no two different instantiations of a shape, so it is a static object. Maybe add `static`? Or change `GetShape` to return an enum?

src/mechanics/interpolation/InterpolationBrickLinear.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 #include "mechanics/interpolation/InterpolationSimple.h"
 #include "mechanics/elements/ElementShapeFunctions.h"
+#include "math/shapes/Hexahedron.h"
```
TTitscher, line 4: As the shape is called `Hexahedron`. Should we rename the `InterpolationBrick` to `InterpolationHexahedron`?

src/mechanics/interpolation/InterpolationCompanion.cpp
```diff
@@ -0,0 +1,171 @@
...
+    case 1:
+        return std::make_unique<InterpolationPyramidLinear>();
+    default:
+        throw Exception(__PRETTY_FUNCTION__,
+                        "Prism interpolation of order " + std::to_string(order) + " is not defined.");
+    }
+}
+
+
+std::unique_ptr<InterpolationSimple> NuTo::CreateSerendipityInterpolation(const Shape& shape, int order)
```
TTitscher: Are you sure about the naming `Serendipity`? It confuses me a bit. Can triangles be serendipty? How would I name a quadratic quad with 9 nodes? Well, by accident it matches Lobatto2. What about 16 nodes?

src/mechanics/interpolation/InterpolationQuadQuadratic.h
```diff
@@ -1,15 +1,16 @@
 #pragma once
 #include "mechanics/interpolation/InterpolationSimple.h"
 #include "mechanics/elements/ElementShapeFunctions.h"
+#include "math/shapes/Quadrilateral.h"
 
 namespace NuTo
 {
-class InterpolationQuadSerendipity : public InterpolationSimple
+class InterpolationQuadQuadratic : public InterpolationSimple
```
TTitscher, line 8: I like the idea of passing the order. So following other classes and your naming, we could have a `InterpolationQuadSerindipity(int order)` here, right? Maybe in another PR, as this is huuuuuge.

src/mechanics/interpolation/InterpolationTetrahedronLinear.h
```diff
@@ -32,5 +33,13 @@ class InterpolationTetrahedronLinear : public InterpolationSimple
...
         return 4;
     }
+
+    const Shape& GetShape() const override
+    {
+        return mShape;
+    }
+
+private:
+    Triangle mShape;
```
TTitscher: Tetrahedron is a triangle?

src/visualize/VoronoiHandler.cpp
```diff
@@ -2,27 +2,60 @@
...
 
-VoronoiHandler::VoronoiHandler(VoronoiGeometry geometry)
-    : mGeometry(geometry)
+std::unique_ptr<HandlerInterface> VoronoiHandler::Clone() const
 {
+    return std::make_unique<VoronoiHandler>(*this);
 }
 
-std::unique_ptr<HandlerInterface> VoronoiHandler::Clone() const
+int GetNumSubCells(const CellInterface& cell)
```
TTitscher: - This relies on a hidden dependency of the integration type numbering in the integration types and the `VoronoiGeometry` methods. And there is no way to check this assertion.
- This will not work for triangles.

I think that the integration type is required for Voronoi visualization. So we could add it as a member at the cell. Meh. Or we pass it to the geometry.

src/visualize/VoronoiHandler.h
```diff
@@ -25,8 +27,7 @@ class VoronoiHandler : public HandlerInterface
 {
 public:
     //! Constructor.
-    //! @param geometry definition of arbitrary voronoi cells
-    VoronoiHandler(VoronoiGeometry geometry);
+    VoronoiHandler(){};
```
TTitscher: Would adding the ctor that takes a `VornoiGeometry` solve the problem for the triangles?

test/mechanics/integrationtypes/CMakeLists.txt
```diff
@@ -2,3 +2,60 @@ add_unit_test(IntegrationTypeTensorProduct
     math/Quadrature.cpp
     math/Legendre.cpp
     )
+    
+add_unit_test(IntegrationCompanion
```
TTitscher: I cannot think of a way to unit test a factory. The fact alone, that you requre over 10 other classes indicates IMO that these should be an integration test.

test/mechanics/integrationtypes/IntegrationTypesTetrahedra.cpp
```diff
@@ -0,0 +1,326 @@
...
+        double y = f(intType.GetLocalIntegrationPointCoordinates(i));
+        double w = intType.GetIntegrationPointWeight(i);
+        result += w * y;
+    }
+    return (result);
+}
+
+double ExactIntegralMonomialTetrahedron(int order, int indx)
+{
+    std::vector<std::vector<double>> expectedResult = {
```
TTitscher: There is really no way for me to make sure that this test is correct. Where do those numbers come from?

TTitscher (changes requested):
If I accept now, there is no way for me to visualize my precious triangles. So I cannot do that. The other points may be debatable. But there is sooo much going on:
1) Shapes. Nice!
2) New integration types. Somehow nice but with thousands of magic numbers (even in the test). Where do they come from? 
3) visualization. It seems like you were looking for an example to "just visualize cells". But IMO you picked the wrong handler. I mean it makes sense for the average visualizer to only depend on the shape. For voronoi, we perform a voronoi decomposition of the integration point coordinates. Conceptually, we do need further infromation, e.g. the integration type, for that.

Minor stuff:
 - Renaming/Naming of serendipity. I do not get that, but maybe it is just me and my wrong understanding of the word.
- We still have ElementShapeFunctions.cpp :P

test/math/Shapes.cpp
```diff
@@ -0,0 +1,99 @@
...
+    Eigen::Vector3d pIn(0.1, 0.1, 0.1);
+    Eigen::Vector3d pOut(0.1, 0.1, -0.1);
+    Eigen::Vector3d pBoundary(0.5, 0.5, 0.5);
+
+    BOOST_CHECK(shape.IsWithinShape(pIn));
+    BOOST_CHECK(!shape.IsWithinShape(pOut));
+    BOOST_CHECK(!shape.IsWithinShape(pBoundary));
+}
+
+BOOST_AUTO_TEST_CASE(HexahedronIsInside)
```
TTitscher: Should -0.1, -0.1, -0.1 be inside? I think it should. Can you add this test? :)

test/mechanics/integrationtypes/IntegrationTypeTriangle.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+}
+
+BOOST_AUTO_TEST_CASE(IntegrateTriangle)
+{
+    int maxOrder = 19;
+    for (int i = 1; i <= maxOrder; i++)
+    {
+        CheckTriangleIntegration(i, i);
+        std::cout << "Done: order " << i << std::endl;
```
TTitscher: I would prefer a [`BOOST_TEST_MESSAGE`](http://www.boost.org/doc/libs/1_46_1/libs/test/doc/html/utf/user-guide/test-output/BOOST_TEST_MESSAGE.html) here. Would get rid of the `iostream` include.

test/visualize/VoronoiGeometries.cpp
```diff
@@ -110,7 +110,9 @@ BOOST_AUTO_TEST_CASE(VoronoiTriangle)
      *  Contains 6 cells and 13 points.
      *
      */
-    auto geometry = Visualize::VoronoiGeometryTriangle(triangleType6Ip);
+    auto geometry = Visualize::VoronoiGeometryTriangle(triangleTypeFourthOrder);
+    std::cout << geometry.voronoiCells.size() << std::endl;
```
TTitscher: Looks like a debug command.

TTitscher (approved):

PR #218, opened 2018-03-15
Remove Metamodel/Optimize/Python

### LINKED ISSUE #200: Remove/Reorganize metamodel & optimize
We have some chunks of dead and untested code in metamodel and optimize. And we should do something about that. (Most of them seem to be helper classes to `NeuralNetwork`.) 
My proposal:

feature   |   action
-----|-----
Optimize::NewtonRaphon | already replaced by `Math::NewtonRaphson::Solve()`
Optimize::ConjugateGradient | Use an external solver lib, e.g. [`Eigen::ConjugateGradient<>`](https://eigen.tuxfamily.org/dox/classEigen_1_1ConjugateGradient.html)
Optimize::MisesWielandt | ??
Metamodel::PolynomialLeastSquareFitting | move to `/math/` (and simplify a bit)
Metamodel::NeuralNetwork   |  Use an external lib. E.g. NuTo::Fem + [TensorFlow](https://www.tensorflow.org/), both provide a python interface. (Well, not yet in NuTo, but soooon.)

If there are no objections, I'd like to tackle this soon, say, right after the next NuTo Meeting.

Psirus:
As just discussed. Fixes #200.


src/metamodel/PolynomialLeastSquaresFitting.h
```diff
@@ -1,51 +0,0 @@
-#pragma once
-
-#include "metamodel/Metamodel.h"
-
-#include <vector>
-
-namespace NuTo
-{
-class PolynomialLeastSquaresFitting : public virtual NuTo::Metamodel
```
TTitscher: The plan was to keep PolynomialLeastSquareFitting and move it to math, right?

TTitscher (changes requested):
As I understood, we wanted to keep PolyLeastSquareFitting and move it to `math`. 
Everything else is nice, >80% coverage. :clinking_glasses:

TTitscher (approved):

PR #220, opened 2018-03-15
Pde prism create

TTitscher:
The goal is to take the triangle between two tetrahedron elements and _blow it up_ to 3d prism. For concrete geometry, this can be used to introduce an interfacial transition zone between matrix and aggregates.

I planned to rewrite `OldNuTo::ElementPrismCreate()`, which is a quite buggy, long and incomprehensive function. After receiving a tremendously helpful code snippet of @pmueller2 (thanks!!), I decided to do that directly in gmsh. As the gmsh code is quite complex and cannot make use of the openCascade features of gmsh3, I added the `GmshWriter` to wrap everything.

The tests consist of running gmsh on the generated `.geo` files and read them with `MeshGmsh` without errors. Do you think that is enough?


applications/integrationtests/geometryConcrete/CMakeLists.txt
```diff
@@ -18,27 +18,17 @@ if(ENABLE_PYTHON)
...
-        Math Base GeometryConcrete Boost::unit_test_framework)
-    target_include_directories(GeometryConcreteAPI
-        PRIVATE ${CMAKE_SOURCE_DIR}/test/tools)
-    add_test(integration::geometryConcrete::GeometryConcreteAPI.cpp
-        GeometryConcreteAPI ${GMSH_PATH})
-    append_to_tests(GeometryConcreteAPI)
+    add_integrationtest(GeometryConcreteAPI)
+    target_link_libraries(GeometryConcreteAPI GeometryConcrete)
+
+    add_executable(GmshWriter EXCLUDE_FROM_ALL GmshWriter.cpp)
```
Psirus: You don't need the `EXCLUDE_FROM_ALL` here, because everything in `integrationtests/` is excluded. Also, have you considered adding the additional option you pass to the test to the `add_integrationtest` macro, which you could then reuse?

applications/integrationtests/geometryConcrete/GmshWriter.cpp
```diff
@@ -0,0 +1,83 @@
...
+    BOOST_CHECK(not meshGmsh.GetPhysicalGroup(opts.physicalGroupMatrix).Empty());
+    BOOST_CHECK(not meshGmsh.GetPhysicalGroup(opts.physicalGroupAggregates).Empty());
+
+    if (opts.interfaceThickness != 0)
+        BOOST_CHECK(not meshGmsh.GetPhysicalGroup(opts.physicalGroupInterfaces).Empty());
+}
+
+std::string GmshPath()
+{
+    return boost::unit_test::framework::master_test_suite().argv[1];
```
Psirus, line 22: Why are you passing this gmsh path around? Can we not assume that it lives in `$PATH`?

applications/integrationtests/geometryConcrete/GmshWriter.cpp
```diff
@@ -0,0 +1,83 @@
...
+    CheckMesh(filename + ".msh", opts);
+}
+
+BOOST_AUTO_TEST_CASE(Gmsh2D)
+{
+    Eigen::MatrixXd m(2, 3);
+    m.row(0) = Eigen::Vector3d(5, 5, 3);
+    m.row(1) = Eigen::Vector3d(15, 15, 3);
+
+    GmshWriter::Options opts(2, 2);
```
Psirus, line 48: One of those cases where *strong types* or maybe the [named parameter idiom](https://isocpp.org/wiki/faq/ctors#named-parameter-idiom) would be nice. But I guess its overkill for such a little-used class.

applications/integrationtests/geometryConcrete/GmshWriter.cpp
```diff
@@ -0,0 +1,83 @@
...
+    Check(FileName("2DNormal"), GmshWriter::Box2D{Eigen::Vector2d(0, 0), Eigen::Vector2d(20, 20)}, m, opts);
+
+
+    opts.interfaceThickness = 0.5;
+    opts.additionalOptions = "Mesh.SecondOrderIncomplete = 1;\n";
```
Psirus, line 53: If you wanted to get really fancy, you could automatically insert a `\n` after every semicolon.

src/geometryConcrete/GmshWriter.cpp
```diff
@@ -0,0 +1,388 @@
+#include "geometryConcrete/GmshWriter.h"
+#include <fstream>
+
+const std::string sAggregate = R"(
```
Psirus, line 4: These strings are much nicer than the old version :+1:

Psirus (approved):

applications/integrationtests/geometryConcrete/GmshWriter.cpp
TTitscher: Not sure. I use this cmake [`find_proram`](https://cmake.org/cmake/help/v3.0/command/find_program.html) to find the executable. By setting `CMAKE_FIND_ROOT_PATH`, you can specify additional search paths. This seemed more like the "cmake" way.

applications/integrationtests/geometryConcrete/CMakeLists.txt
TTitscher: good point!

scripts/cmake/AddIntegrationtest.cmake
```diff
@@ -4,18 +4,11 @@ function(append_to_tests TestName)
...
-        endforeach()
-    endif()
-
     add_executable(${test} ${test}.cpp)
     target_link_libraries(${test} Mechanics Math Base Visualize Metamodel
         Boost::unit_test_framework ${LAPACK_LIBRARIES})
     target_include_directories(${test} PRIVATE ${CMAKE_SOURCE_DIR}/test/tools)
     get_filename_component(module ${CMAKE_CURRENT_SOURCE_DIR} NAME)
-    add_test(integration::${module}::${test}.cpp ${test} ${AdditionalArgs})
+    add_test(integration::${module}::${test}.cpp ${test} ${ARGN})
```
TTitscher: Lesson learned. :)

PR #221, opened 2018-03-16
Create only one NuTo lib

Psirus:
What used to be Base, Math, GeometryConcrete, Mechanics and Visualize are merged into one. This makes the cmake code a bit simpler and also allows linking simply using `-lnuto`. 

Currently, a lot of the visualize unit tests link to NuTo. So either we change these tests, or move them to integrationtests, or ...?


TTitscher (approved):
As for the visualize topic. Many tests consist of actually writing files and reading them. So for now, I think they would be better suited for integration tests. 

On the other hand, it seems like the _easy way out_. Calling it integration test, just because it is hard to think of proper unit tests... One way would be to test the `Visualizer` by mocking a `HandlerInterface`. Maybe overkill though.

Well, you decide. :stuck_out_tongue:

pmueller2 (approved):

PR #222, opened 2018-03-16
Rename src/ to nuto/

Psirus:
As discussed in the meeting, this will rename `src/` to `nuto/`, and the includes change from `#include "base/Whatever.h"` to `#include "nuto/base/Whatever.h"`.


scripts/cmake/CreateNutoModule.cmake
```diff
@@ -4,12 +4,11 @@ function(create_nuto_module ModuleName ModuleSources)
...
     target_include_directories(${ModuleName} PUBLIC
-        $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/src>
-        $<INSTALL_INTERFACE:include/nuto>
+        $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>
+        $<INSTALL_INTERFACE:include/>
         )
 
     install(TARGETS ${ModuleName} EXPORT NuToTargets
         LIBRARY DESTINATION lib
-        INCLUDES DESTINATION include/nuto
```
Psirus, line 13: This is the one additional deletion. Now, `#include "base/Exception.h"` is no longer be possible in projects using NuTo.

TTitscher (approved):
So now we just added `nuto` in front of every include. This works fine, I guess. I also saw that `InterpolationCompanion` includes via relative path.
What do you suggest as the way to go from here. 
- Are we happy with `#include "nuto/math/stuff/Stuff.h"`?
- Do we want to go for `#include "Stuff.h"` or even `#include ../stuff2/OtherStuff.h"`? (I personally prefer this..) How to we get there?
    - Change it for every new code?
    - Change it whenever we come accross?
    - A _fix includes_ sprint?

Psirus:
For same directory, I'd just use the `#include "Stuff.h"`. For everything else, I think `#include "nuto/math/stuff/Stuff.h"` is fine. Personally, I don't like `#include ../stuff2/OtherStuff.h"`, but in the end, not important.

As for when to change: when you're writing new code or touching it anyway and it bothers you. No need to go over everything to "fix" it, imho.

joergfunger (approved):
I would also opt for not using the relative path #include "../my/math", and I would do it on a "whenever I touch a file, I'll change it" rule

PR #234, opened 2018-03-27
Pde gradient damage calibration

TTitscher:
In local damage models, the fracture energy parameter of the damage laws can be directly related to the global fracture energy. Damage localizes in one (layer of) elements and you can analytically integrate that. Due to the nonlocality of the gradient damage model, damage does _not_ localize in a single element, but in many, to different degrees. This makes an analytical integration hard/impossible. With additional features like a damage dependent nonlocal parameter, it gets even harder.

These commits adds an example that provides this calibration. Using the secant methods (we cannot calculate derivatives here), the _local_ gf parameter is calibrated to match a given _global_ Gf parameter in a uniaxial tensile test.

__Edit:__ ... and the _global_ Gf is what you measure in experiments.

### DIFF
--- a/applications/examples/GradientDamageCalibration.cpp
+++ b/applications/examples/GradientDamageCalibration.cpp
@@ -115,32 +115,39 @@ double GlobalFractureEnergy(TGdm& gdm, double L = 50, int nElements = 200, doubl
     return GfTotal - lPreDamage * preDamageIntegral;
 }
 
+double FindRootWithoutDerivative(std::function<double(double)> f, double guess, int significantBits = 10,
+                                 long unsigned maxIter = 20, double factor = 2)
+{
+    std::pair<double, double> result = boost::math::tools::bracket_and_solve_root(
+            f, guess, factor, true, boost::math::tools::eps_tolerance<double>(significantBits), maxIter);
+    // the root is somewhere between result.first and result.second
+    return 0.5 * (result.first + result.second);
+}
+
 int main()
 {
     const double GlobalFractureEnergyParameter = 0.1;
 
     auto f = [&](double gf) {
-        std::cout << "Calculating for gf = " << gf << " ... ";
+        std::cout << "Calculating for local gf = " << gf << " ... ";
         std::cout << std::flush;
         DofType d("Displacements", 1);
         ScalarDofType eeq("NonlocalEquivalentStrains");
 
         Laws::LinearElasticDamage<1> elasticLaw(E, nu);
-        Constitutive::DamageLawExponential dmg(k0, ft / gf, 0.99999);
+        Constitutive::DamageLawExponential dmg(k0, ft / gf, 0.999999);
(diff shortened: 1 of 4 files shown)


applications/examples/GradientDamageCalibration.cpp
```diff
@@ -0,0 +1,176 @@
...
+    adaptiveSolve.dtMin = 1.e-10;
+    adaptiveSolve.dtMax = 0.01;
+    adaptiveSolve.Solve(1.);
+
+    loadDisplacement.close();
+    Eigen::VectorXd loads = EigenIO::ReadFromFile(loadDispFileName).col(2);
+    Eigen::VectorXd disps = EigenIO::ReadFromFile(loadDispFileName).col(1);
+
+    Tools::GlobalFractureEnergyIntegrator gfIntegrator(loads, disps);
+    double GfTotal = gfIntegrator.IntegrateSofteningCurve(1., 0.1);
```
Psirus: 1. What is this »"crack" area«? The cross-section of the specimen?
2. I assume `rForceThreshold` is in "force units". Would it make sense to supply that in "percent of maximum force"?

applications/examples/GradientDamageCalibration.cpp
```diff
@@ -0,0 +1,176 @@
...
+    return GfTotal - lPreDamage * preDamageIntegral;
+}
+
+//! Finds the root of @fFunction by using the secant method (no derivatives required)
+//! @param fFunction function to find the root for
+//! @param x0 first guess
+//! @param x1 second guess
+//! @param tolerance algorithm stopts if abs(fFunction) < tolerance
+//! @param maxIterations algorithm stops after this number of steps
+double SecantMethod(std::function<double(double)> fFunction, double x0, double x1, double tolerance = 1.e-4,
```
Psirus: This function seems to be useful beyond this example. Do we want to move it to `math/`?

applications/examples/GradientDamageCalibration.cpp
```diff
@@ -0,0 +1,176 @@
...
+{
+    const double GlobalFractureEnergyParameter = 0.1;
+
+    auto f = [&](double gf) {
+
+        DofType d("Displacements", 1);
+        ScalarDofType eeq("NonlocalEquivalentStrains");
+
+        Laws::LinearElasticDamage<1> elasticLaw(E, nu);
+        Constitutive::DamageLawExponential dmg(k0, ft / gf, 0.9999999);
```
Psirus: You forgot a nine :ghost:

Psirus (approved):

applications/examples/GradientDamageCalibration.cpp
TTitscher: You want to apply a big enough boundary condition so that the forces really go down to zero. But if they are _actually_ zero, the matrices become singular. But you are right, one more 9 just to be sure. :stuck_out_tongue:

Issue #232, opened 2018-03-26
Damage material parameters


TTitscher:
### Problem

Currently, a lot of code is required to instantiate the appropriate [laws](https://github.com/nutofem/nuto/blob/PDE_reviewed/test/mechanics/constitutive/LocalIsotropicDamage.cpp#L38)/[integrands](https://github.com/nutofem/nuto/blob/PDE_reviewed/test/mechanics/integrands/GradientDamage.cpp#L114).  Like 

~~~cpp
template <int TDim>
auto TestLaw()
{
    // Define the law using policy based design principles, hopefully applied correctly
    using Damage = Constitutive::DamageLawExponential;
    using StrainNorm = Constitutive::ModifiedMisesStrainNorm<TDim>;
    using Evolution = Laws::EvolutionImplicit<TDim>;
    using Law = Laws::LocalIsotropicDamage<TDim, Damage, Evolution>;

    Damage dmg(kappa0, beta, alpha);
    Evolution evolutionEq(StrainNorm(nu, fc / ft), /*numCells=*/1, /*numIps=*/1);
    Laws::LinearElasticDamage<TDim> elasticDamage(E, nu);

    return Law(elasticDamage, dmg, evolutionEq);
}
~~~

Some material parameters are required multiple times (`kappa0 = ft/E`) which is error prone and long. Also parameters like alpha or beta are not intuitive. But they do relate to real material parameters.

### Solution

1) Define a material class. 
Dealing with mechanical softening problems often involves the same set of parameters. Like
~~~cpp
struct SofteningMaterial
{
  double youngsModulus;
  double poissonRatio;
  double tensileStrength;
  double compressiveStrength;
  double residualStrength;
  double fractureEnergy;
  double nonlocalRadius;
}
~~~
2) Add constructors that takes a `SofteningMaterial` to the softening laws/integrands, as well as the deeper classes like (strain norms, damage laws, linear elastic damage). This allows a proper policy based design with defaulted template arguments. The code above would then look like
~~~cpp
template <int TDim>
auto TestLaw(SofteningMaterial material)
{
    return Laws::LocalIsotropicDamage<TDim>(material); // with reasonable the default template arguments
}
~~~

3) Still allow construction with the individual parameters. Still allow customization by changing the template parameters. 

### Discussion:

- Is that relevant for someone else? I think so as we often deal with concrete. 
- Naming alright? Access via public members alright? With `m` (`mYoungsModulus` vs `youngsModulus`)?

PR #235, opened 2018-03-28
Pde softening material

TTitscher:
Basically solves #232.

- Introduce a `Material` namespace, in this case containing `Material::Softening` with all necessary parameters. A `Material::DefaultConcrete()` is provided and used in the tests.
- Main code addition is adding ctors that take this material instead of the single parameters
- Allows defaulting most of the template parameters of `LocalIsotropicDamage` and `GradientDamage` and simplifies the ctors
- At least one damage law parameter is related to the fracture energy. This obviously differs for linear softening and exponential softening. Test are added to ensure correctness here.

### DIFF
--- a/applications/integrationtests/mechanics/StructuralGradientDamage.cpp
+++ b/applications/integrationtests/mechanics/StructuralGradientDamage.cpp
@@ -7,7 +7,6 @@
 
 #include "nuto/mechanics/integrands/GradientDamage.h"
 #include "nuto/mechanics/integrands/NeumannBc.h"
-#include "nuto/mechanics/constitutive/damageLaws/DamageLawExponential.h"
 #include "nuto/mechanics/mesh/UnitMeshFem.h"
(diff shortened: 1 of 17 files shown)


applications/integrationtests/mechanics/StructuralGradientDamage.cpp
```diff
@@ -29,22 +28,14 @@ BOOST_AUTO_TEST_CASE(Integrand)
     DofType d("Displacements", 1);
     ScalarDofType eeq("NonlocalEquivalentStrains");
 
-    /* MATERIAL */
-    double E = 30000;
-    double nu = 0.2;
-    double ft = 4;
-    double fc = 40;
-    double gf = 0.021 * 10; // we have to adjust the material parameters due to nonlocal decreasing interaction
```
Psirus, line 37: That was probably a useful comment for someone comparing the 1D and 2D test.

Psirus (approved):

Issue #194, opened 2018-02-14
Decoupling of assembler and cells


Psirus:
What's the problem?
-------------------

Currently, the assemble has functions like

~~~cpp
GlobalDofVector BuildVector(const Group<CellInterface>& cells, CellInterface::VectorFunction f) const;
~~~

and in there does things like calling `cell.Integrate(f)`. These things (cells, functions-to-be-integrated) are not really part of assembly. Ideally it would just get a vector (matrix) and the numbering and insert them it the appropriate place into the global matrix.

Problems:
- Does more than it should: assembly + call `cell.Integrate()`
- Unnecessary dependency on `Group` or `CellInterface`
- Can't assemble matrices that don't originate from cells
- Modifying the local matrix before assembly leads to duplicate code (`BuildDiagonallyLumpedMatrix`)

Proposal
--------

Instead of writing the cells and functions directly at the assembler, the assembler would get a *generator* that will return vector (matrix) + numbering when asked. The assembler does not, and should not, care how these are generated. Working with functions and cells could be a derived class / implementation of a generator base / concept. If additional transformations are to be applied to a local output (mass lumping, condensation), these can be performed in the generator.

Code
----

The return type of such a generator will be assumed to be `using Entry = std::pair<Eigen::VectorXd, Eigen::VectorXi>;`. A simple implementation could be

~~~cpp
class EntryGenerator
{
public:
    virtual Entry next() = 0;
    virtual bool isValid() = 0;
};
~~~

which would allow the assembler to do something like this

~~~cpp
while(generator.isValid())
{
    Entry entry = generator.next();
    globalVec[entry.second] = entry.first;
}
~~~

This is nice because it is dead simple. Another approach, for fans of syntactic sugar and range-based for loops:

~~~cpp
class EntryGenerator
{
public:
    virtual EntryIterator begin() = 0;
    virtual EntryIterator end() = 0;
    // ...
};
~~~

could result in nice syntax like this

~~~cpp
for (Entry entry : generator)
{
    globalVec[entry.second] = entry.first;
}
~~~

However, this results in a bit of code for the iterators. A complete example can be seen [here](https://gist.github.com/Psirus/6474433fe0a08f969d624b937f5f4fc4).

Questions
---------

- Do you like the idea?
- What would an `Entry` look like?
- Dead simple generator vs iterators?
- Dynamic vs static polymorphism of generators?
- `<insert-your-concern-here>`

pmueller2:
[...]

Issue #164, opened 2018-01-17
PDE remove J K JJ JK ...


TTitscher:
Disclaimer: This is the idea of @Psirus, so I cannot be sure to accurately represent his views. But he was ~~~too lazy~~~ not motivated enough to write it himself.

## Problem

`GlobalDofVector` and `GlobalDofSparseMatrix` are the main global types returned from the assembler. They are designed to meet the requirements for our [linear constraints](http://nuto.readthedocs.io/en/master/Constraints.html) that eliminate dofs from the global system. They do that by splitting the types in _independent_(`J`) and _dependent_(`K`) parts. For other types of constraints (penalty, lagrange multiplier, ...) this is not needed. 

## Solution

We make the subdivision into `J/K` not the default case, but a special case. The assembler returns the existing `DofVector` and `DofSparseMatrix` instead of `GlobalDofVector` and `GlobalSparseMatrix`. The latter ones are removed.
Possible use cases:
- penalty method:
    - chose an arbitrary dof numbering
    - assemble the system
    - apply penalty terms to some of those entries
- linear constraints:
    - chose a dof numbering that sorts dependent dofs to the end
    - assemble the system
    - do the calculations including `CMat` on `BlockViews` of the matrix, e.g. [`topLeftCorner`](https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#a5b8664897a39a39fbb8e98d70b94f16d) will correspond to `GlobalSparseMatrix.JJ`

In NuTo_Master those details are hidden in a `ApplyCMatrix` method. This method can be rewritten to work on those views instead of submatrices `JJ, KJ,...` and will be hidden.

## Discussion

Upon external request, I [benchmarked](https://gist.github.com/TTitscher/c122fb60b12c2ba0fa4e5e488a766003#file-applycmat-cpp) it. 

~~~
----------------------------------------------------------
Benchmark                   Time           CPU Iterations
----------------------------------------------------------
ApplyCMatFull       103761570 ns  103737108 ns          6
ApplyCMatSeparated  100617959 ns  100616271 ns          6
~~~
(For 2M dofs, 100 dependent, sparsity 0.0001. `Separated` means with JJ, JK, ...)

Even though it is a bit slower, IMO a solver call will dominate the run time by far.

TTitscher:
I just figured out that this may not be as trivial as I thought - still doable however. If we regard multiple dof types `D` and `T`, we currently have:

~~~
[...]

PR #236, opened 2018-03-29
Pde assembler

TTitscher:
- Step towards #194
- Assembles directly into `DofVector/DofMatrixSparse` so somehow requires #164 to be 'wired' to the rest of NuTo. It may help _Team JK_ (@Psirus @joergfunger ) though.
- Advantage over `SimpleAssembler`: Option to `SetZero()` the matrix and keep the nonzero entries for the next assembly.
- remark: It has not `AddLumped` method. IMO this should somehow be done in a `LumpedCellAssembler` or something that transforms the cell-vector into a matrix and calls `MatrixAssembler::Add(...)`. Or is this too inefficient?

### DIFF
--- a/nuto/mechanics/cell/Assembler.h
+++ b/nuto/mechanics/cell/Assembler.h
@@ -1,5 +1,4 @@
 #pragma once
-#include "nuto/mechanics/dofs/DofInfo.h"
 #include "nuto/mechanics/dofs/DofContainer.h"
 #include "nuto/mechanics/dofs/DofVector.h"
 #include "nuto/mechanics/dofs/DofMatrix.h"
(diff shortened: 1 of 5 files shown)


nuto/mechanics/cell/Assembler.h
```diff
@@ -0,0 +1,87 @@
...
+    //! resizes the internal DofVector
+    //! @param sizes number of global dofs for each dof type
+    void Resize(DofContainer<int> sizes);
+
+    //! adds an arbitrary contribution `v` to the internal DofVector
+    //! @param v contribution, e.g. a local element/cell vector
+    //! @param numbering mapping from local (0...numbering.size()) to global (0...sizes)
+    //! @param dofTypes specific dof types to assemble - an empty value will assemble all available dof types from the
+    //! numbering
+    void Add(const DofVector<double>& v, const DofVector<int>& numbering, std::vector<DofType> dofTypes = {});
```
joergfunger, line 55: Why do we need the dofTypes? I would assume that the DofVector should only include the dof types to be numbered

joergfunger (commented):
The general idea is very nice. However, I'm not sure, if we need multiple assemblers (for vector/matrix). Should the assembler just provide functions that add one entity, e.g. an element matrix (input) to another entity (e.g. a triplet list in case of eigen). This last triplet list could be be wrapped to include the Get, Finish functionality. 

This would mean one Assembler that provides functions to add to a vector and to add to a matrix. The numbering should IMO be something that is stored at the assembler as a private variable.

TTitscher:
I will add a static `Add` method for the `VectorAssembler` as well. I can also collect the static methods in a common namespace like
~~~cpp
void Assembler::Add(DofVector& rVector, DofVector v, DofVector numbering);
void Assembler::Add(DofMatrixSparse& rMatrix, DofMatrix m, DofVector numbering);
void Assembler::Add(DofContainer<TripletList>& rTriplets, DofMatrix m, DofVector numbering);
~~~

I still like to keep the classes as they avoid reallocations without changing the syntax like:
~~~cpp
MatrixAssembler assembler(...);
for (double t : timeSteps)
{
   const auto& K = BuildMatrix(assembler, cells.begin(), cells.end(), HessianFunction);
   // in the first call of this loop, a K will be build from a triplet list
   // the second time, the nonzeros are automatically reused and
   // assembly is done with `coeffRef`.
}
~~~

nuto/mechanics/cell/Assembler.h
TTitscher: I wouldn't have made this assumption. But it seems like a nice idea and we can certainly do that!

joergfunger:
Can we assure that the sparsity pattern between different time steps does not change? In particular, we do not assemble  entries that are almost zero. In case of e.g. the gradient damage model that stress is only a function of the nonlocal strains as soon as we exceed the damage initiation threshold. Thus, this part of the matrix is initially zero and thus cannot be accessed via coeffRef.
Furthermore, what is the advantage of using this direct access using coeffRef?

TTitscher:
> In particular, we do not assemble entries that are almost zero.

Well, we probably should do that. Currently, we don't. In anyway, `coeffRef` will insert (possibly inefficient) the entry anyways. https://eigen.tuxfamily.org/dox/classEigen_1_1SparseMapBase_3_01Derived_00_01WriteAccessors_01_4.html#title6

> Furthermore, what is the advantage of using this direct access using coeffRef?

It is not exactly `coeffRef` that makes the difference. I used it to indicate that we do not need the triplets and can reuse an existing matrix.
When I profile my examples, 10%-20% of the time is spend in various memory allocation methods. This does not seem right. And I think that the biggest chunk of memory is the sparse matrix. In every iteration, we allocate the triplet lists, allocate a `Eigen::SparseMatrix`, deallocate triplet lists and deallocate the matrix. With the proposed assembler, this would be solved conveniently/automatically.

joergfunger (approved):

pmueller2 (approved):
Documented and understandable code (for me this is not always the case). I like that. 
Since the assembly loop over cells generating local matrices and such is not part of this assembler here, there is also no need for any lumped Matrix stuff (I would just generate a local mass matrix, transform it to diagonal and assemble it with the Vector Assembler).

PR #237, opened 2018-04-04
Implement LinearElasticDamage for PLANE_STRESS

TTitscher:
The class performs a deviatoric/volumetric split of the stresses and strains for its calculations. This was not (super) trivial for `PLANE_STRESS` conditions since the Kronecker delta (named `D` or `mD`) for this situation differs. For `PLANE_STRAIN` you can write
~~~
eps_volumetric = 1/3 D.transpose() * strain
eps_deviatoric = strain - D * eps_volumetric
with
D.transpose() = [ 1  1  0 ] (in 2D PLANE_STRAIN)
~~~
which is, with trivial modifications, also valid for 1D and 3D.

For `PLANE_STRAIN`, we have to introduce a second Kronecker D2, namely:
~~~
eps_volumetric = 1/3 D2.transpose() * strain
eps_deviatoric = strain - D * eps_volumetric
with
D.transpose() = [ 1  1  0 ] (in 2D, always)
D2 = (1 + nu/(nu - 1)) * D (only for 2D PLANE_STRESS)
~~~

I am not :100: % sure if this is correct. But the tests
- LinearElasticDamage::Stress(strain, omega) == (1 - omega) * LinearElastic::Stress(strain)
    - (the same for all tangents, PLANE_STRAIN and PLANE_STRESS)

run smoothly.

### DIFF
--- a/test/mechanics/constitutive/LinearElasticDamage.cpp
+++ b/test/mechanics/constitutive/LinearElasticDamage.cpp
@@ -14,7 +14,7 @@ void CheckIsotropic(NuTo::ePlaneState planeState = NuTo::ePlaneState::PLANE_STRA
     BOOST_TEST_MESSAGE("TDim = " << TDim);
     NuTo::Laws::LinearElasticDamage<TDim> linearElasticDamage(E, nu, NuTo::Laws::FULL, planeState);
     NuTo::Laws::LinearElastic<TDim> linearElastic(E, nu, planeState);
-    NuTo::EngineeringStrain<TDim> strain = NuTo::EngineeringStrain<TDim>::Random();
+    NuTo::EngineeringStrain<TDim> strain = NuTo::EngineeringStrain<TDim>::Constant(0.1);
     double omega = 0.4;
     BoostUnitTest::CheckEigenMatrix((1. - omega) * linearElastic.Stress(strain),
                                     linearElasticDamage.Stress(strain, omega));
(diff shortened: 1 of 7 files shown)


test/mechanics/constitutive/LinearElasticDamage.cpp
```diff
@@ -9,11 +9,12 @@ constexpr double nu = 0.36;
...
-void CheckIsotropic()
+void CheckIsotropic(NuTo::ePlaneState planeState = NuTo::ePlaneState::PLANE_STRAIN)
 {
-    NuTo::Laws::LinearElasticDamage<3> linearElasticDamage(E, nu);
-    NuTo::Laws::LinearElastic<3> linearElastic(E, nu);
-    NuTo::EngineeringStrain<3> strain = NuTo::EngineeringStrain<3>::Random();
+    BOOST_TEST_MESSAGE("TDim = " << TDim);
+    NuTo::Laws::LinearElasticDamage<TDim> linearElasticDamage(E, nu, NuTo::Laws::FULL, planeState);
+    NuTo::Laws::LinearElastic<TDim> linearElastic(E, nu, planeState);
+    NuTo::EngineeringStrain<TDim> strain = NuTo::EngineeringStrain<TDim>::Random();
```
joergfunger: Does this random initialization lead to different tests for each run? If yes, is this what we want, or should we at least set the seed for this random initialization?

```cpp
srand((unsigned int) time(0));
```

joergfunger (approved):
I've checked the derivation of these operators D/D2, and I get similar results.

test/mechanics/constitutive/LinearElasticDamage.cpp
TTitscher: Those should be the same values on every run. But I think that we do not want to do that. I'll replace what by some hardcode random values. [relevant xkcd :smile:](https://xkcd.com/221/)

vhirtham (approved):
Looks fine to me

PR #238, opened 2018-04-05
Pde remove jk

joergfunger:
This is a first try to get this remove jk stuff running. A meaningful benchmark is however not possible. This is due to several issues that have to be further addressed.
- There are many conversions between Eigen and DofVectors. This is true for the Hessian/Gradient, but also for the Cmat
- One of the reasons is that we require DofVectors as Input/Output of the Newton scheme. However, internally we solve with eigen matrices. This has some weird effects, such as the norm of the residual as a dofVector not being zero (since the constraint dofs are included), but when converting it, those components are removed. From my point of view, this nonzero residual vector for (real) vanishing residual seems counterintuitive.


nuto/mechanics/constraints/Constraints.h
```diff
@@ -30,13 +30,30 @@ class Constraints
...
-    //! @brief builds a sparse matrix containing the constraint terms for a specific dof type
+    //! @brief builds the rhs of the constraint equations in a sparse global vector
     //! @param dof dof type
-    //! @param numIndependentDofs number of independent dofs for the dof type dof, required for a proper resize
-    //!        of the sparse matrix
+    //! @param numDofs total number of dofs for a specfic dof type
+    //! @param time old time
+    //! @param time new time
+    //! @return sparse rhs vector of the constraint equations rhs(newTime) - rhs(oldTime)
+    Eigen::SparseVector<double> GetSparseGlobalDeltaRhs(DofType dof, int numDofs, double oldTime, double newTime) const;
```
TTitscher: This class has a quadruple code duplication in all `Rhs` methods. It seems like `AddRhs` should be the general method. For convenience, it should be called by `GetRhs` that allocates a vector, calls `AddRhs`, returns it. 

As for the sparse methods. I would vote for a simple `Eigen::VectorXd` instead of a sparse vector (so just use `GetRhs`). I totally see that this vector is sparse, like conceptual wise. But at this point, it adds more code and more complexity. And I doubt that this will have any performance impact.
As for the `Delta`. There is this common guideline that says that we should avoid arguments of the same type (`double`) next to each other. (I am aware that there are other places, where we do not follow that...) In this case, I would always doubt what the two times mean. So it returns `new - old` but takes `old, new`. But if I am not familiar with this method, I could also think '_there is a `Delta` in there_, maybe it is meant as `old, delta`'.
Conclusion. To simplify things, I propose to remove the sparse rhs methods. `GetSparseGlobalRhs` should be covered with `GetRhs` or `AddRhs` (better `AddToRhs`?), and `GetSparseGlobalRhs(...,old, new)` transforms to `GetRhs(new) - GetRhs(old)` at the call site. This will reduce the code and the dependencies, thus also the compile times.

nuto/mechanics/constraints/Constraints.h
```diff
@@ -30,13 +30,30 @@ class Constraints
...
+    void AddRhs(DofType dof, double time, Eigen::VectorXd* destination) const;
+
+    //! @brief builds a sparse matrix containing the constraint terms for a specific dof type and a unit matrix for the
+    //! independent dofs
+    //! @param dof dof type
+    //! @param numDofs number dofs for the dof type, required for a proper resize of the sparse matrix
     //! @return sparse matrix containing the constraint terms where the last block with size (numDependent x
     //! numDependent is removed)
-    Eigen::SparseMatrix<double> BuildConstraintMatrix(DofType dof, int numIndependentDofs) const;
+    Eigen::SparseMatrix<double> BuildUnitConstraintMatrix2(DofType dof, int numDofs) const;
```
TTitscher: `2`. And the `@return` seems to be out of date.

nuto/mechanics/constraints/Constraints.cpp
```diff
@@ -33,15 +33,71 @@ Eigen::VectorXd Constraints::GetRhs(DofType dof, double time) const
...
+    }
+
+    const Equations& equations = mEquations[dof];
+    int numEquations = equations.size();
+    int numIndependentDofs = numDofs - numEquations;
+
+    Eigen::SparseMatrix<double> matrix(numDofs, numIndependentDofs);
+    matrix.setZero();
+    std::vector<Eigen::Triplet<double>> tripletList;
+    for (auto i = 0; i < numIndependentDofs; i++)
```
TTitscher: This still relies on the fact that dependent dofs have a higher dof number than independent dofs. Is there a reason to not remove this restriction? I would actually like to see that in this PR since this is the real benefit of removing JK, right?

nuto/mechanics/dofs/DofVector.h
```diff
@@ -44,6 +44,19 @@ class DofVector
         return *this;
     }
 
+    //! @brief performs _uninitialized addition_ that resizes the data to the length of `rhs`
+    DofVector& operator-=(const DofVector& rhs)
```
TTitscher, line 48: Even though all the new methods in this class
- `operator -=
- `operator -`
- `SetZero`
- two times `operator()`

seem trivial, I would insist on unit tests. The `DofVector` is a very fundamental type and should be tested properly. 

Also, the implementation of this method could be replaced by 
~~~cpp
return AddScaled(rhs, -1.);
~~~
(Haven't noticed before, but `operator+` should be replaced in the same fassion.

nuto/mechanics/solver/Solve.cpp
```diff
@@ -1,23 +1,99 @@
...
+
+    for (auto rdof : dofs)
+        for (auto cdof : dofs)
+            if (rdof.Id() != cdof.Id())
+                C_dof(rdof, cdof) = Eigen::SparseMatrix<double>(C_dof(rdof, rdof).rows(), C_dof(cdof, cdof).cols());
 
-GlobalDofVector Solve(GlobalDofMatrixSparse K, GlobalDofVector f, Constraint::Constraints bcs, DofType dof,
-                      int numIndependentDofs, double time, std::string solver)
+    auto C = ToEigen(C_dof, dofs);
+    Eigen::SparseMatrix<double> Kmod = C.transpose() * K_full * C;
```
TTitscher, line 24: There are some ideas to get rid of this `ToEigen/FromEigen` mess by assembling into `Eigen` types in the first place. We can and should discuss that in another issue/PR. For now, I really like to introduce
~~~cpp
DofVector<T> DofMatrixSparse<T>::operator*(const DofVector<T>& other) const
~~~

For the `C^T * K * C`, we can also introduce
~~~cpp
DofMatrixSparse<T> DofMatrixSparse<T>::operator*(const DofMatrixSparse<T>& other) const
~~~
or, maybe better
~~~cpp
DofMatrixSparse<double> SomeCompanion::ApplyC(const DofMatrixSparse<double>& K, const DofMatrixSparse<double>& cmat) const
~~~


There are several places, mainly in `QuasistaticSolver` where this would simplify things. For now. Until we may decide to also remove the global DofVector/SparseMatrix.

nuto/mechanics/solver/Solve.cpp
```diff
@@ -1,23 +1,99 @@
...
+    Eigen::VectorXd u = EigenSparseSolve(Kmod, fmod, solver);
+    u = C * u;
+
+    // TODO: for correct size
+    DofVector<double> result = f;
+    FromEigen(u, f.DofTypes(), &result);
+    return result;
+}
+
+DofVector<double> NuTo::SolveTrialState(const DofMatrixSparse<double>& K, const DofVector<double>& f, double oldTime,
```
TTitscher, line 36: I am not sure about this method here. `Solve` generally solves the system once. This can be used in for linear solves or nonlinear iterations. 
I understand that we may want to reuse `SolveTrialState` in the future. For now, it is used once in `QuasistaticSolver` and could be implemented differently there. E.g. we do not need to build the full constraint matrix from scratch in every call. I mean, before this PR, the constraint matrix contained mostly zeros and I would not care. Now, it has loads of ones in there. But my concerns are mainly conceptually. That this method is very specific for softening problems. (although it does not hurt in other problems... not sure.)

In any case it contains code duplication. Why not call `Solve(K, K*deltaBrhs, ...)`?

nuto/mechanics/solver/Solve.h
```diff
@@ -1,18 +1,32 @@
...
-#include "nuto/mechanics/dofs/GlobalDofMatrixSparse.h"
+#include "nuto/mechanics/dofs/DofVector.h"
+#include "nuto/mechanics/dofs/DofMatrixSparse.h"
 #include "nuto/mechanics/constraints/Constraints.h"
 
 namespace NuTo
 {
 
-Eigen::VectorXd Solve(Eigen::SparseMatrix<double> A, Eigen::VectorXd b);
+DofVector<double> Solve(const DofMatrixSparse<double>& K, const DofVector<double>& f, Constraint::Constraints& bcs,
```
TTitscher, line 10: What do you think about replacing `Constraints::Constraints& bcs` with `DofMatrixSparse<double> Cmat`? This allows calculating CMat once the constraints change. Or maybe overload the method accordingly.

nuto/mechanics/tools/QuasistaticSolver.cpp
```diff
@@ -26,16 +30,16 @@ void QuasistaticSolver::SetConstraints(Constraint::Constraints constraints)
...
+        mX = mProblem.RenumberDofs(constraints, mDofs, DofVector<double>());
     else
-        mX = mProblem.RenumberDofs(constraints, mDofs, ToGlobalDofVector(ToEigen(mX, mDofs))).J;
+        mX = mProblem.RenumberDofs(constraints, mDofs, mX);
 
     for (auto dofI : mDofs)
         for (auto dofJ : mDofs)
             if (dofI.Id() == dofJ.Id())
-                mCmat(dofI, dofI) = constraints.BuildConstraintMatrix(dofI, mX[dofI].rows());
+                mCmatUnit(dofI, dofI) = constraints.BuildUnitConstraintMatrix2(dofI, mX[dofI].rows());
```
TTitscher: As far as I can tell, `mCmatUnit` is not used anymore. It is rebuild on every solve. I prefer keeping `mCmatUnit` here and adding the possibility to pass an existing `Cmat` to `Solve`.

nuto/mechanics/tools/QuasistaticSolver.cpp
```diff
@@ -44,134 +48,104 @@ void QuasistaticSolver::SetGlobalTime(double globalTime)
...
 }
 
-double QuasistaticSolver::Norm(const Eigen::VectorXd& residual) const
+double QuasistaticSolver::Norm(const DofVector<double>& residual) const
 {
-    return residual.cwiseAbs().maxCoeff();
+    // all these conversions should be removed at some point, workaround to remove JK
+    Eigen::VectorXd tmp = ToEigen(residual, mDofs);
+    auto C = ToEigen(mCmatUnit, mDofs);
+    return (C.transpose() * tmp).norm();
```
TTitscher, line 89: I like that now pass `Dof` types instead of `Eigen`. We could now use this feature to introduce separate `mTolerance` for each dof type. Not urgent though.

test/mechanics/cell/SimpleAssembler.cpp
```diff
@@ -101,17 +100,17 @@ BOOST_AUTO_TEST_CASE(AssemblerHessian)
...
-
-    BoostUnitTest::CheckEigenMatrix(Eigen::MatrixXd(hessian.JJ(d, d)), JJ);
-    BoostUnitTest::CheckEigenMatrix(Eigen::MatrixXd(hessian.JK(d, d)), JK);
-    BoostUnitTest::CheckEigenMatrix(Eigen::MatrixXd(hessian.KJ(d, d)), KJ);
-    BoostUnitTest::CheckEigenMatrix(Eigen::MatrixXd(hessian.KK(d, d)), KK);
+
+    Eigen::MatrixXd hessianE = (Eigen::MatrixXd(5,5) << 11, 12, 13,  0,  0,
+                                                       21, 22, 23,  0,  0,
+                                                       31, 32, 44, 12, 13,
+                                                        0,  0, 21, 22, 23,
```
TTitscher, line 109: That is much clearer now! :+1: 
I feared that clang format would destroy that precious formatting...

TTitscher (changes requested):
This is a nice and huge change and, I think, simplifies things a lot by decoupling assembly from constraints. There are some points I like to change. Summarized into categories:

crucial:
- Make building CMat independent from dof numbering. If dependent dofs are (accidentally) at the end, this is fine. But I think the whole point is to allow a free numbering, right? This is not possible in the current state. 
- Unit tests for the new `DofVector` methods
- eliminate code duplication
- fix warnings (last travis build that is allowed to fail)

Nice to have:
- remove the rhs SparseVector
- add (and use) `operator*` for DofMatrixSparse*DofVector
- add a ApplyC(DofMatrixSparse K,C) method

matter of taste:
- move `SolveTrialState` back to `QuasistaticProblem`
- avoid building CMat from scratch on every solve (was previously empty, now contains nDof ones)

nuto/mechanics/constraints/Constraints.h
joergfunger: There is a difference between GetRHS and GetGlobalRHS. The former returns a vector of the dimension similar to the number of constraints, the latter blows this up to the global dimension (all dof). The reason for using the sparse vector is that this vector is very sparse (in load controlled setups with zero rhs of the constraints it is actually completely vanishing).

nuto/mechanics/solver/Solve.cpp
joergfunger: I agree that this conversion to and from eigen is very messy. However, I would not prefer hiding this conversion in another layer (in this operator there is either a similar conversion or a manual implementation of the multiplication).

nuto/mechanics/solver/Solve.cpp
joergfunger: This method is not specific for softening problems, since for other linear problems it directly gives the correct result in one iteration (without significant additional effort). 
Regarding the code duplication: There is some code duplication, however doing it the way as you describe above would require to implement the DofMatrix DofVector multiplication, but as stated above, I don't think this is the way to go.

nuto/mechanics/solver/Solve.cpp
TTitscher: I suggest to avoid the conversion and perform the _manual implementation_. But I see no problem in that. We also did a manual implementation of `operator+` and `operator-` and `operator*(scalar)`.

TTitscher (approved):

test/mechanics/cell/SimpleAssembler.cpp
vhirtham: You can always use `// clang-format off` and `// clang-format on` to forbid clang format to destroy your precious formatting ... my precious :trollface:

PR #240, opened 2018-04-11
Pde global logging

TTitscher:
Add three global logger objects, accessible via
~~~cpp
#include <nuto/base/Logger.h>

NuTo::Log::Info << "info stuff\n";
NuTo::Log::Debug << "debug stuff\n";
NuTo::Log::Error << "error stuff\n";
~~~

To identify the output coming from these loggers, a prefix is added. The above code will produce
~~~
Info|  info stuff
Debug| debug stuff
Error| error stuff
~~~
either on `cout` (controlled by `::SetQuiet(bool)`) or in a file (controlled by `::OpenFile(string)`). By default, debug is quiet.


nuto/base/Logger.cpp
```diff
@@ -13,28 +15,68 @@ void NuTo::Logger::OpenFile()
     mLogFile.clear();
     if (mLogFileName.length() != 0)
     {
-        mLogFile.open(mLogFileName.c_str(), std::fstream::in | std::fstream::out | std::fstream::app);
+        mLogFile.open(mLogFileName, std::fstream::in | std::fstream::out | std::fstream::app);
```
vhirtham: Why `std::fstream::in`? Do we read somewhere?

nuto/base/Logger.cpp
```diff
@@ -13,28 +15,68 @@ void NuTo::Logger::OpenFile()
...
         if (!mLogFile.is_open())
-            throw Exception("[NuTo::Logger::OpenFile]File could not be opened.");
+            throw Exception(__PRETTY_FUNCTION__, "File " + mLogFileName + " could not be opened.");
     }
 }
 
-void NuTo::Logger::OpenFile(std::string rFileString)
+void NuTo::Logger::OpenFile(std::string filename)
```
vhirtham: Maybe `void NuTo::Logger::OpenFile()` should just call this one with `mLogFileName` instead of having its own implementation.

vhirtham (commented):
I am not sure about using global loggers. In my opinion they should be local objects. Maybe I want Infos about A and not about B, but both use the same logger type. What now? 

I am also not sure about bothering with log files internally. You can always redirect the output to the shell into a file.

Generally I think all the Logger should be is a better, object internal  std::cout with a quiet switch.

YMMV

nuto/math/EigenSparseSolve.cpp
```diff
@@ -20,6 +22,7 @@ namespace NuTo
 template <typename TSolver>
 Eigen::VectorXd SolveWithSolver(const Eigen::SparseMatrix<double>& A, const Eigen::VectorXd& b)
 {
+    Timer t(__FUNCTION__, true, Log::Debug);
```
joergfunger, line 25: Why is the timer class inside the solver, shouldn't that be outside? Where is it used?

joergfunger (commented):
In general, we should ask ourselves what is the information that the logger class should output. As far as I understand it, this is split here into Debug Output (that is almost never used in the final versions, only temporary for debugging), standard information (time for the solver) and error information. What is the advantage of using for the latter ones a (global) logger instead of a (global) stdcout and stderr? From my point of view, this can make sense when using different local loggers (i.e. because I want to redirect only the output of my solver to some file).

joergfunger (approved):

TTitscher:
I find it quite handy to collect the log files in one place, e.g. the post processing directory. In my simulations, these directories are created at run time and I cannot redirect the output to this not-yet-existing directory.

PR #242, opened 2018-04-13
Pde constant numbering independent of constraints

joergfunger:
In the already merged removeJK branch, there was still the renumbering based on the constraints included. This is no longer required, and was now removed here. While implementing this, I realized that in almost all cases, when we loop over the constraint equations, we either need the dependent term or the  independent terms, never both. Also, every equation should have  one single dependent term. As a consequence, I have split the terms in equation into the dependent term that is always set in the constructor, and the (now called independent) terms that are the other terms. That requires some changes, i.e. constraintEquation.AddTerm has to be replaced by constraintEquation.AddIndependentTerm. I could have also not renamed this, however I thought this would be clearer for people who used this functionality and with the renaming are aware that the first term in this list is no longer the dependent term.

### DIFF
--- a/nuto/mechanics/constraints/Constraints.cpp
+++ b/nuto/mechanics/constraints/Constraints.cpp
@@ -111,7 +111,7 @@ Eigen::SparseMatrix<double> Constraints::BuildUnitConstraintMatrix(DofType dof, 
 
             if (std::abs(coefficient) > 1.e-18)
                 tripletList.push_back(
-                        Eigen::Triplet<double>(numIndependentDofs + iEquation, globalDofNumber, -coefficient));
+                        Eigen::Triplet<double>(equation.GetDependentDofNumber(), globalDofNumber, -coefficient));
         }
     }
     matrix.setFromTriplets(tripletList.begin(), tripletList.end());
--- a/nuto/mechanics/constraints/Constraints.cpp
+++ b/nuto/mechanics/constraints/Constraints.cpp
@@ -84,17 +84,23 @@ Eigen::SparseMatrix<double> Constraints::BuildUnitConstraintMatrix(DofType dof, 
 
     // add unit entry for all independent dofs
     std::vector<Eigen::Triplet<double>> tripletList;
+    int curIndependent(0);
     for (auto i = 0; i < numDofs; i++)
     {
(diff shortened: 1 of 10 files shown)


nuto/mechanics/constraints/Constraints.cpp
```diff
@@ -103,17 +109,6 @@ Eigen::SparseMatrix<double> Constraints::BuildUnitConstraintMatrix(DofType dof,
...
-                int transformedDof = globalDofNumber - numIndependentDofs;
-                if (transformedDof != iEquation)
-                    throw Exception(__PRETTY_FUNCTION__, "The numbering of the dependent dofs "
-                                                         "is not in accordance with the equation numbering.");
-                continue; // Do not put the value into the matrix.
-            }
-
             if (std::abs(coefficient) > 1.e-18)
                 tripletList.push_back(
                         Eigen::Triplet<double>(numIndependentDofs + iEquation, globalDofNumber, -coefficient));
```
pmueller2: This should change to
`Eigen::Triplet<double>(equation.GetDependentDofNumber(), globalDofNumber, -coefficient));`
Seems to be a leftover from the old numbering.

nuto/mechanics/constraints/Constraints.cpp
```diff
@@ -84,17 +84,23 @@ Eigen::SparseMatrix<double> Constraints::BuildUnitConstraintMatrix(DofType dof,
...
+            tripletList.push_back(Eigen::Triplet<double>(i, curIndependent, 1.));
+            curIndependent++;
+        }
     }
+    if (curIndependent != numIndependentDofs)
+        throw Exception(__PRETTY_FUNCTION__, "There is something wrong with the number of independent dofs.");
 
     // add entries for constraint dofs
     for (int iEquation = 0; iEquation < numEquations; ++iEquation)
     {
```
pmueller2, line 101: Could be a range based loop over the equations

pmueller2 (dismissed):
I think there is a small bug in the building of the constraint matrix (see the comment). A small test would be nice that checks the correctness of the constraints matrix for not JK ordered numbering.

pmueller2 (approved):

Issue #233, opened 2018-03-26
Overloading of NodeAtCoordinate 


vhirtham:
The `NodeAtCoordinate` function of our `MeshFEM` class is currently overloaded with two versions. If you just pass the coordinates, you will get the coordinate nodes. If you throw in a additional dof, you get a dof Node. Since the stupid user (me) is still used to the old approach with only one node storing all the dof values I find the overloading here a little bit disadvantageous. I wanted to add a new Dof-Element to my mesh that uses already existing nodes. I did this:

~~~cpp
leftElementCollection.AddDofElement(
            dofWV, ElementFem({&mMesh.NodeAtCoordinate(coordLeftBoundary)}, boundaryInterpolation));
~~~ 

where `leftElementCollection` was just a reference to the corresponding element Collection. Did compile, but crashed somewhere on the way because of some missing instances (new feature) because I missed to pass the dof. The whole time I was on the wrong trail while searching for the errors origin, cause the function call didn't look obviously wrong to me. 

So maybe we could rename the functions to `CoordinateNodeAtCoordinate` and  `DofNodeAtCoordinate` to make it a little bit clearer. There might be other and better solutions, but I am lacking the big picture to tell which is the best approach.

TTitscher:
I had this issue myself several times. Mostly, when it comes to constraints and I accidentally try to constrain coordinate nodes. I added warnings for those cases. But I believe that we should solve that with a type separation of 

| `CoordinateNode`  | `DofNode` |
| --------------------- | --------------- |
| has coordinates | has multiple instances of dof values |
| -- | has a dof numbering |
| -- | can be constrained | 

and similarly `CoordinateElementInterface`  and  `DofElementInterface`. This will make other things more clear. You cannot accidentally build a jacobian from a displacement element. I am not sure about the design though. Mutliple inheritance?

~~~cpp
ElementInterface
  ::Interpolate() = 0
  ::N() / dNdXi()

CoordinateElement : ElementInterface
   ::ExtractCoordinates()

CoordinateElementFem : CoordinateElement
   ::GetNumNodes()
   ::GetNode()

CoordinateElementIga : CoordinateElement
   ::GetControlPoints()
   ::??

DofElement : ElementInterface
   ::ExtractDofValues(doftype, instance)
   ::DofNumbering()

....
~~~
Or maybe leave out `ElementInterface` to avoid the multiple inheritance?

[...]

PR #243, opened 2018-04-16
Node coordinate

vhirtham:
This PR is related to issue [#233](https://github.com/nutofem/nuto/issues/233). I implemented most things from the discussion. I skipped the separate interfaces for now, since I am not sure if it is necessary. 

In the ElementFEM class, I specialized two functions depending on the node type:

- `GetDofNumbering` will cause an assertion to fail, if you use it with coordinate nodes. I can also change it to an exception if you like that better. 
- `ExtractNodeValues` does basically the same for both specializations. Only difference is one naming related function call. I can put that function call into a separate function and specialize this one for maximum code re-usage, if you want.

I tried to use `std::enable_if` to avoid template specialization but since both functions are virtual, it is not possible or at least not trivial.


joergfunger (commented):
The removal of the additional suffixes (simple,..) stemming from the parallel development with master is necessary. Furthermore, I believe that it makes sense to have a coordinate element and a dof-element, in particular when moving to a separate representation of the geometry with coordinate elements. The inclusion of some virtual methods in the coordinate element with an exception is suboptimal, but hopefully removed as soon as this geometry class is added.

TTitscher:
As I said several times in the corresponding issue, I do not like this _old nuto_ style with virtual throw either. I think that the template approach is not suitable here. (What should TNode be, other than CoordinateNode or DofNode?) Two separate interfaces would be my preferred solution.

vhirtham:
The problem here is, that separate classes do not solve the "throw" problem, cause the corresponding function is pure virtual in the elementInterace class. Even if you remove the template, you still have the "throw override" in the coordinate element. The ElementInterface itself is needed for the N and B Matrix. Solution might be to move the function to the derived classes out of the interface, but I am not sure if that works (without touching hundreds of lines of code).

TTitscher:
Those separate classes may not have a common base. Or this common base has only common methods, like N, B or something. I mean, I can, again, refer to my first post in the issue, where I posted a sketch of what I would like. Maybe there is something wrong with that too, but you never really commented on it.

pmueller2:
Group "Separate geometry and dofs/interpolation" decided to continue the work started here to separate CoordinateMesh and DofMesh as a first step towards a FunctionSpaceLike thing.

The pull request will (hopefully) be reopened when we are done.

PR #245, opened 2018-04-23
Pde fix another bug in constraint matrix

pmueller2:
Fixes a bug when building the constraint matrix.

A prerequisite when doing this constraints stuff is a numbering that is a permutation of 0,...,numDofs or in other words: it is a dense/continuous numbering starting with 0. The constraint matrix is rectangular with as much rows as dofs and as much columns as there are independent dofs. When referring to the rows of the constraint matrix one uses the original numbering but for the columns the numbering of the independent dofs is needed (with the dependent ones removed but continuous, starting at 0). There is now a method that returns this ordered numbering that is used by the BuildConstraintMatrix routine.

The actual bugfix is indicated with a comment.


nuto/mechanics/constraints/Constraints.cpp
```diff
@@ -104,14 +121,15 @@ Eigen::SparseMatrix<double> Constraints::BuildUnitConstraintMatrix(DofType dof,
...
+            assert(independentDofNumber < numIndependentDofs);
             if (globalDofNumber == -1 /* should be NodeSimple::NOT_SET */)
                 throw Exception(__PRETTY_FUNCTION__,
                                 "There is no dof numbering for a node in equation" + std::to_string(iEquation) + ".");
 
             if (std::abs(coefficient) > 1.e-18)
                 tripletList.push_back(
-                        Eigen::Triplet<double>(equation.GetDependentDofNumber(), globalDofNumber, -coefficient));
+                        Eigen::Triplet<double>(equation.GetDependentDofNumber(), independentDofNumber, -coefficient));
         }
```
pmueller2, line 133: Here is the bug fix: exchanges globalDofNumber with independentDofNumber.

joergfunger (commented):
The solution fixes the problem. My only remark is related to the vectorJK. Its entries store for a given globalDofNumber either the dependent dof - number (if the dof is constrained) or the independent dof - number (if the dof is not constrained). For me, the information whether the dof is constrained or independent is closely related and should be stored somewhere as well (in order to be able to use the vector).

nuto/mechanics/constraints/Constraints.h
```diff
@@ -8,6 +8,21 @@ namespace NuTo
 {
 namespace Constraint
 {
+
+struct JKNumbering
+{
+    JKNumbering(Eigen::VectorXi indices, int numK)
```
joergfunger: The documentation of this class is still missing.

joergfunger (approved):
If the documentation of the class JKNumbering is completed, the pull request can be merged.

PR #247, opened 2018-04-25
Check if shapes match when adding interpolations/cells

Psirus:
Previously, when you called `AddDofInterpolation` with a quad interpolation on a mesh full of triangles, it finished without complaining, adding nodes at weird locations. This checks if the shapes of the two interpolations are the same.

The same applies for adding cells with an `ElementCollection` and an `IntegrationType`.

Also fixed a couple of typos and added outstream operators for the shapes.


Psirus:
Hehe, it already found a bug^^

In the `StructuralGradientDamage` test (specifically, [here](https://github.com/nutofem/nuto/blob/d67a8d4ae2409547ec5b7a1c96b7ead4087eb426/applications/integrationtests/mechanics/StructuralGradientDamage.cpp#L186)), we are trying to integrate line elements with a triangle integration type:
~~~cpp
IntegrationTypeTriangle integration(4);
auto cellsLeft = cellStorage.AddCells(leftElements, integration);
~~~

pmueller2 (approved):
I like these shapes. Maybe we could also implement these checks for the visualizer (Voronoi).

vhirtham (approved):

PR #248, opened 2018-04-26
Add Odeint<->Eigen helper file + small example

Psirus:
see above


pmueller2 (approved):
I am not really sure if /math is the right place for the compatibility file.

Psirus:
Muhaha, now this grave mistake will be set in stone for all eternity

Issue #210, opened 2018-03-01
Remove `ElementShapeFunctions.h/cpp`


Psirus:
I think this `ElementShapeFunctions.h/cpp` file is way to big, and collects all shape functions *for no good reason*. And it shows in names like `DerivativeShapeFunctionsTriangleOrder1`, which could just be named `DerivativeShapeFunctions` at the appropriate interpolation class. So I think moving that there is the nicer approach. In 23b8c23a I have tried that for `TriangleLinear`. Does anyone see a problem with continuing that?

TTitscher:
I am not sure about this one: 
- Ok, this class is definitely to big. 
- But it is one of the few files that survived the PDE transition. 
- It defines N,B, nodal vales as fixed size Eigen matrices. This avoids dynamic allocation(!) and provides slightly faster algorithms (up to a certain size). So they can be reused if you try to write own as-fast-as-possible components. To match the `InterpolationSimple` interface, these will be dynamic.

What if, for now, we provide them as static functions in the corresponding classes like:
~~~cpp
// class InterpolaitonTriangleLinear
static Eigen::Fix ShapeFunctions(ip){...}

Eigen::Dynamic GetShapeFunctions(ip) const override
{
return ShapeFunctions(ip);
}
~~~
The big file is gone, the change is also mainly copy-paste and we keep the option on fixed size calculations.
__edit:__ Plus, we would have them available as static functions, which can be handy, e.g. [in tests](https://github.com/nutofem/nuto/commit/23b8c23a41252b53f2b43f0b320369737a465f1f#diff-efb997cb4ef30e76f06623a34a6d3ec5R68)

Psirus:
It adds a bit of extra code, since everything is now there twice (see [InterpolationTriangleLinear.h](https://github.com/nutofem/nuto/blob/remove_shapefunctions/src/mechanics/interpolation/InterpolationTriangleLinear.h) and [InterpolationTriangleLinear.cpp](https://github.com/nutofem/nuto/blob/remove_shapefunctions/src/mechanics/interpolation/InterpolationTriangleLinear.cpp)), but I think its fine.

joergfunger:
Why can't these methods that return shape functions and derivatives be always static?

TTitscher:
They have to implement the [`InterpolationInterface`](https://github.com/nutofem/nuto/blob/PDE_reviewed/src/mechanics/interpolation/InterpolationSimple.h). And virtual methods cannot be static.

[...]

PR #250, opened 2018-04-26
Pde remove shapefunctions

pmueller2:
This basically continues work related to #210. Shape functions are moved from `ElementShapeFunctions` to the interpolation classes as static functions and renamed. It is done in the same way as for the linear triangle already in branch 23b8c23a also referenced in the issue.


nuto/mechanics/interpolation/InterpolationTriangleLinear.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    derivativeShapeFunctions(1, 0) = 1.0;
+    derivativeShapeFunctions(1, 1) = 0.0;
+
+    derivativeShapeFunctions(2, 0) = 0.0;
+    derivativeShapeFunctions(2, 1) = 1.0;
+    return derivativeShapeFunctions;
+}
+
+//////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
Psirus: Copy&paste remainder.

nuto/mechanics/interpolation/TypeDefs.h
```diff
@@ -7,19 +7,18 @@ namespace NuTo
 
 typedef Eigen::VectorXd NodeValues;
 // typedef Eigen::VectorXd NaturalCoords;
-typedef Eigen::VectorXd ShapeFunctions;
 typedef Eigen::MatrixXd DerivativeShapeFunctionsNatural;
 typedef Eigen::MatrixXd DerivativeShapeFunctionsGlobal;
 typedef Eigen::MatrixXd BMatrixStrain;
 typedef Eigen::MatrixXd BMatrixGradient;
 typedef Eigen::MatrixXd NMatrix;
```
Psirus: Can we not get rid of all of these, except for `NaturalCoords` and `GlobalCoords`?

Psirus (approved):
Nice, thanks for this labour-intensive stuff.

Also, I was wondering if it would make sense to make the `Shape` of the interpolation type public, so that you could get the actual type in case you wanted it, instead of `Shape&`?

nuto/mechanics/interpolation/TypeDefs.h
pmueller2: I will remove them.

nuto/mechanics/cell/Jacobian.h
```diff
@@ -54,43 +53,43 @@ class Jacobian
...
+                Eigen::Matrix<double, 3, 2> jacobian = CalculateFixedSize<3, 2>(nodeValues, derivativeShapeFunctions);
+                Eigen::Matrix<double, 3, 3> extendedJacobian;
                 mNormal = (jacobian.col(0)).cross(jacobian.col(1)).normalized();
                 extendedJacobian.col(0) = jacobian.col(0);
                 extendedJacobian.col(1) = jacobian.col(1);
                 extendedJacobian.col(2) = mNormal;
                 mJacobian = jacobian;
-                mInvJacobian = extendedJacobian.inverse().block<2,3>(0,0);;
+                mInvJacobian = extendedJacobian.inverse().block<2, 3>(0, 0);
+                ;
```
Psirus, line 76: Superfluous semicolon.

nuto/mechanics/cell/Jacobian.h
```diff
@@ -54,43 +53,43 @@ class Jacobian
...
             default:
             {
                 throw;
             }
             }
-        } else
+        }
+        else
         {
             throw;
```
Psirus, line 88: What is going on here? Why don't these give at least some indication what's wrong? I know this was not part of the PR, I just noticed this now.

pmueller2: Will be addressed in another PR.

PR #252, opened 2018-04-27
Include eigen_resize before algebra dispatcher

Psirus:
...otherwise, the user files become dependent on include order.

### DIFF
--- a/nuto/math/EigenOdeintCompatibility.h
+++ b/nuto/math/EigenOdeintCompatibility.h
@@ -1,7 +1,7 @@
 #pragma once
 
-#include <boost/numeric/odeint/external/eigen/eigen_algebra_dispatcher.hpp>
 #include <boost/numeric/odeint/external/eigen/eigen_resize.hpp>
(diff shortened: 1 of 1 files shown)


nuto/math/EigenOdeintCompatibility.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "Eigen/Core"
 #include <boost/numeric/odeint/external/eigen/eigen_algebra_dispatcher.hpp>
```
pmueller2: I think it is better to just exchange the two includes. 
1) This would resolve the problem since eigen_resize includes Eigen/Dense
2) Other prerequisites of the eigen_algebra_dispatcher are also included there, namely:
- #include <boost/core/enable_if.hpp>
- #include <boost/type_traits/is_base_of.hpp>
- #include <boost/numeric/odeint/algebra/algebra_dispatcher.hpp>
#include <boost/numeric/odeint/algebra/vector_space_algebra.hpp>

pmueller2 (dismissed):
Better change the two includes, see comment.

pmueller2 (approved):

PR #256, opened 2018-05-08
Add ability to create Debian packages

Psirus:
This will add a target `package` that generates a .deb package of NuTo. That can than be installed with `dpkg -i <packageName>.deb`. The advantage is that you can easily uninstall (`sudo apt remove nuto`) the files compared to the `make install` way.


pmueller2 (approved):
Cool. Just a thing I noticed: NuTo will now be installed in /usr/include, while previously with make install it ended up in /usr/local/include - so better be careful when using both.

PR #257, opened 2018-05-14
Add parallel matrix assembly using OpenMp

pmueller2:
Adds parallelization of the assembly loops using OpenMp.
CMake option ENABLE_OPENMP (default True) is added to turn it off.
To change the default number of threads set the environment variable
OMP_NUM_THREADS.


nuto/mechanics/cell/SimpleAssembler.cpp
```diff
@@ -36,85 +36,115 @@ std::vector<DofType> DofIntersection(std::vector<DofType> one, std::vector<DofTy
...
                     }
                 }
             }
         }
+#pragma omp critical
+        {
+            for (DofType dofI : dofTypes)
+                for (DofType dofJ : dofTypes)
+                {
+                    triplets(dofI, dofJ).splice(triplets(dofI, dofJ).end(), localtriplets(dofI, dofJ));
```
TTitscher, line 145: I almost always prefer vector over everything. I see your point. O(1) merging of all the thread local containers. But you pay for a bad cache layout of the list. 
I would be interested in a benchmark of
a) the performance vs std::copy(vector, vector,...)
b) the actual speedup

Btw: I thought that std::deque could provide O(1) concatenation. But it seems that this is not the case.

scripts/cmake/SetCompilerFlags.cmake
```diff
@@ -7,6 +7,11 @@ macro(set_compiler_flags)
     # the flags as they are
     set(PYTHON_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 
+    # openmp specific
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
+      set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
```
Psirus: Wrong indentation.

Psirus (approved):

nuto/mechanics/cell/SimpleAssembler.cpp
pmueller2: The bottleneck in the matrix assembly is the Eigen conversion from triplet lists to sparse matrix and since the simple assembler will vanish probably anyway I would just leave it as it is to have at least some parallel functionality (I read that also some Eigen algorithms profit from enabling omp) and do no further optimizing.

Mini benchmark of concatenation of two ranges (list/vector) with 1000 elements each using splice, insert and copy. For the vectors space is preallocated using reserve.

Benchmark          |      Time       |    CPU        |Iterations
-----------------------|-----------------|---------------|---------------
BM_ListSplice      |     58 ns       |  58 ns         | 10709471
BM_VectorInsert  |      712 ns    |    712 ns      |  968647
BM_VectorCopy  |      34952 ns |     34952 ns |     19807

Maybe not surprising.

TTitscher: > Maybe not surprising.

O(1) vs O(n), yep, not surprising. On the other hand, vector allocation is O(1) or O(log n) and list allocation is O(n). Traversing a vector is faster because of cache locality. So the question is: How does the complete process of assembly react to this?

So a simulation of the conatiner based assembly could look like the last two benchmarks, for 1000 elements created twice, concatenated once and traversed twice.

Benchmark                       |            Time   |    CPU  |Iterations
-----------------------|--------------------|----------|------------------
BM_CreateVector       |                   4739 ns     |  4739 ns   |  142789
BM_CreateList             |              78747 ns    |  78746 ns    |   8126
BM_TraverseVector      |                  2323 ns   |    2323 ns  |   307932
BM_TraverseList            |              3752 ns     |  3751 ns   |  156200
BM_CreateConcatenateTraverseVector  |    23527 ns   |   23527 ns   |   30166
BM_CreateConcatenateTraverseList     |  171229 ns  |   171224 ns     |  4129

pmueller2: You are right. Using std::vector will speed up the conversion from triplet lists to sparse matrix and the overall result is then faster (using insert for concatenation of vectors). I will change that.

PR #259, opened 2018-05-16
Add Ubuntu as another Travis build

### LINKED ISSUE #225: Compatibility with Ubuntu LTS?
Recently, we tried building NuTo PDE on an Ubuntu 16.04 based system, and sadly, that did not work at all.

Do we want to be compatible to Ubuntu LTS? If so, we would have to test that on Travis. To make it easier, we could also decide to only start supporting that with the next LTS, 18.04, to be released on 2018-04-26.

In general, I would like it to be compatible to more systems, but it means that some newer C++ features etc. will not be available.

Psirus:
What it says on the tin. Apparently works without any modifications. Fixes #225.


pmueller2 (approved):

vhirtham (approved):
Thx for doing this! :+1:

PR #260, opened 2018-05-16
Add determinant check to Gmsh import.

pmueller2:
Throws if det(J) <= 0.


pmueller2:
The 3D GmshWriter tests fail due to distorted 2nd order elements generated by gmsh that have negative jacobians. I need help with this. Maybe someone has experience in fixing these kind of meshing errors?

Psirus (approved):

PR #261, opened 2018-05-17
Remove IntegrationTypeEnum

Psirus:
Less Enum. Much coverage. :dog:


nuto/mechanics/integrationtypes/IntegrationType1D2NBoundaryGauss3Ip.h
```diff
@@ -1,13 +1,13 @@
...
+#include "IntegrationTypeBase.h"
 
 namespace NuTo
 {
 //! @author Stefan Eckardt, IFF
 //! @date July 2010
 //! @brief integration types in 1D with two nodes Gauss integration and 3 integration points
-class IntegrationType1D2NBoundaryGauss3Ip : public IntegrationType1D
+class IntegrationType1D2NBoundaryGauss3Ip : public IntegrationTypeBase
 {
```
pmueller2: I think this class can go away. It is just ordinary Gauss integration with one boundary point added. Don't know why using this.

(PS: If one would use Gauss integration where say the left node (-1) is fixed and the right one free (if this is also fixed then its Lobatto integration) this is called Radau integration and the weights and ips are different.)

pmueller2 (approved):

PR #268, opened 2018-05-31
Add functionality to groups

vhirtham:
I added a variadic template function that unites more than two groups. So instead of writing something like:

~~~cpp
auto unitedGroup = Unite(group1, Unite(Unite(group2, group3), Unite(group4, group5));
~~~

you can now write:

~~~cpp
auto unitedGroup = Unite(group1, group2, group3, group4, group5);
~~~

This works for an arbitrary number of groups. Test is also included.

### DIFF
--- a/test/base/Group.cpp
+++ b/test/base/Group.cpp
@@ -54,53 +54,45 @@ struct GroupTestFixture
     Group<Foo> groupTwo;
 };
 
+
+BOOST_FIXTURE_TEST_CASE(directAccessOperator, GroupTestFixture)
+{
+    std::vector<Foo*> expectedOne = {&a, &b, &c};
+    for (unsigned int i = 0; i < groupOne.Size(); ++i)
+        BOOST_CHECK(groupOne[i] == *expectedOne[i]);
+
+    std::vector<Foo*> expectedTwo = {&e, &d, &c};
+    for (unsigned int i = 0; i < groupTwo.Size(); ++i)
+        BOOST_CHECK(groupTwo[i] == *expectedTwo[i]);
+}
+
 BOOST_FIXTURE_TEST_CASE(unionTest, GroupTestFixture)
 {
     auto unionGroup = Unite(groupOne, groupTwo);
     BOOST_CHECK_EQUAL(unionGroup.Size(), 5);
-    auto it = unionGroup.begin();
-    BOOST_CHECK(*it == a);
 
-    std::advance(it, 1);
-    BOOST_CHECK(*it == b);
-
-    std::advance(it, 1);
-    BOOST_CHECK(*it == c);
-
-    std::advance(it, 1);
-    BOOST_CHECK(*it == e);
-
-    std::advance(it, 1);
-    BOOST_CHECK(*it == d);
+    std::vector<Foo*> expected = {&a, &b, &c, &e, &d};
+    for (unsigned int i = 0; i < unionGroup.Size(); ++i)
+        BOOST_CHECK(unionGroup[i] == *expected[i]);
 }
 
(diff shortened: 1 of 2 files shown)


pmueller2 (approved):
Nice.

test/base/Group.cpp
```diff
@@ -74,6 +74,35 @@ BOOST_FIXTURE_TEST_CASE(unionTest, GroupTestFixture)
...
+
+    std::advance(it, 1);
+    BOOST_CHECK(*it == d);
+
+    std::advance(it, 1);
+    BOOST_CHECK(*it == a);
+
+    std::advance(it, 1);
+    BOOST_CHECK(*it == b);
+}
```
Psirus: I know that this is just copied from the test above, but wouldn't something like this be simpler?
~~~cpp
int i = 0;
std::vector<Foo*> expected = {&c,&e,&d,&a,&b};
for (auto& entry : unionGroup)
    BOOST_CHECK(entry == *expected[i++]);
~~~

Psirus (approved):

test/base/Group.cpp
```diff
@@ -74,6 +74,35 @@ BOOST_FIXTURE_TEST_CASE(unionTest, GroupTestFixture)
...
+    Group<Foo> groups[5];
+    groups[0].Add(c);
+    groups[1].Add(e);
+    groups[2].Add(e);
+    groups[2].Add(d);
+    groups[3].Add(a);
+    groups[3].Add(e);
+    groups[4].Add(b);
+
+    auto unionGroup = Unite(groups[0], groups[1], groups[2], groups[3], groups[4]);
```
Psirus: Why not just
~~~cpp
Group<Foo> groupThree;
groupThree.Add(c);
groupThree.Add(a);

auto unionGroup = Unite(groupOne, groupTwo, groupThree);
~~~
instead of the C-Style array?

vhirtham:
I changed some code addressing the remarks of @Psirus . I also added the direct access operator and a corresponding test.

Psirus:
Why do we need the direct access operator?

vhirtham:
Can't come up with an example right now, but there were some occasions in the past, where I wanted to use it. @pmueller2  also mentioned it to be a nice addition. I think it doesn't hurt, since it just a pass through function using the underlying base types functionality. Additionally I prefer something like this

~~~ cpp
std::vector<Foo*> expected = {&a, &b, &c, &e, &d};
for (unsigned int i = 0; i < unionGroup.Size(); ++i)
       BOOST_CHECK(unionGroup[i] == *expected[i]);
~~~

instead of 

~~~cpp
int i = 0;
std::vector<Foo*> expected = {&c,&e,&d,&a,&b};
for (auto& entry : unionGroup)
    BOOST_CHECK(entry == *expected[i++]);
~~~

I think the first variant of comparing entries of two containers is a little bit clearer than using an indexed access and an iterator access. YMMV.

To sum it up: We might not **need** it, but it might be handy in some cases and it does not hurt.

YMMV.

TTitscher:
Two sorries, 1st for being late, 2nd for putting it here instead of an issue, I'll make a separate issue soon.

IMO the direct access operator goes beyond the original design idea of the group - a container holding references to non-duplicate objects in some deterministic way. With this access we suddenly have questions like: _What is the fifth element after uniting [1,2,3] and [4,5]?_ Or better: _How should unite be implemented so that it is intuitive for the user?_

I would solve this differently, by allowing an iterator-like interface to all the functions that currently use a `Group`. This could be done with a [`std::span`](http://en.cppreference.com/w/cpp/container/span) or similar implementations in lower c++ standards, e.g. [this one](https://github.com/martinmoene/span-lite). Then, no one is forced to use the groups. If you need vector-like access - use a vector. If you need a lot of set-like functions and uniqueness - use a set.

PR #270, opened 2018-06-07
Improve PhysicalEquations

Psirus:
This was already fixed in `master`, so now cherry-picking it for `develop`. The function `DensityLiquidWater` has been mangled by clang-format, and wasn't ideal to begin with. This proposes an improvement.


vhirtham (approved):

nuto/math/CubicSplineInterpolation.cpp
```diff
@@ -42,7 +42,7 @@ double NuTo::Math::CubicSplineInterpolation::operator()(double x)
     return y;
 }
 
-double NuTo::Math::CubicSplineInterpolation::derivative(double x)
+double NuTo::Math::CubicSplineInterpolation::derivative(double x) const
```
TTitscher, line 45: I know, you did not touch that, but for our naming convention, this could be changed to a capital `D` in `Derivative`, same goes for `Bisection`.

TTitscher (approved):

pmueller2 (approved):
Just a comment: I found 
<boost/math/interpolators/cubic_b_spline.hpp>.
Maybe we want to use it?

Psirus:
The boost class assumes equidistant spacing. This is not the case here.

NAThompson:
If you don't want equidistant spacing, boost also provides barycentric rational interpolation. O(n) construction and O(n) evaluation.

